PR #5, opened 2017-06-05
Add value indexer, label ordering and labels param

imatiach-msft:
Add value indexer, label ordering and labels param directly to metadata for scenario 5.4: User can set and get positive and negative label mappings for binary classification problems
The reindex and labels parameters interact as:
reindex -> false
labels -> false
as before, assume all double values, don't use metadata, assume natural ordering
reindex -> true
labels -> false
as before, index, use natural ordering of string indexer
reindex -> false
labels -> true
assume user knows indexing, apply label values. Currently only string type supported.
reindex -> true
labels -> true
validate labels matches column type, try to recast to label type, reindex label column


src/core/schema/src/main/scala/Categoricals.scala
```diff
@@ -261,6 +261,28 @@ class CategoricalMap[T](val levels: Array[T], val isOrdinal: Boolean = false) ex
...
+    */
+  def getDataType(columnMetadata: Metadata): DataType = {
+    val dataType =
+      if (columnMetadata.contains(ValuesString)) DataTypes.StringType
+      else if (columnMetadata.contains(ValuesLong)) DataTypes.LongType
+      else if (columnMetadata.contains(ValuesInt)) DataTypes.IntegerType
+      else if (columnMetadata.contains(ValuesLong)) DataTypes.LongType
+      else if (columnMetadata.contains(ValuesDouble)) DataTypes.DoubleType
+      else if (columnMetadata.contains(ValuesBool)) DataTypes.BooleanType
+      else throw new Exception("Unrecognized datatype in MML metadata")
```
drdarshan: Use pattern matching?
Also, what happens if the metadata contains multiple types, for example, int and categoricals?

drdarshan: Please avoid throwing a generic exception.. Can you just get these types off a Map?

src/value-indexer/src/main/scala/ValueIndexer.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+    * Fits the dictionary of values from the input column.
+    *
+    * @param dataset The input dataset to train.
+    * @return The model for transforming columns to categorical.
+    */
+  override def fit(dataset: Dataset[_]): ValueIndexerModel = {
+    val dataType = dataset.schema(getInputCol).dataType
+    val levels = dataset.select(getInputCol).distinct().collect().map(row => row(0))
```
drdarshan: Issues:
1. How to deal with missing values in the categorical columns?
2. How to deal with invalid values? Use the same behavior as StringIndexer or add an option to map invalid values to a "garbage" bin. 
3. Later: we need a way to handle merging of levels at some point. In AzureML, we would get datasets that were not properly schematized and ComputeModelStatistics needed to deal with mismatched levels
4. Much later: Need to specify an ordering function so distinct behaves deterministically on arbitrary values

src/value-indexer/src/main/scala/ValueIndexer.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    // Transform the input column to categorical
+    dataType match {
+      case _: IntegerType => addCategoricalColumn[Int](dataset)(Ordering[Int])
+      case _: LongType => addCategoricalColumn[Long](dataset)(Ordering[Long])
+      case _: DoubleType => addCategoricalColumn[Double](dataset)(Ordering[Double])
+      case _: StringType => addCategoricalColumn[String](dataset)(Ordering[String])
+      case _: BooleanType => addCategoricalColumn[Boolean](dataset)(Ordering[Boolean])
+      case _ => throw new Exception("Unsupported Categorical type " + dataType.toString)
```
drdarshan: Please use a more specific exception type...

src/value-indexer/src/test/scala/VerifyValueIndexer.scala
```diff
@@ -0,0 +1,88 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
```
drdarshan: Need to add tests for missing values in levels (make sure they are at the end), floating point levels (where x != y but x.toString == y.toString, infs and nans in the column etc)

src/value-indexer/src/main/scala/ValueIndexer.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  @DeveloperApi
+  override def transformSchema(schema: StructType): StructType =
+    ValueIndexer.validateAndTransformSchema(schema, getOutputCol)
+}
+
+/**
+  * Model produced by [[ValueIndexer]].
+  */
+class ValueIndexerModel(val uid: String,
```
drdarshan: I'm just a bit uncomfortable with making this public. It's a nice abstraction to have models all be only generated via a call to a .fit so the semantics are clear. Please check if this is feasible or a good idea..

src/value-indexer/src/main/scala/IndexToValue.scala
```diff
@@ -0,0 +1,81 @@
...
+          val map = CategoricalUtilities.getMap[String](dataset.schema(getInputCol).metadata)
+          udf((index: Int) => map.getLevel(index))
+        }
+        case _: BooleanType => {
+          val map = CategoricalUtilities.getMap[Boolean](dataset.schema(getInputCol).metadata)
+          udf((index: Int) => map.getLevel(index))
+        }
+        case _ => throw new Exception("Unsupported type " + dataType.toString)
+      }
+    dataset.withColumn(getOutputCol, getLevel(dataset(getInputCol)).as(getOutputCol))
```
drdarshan: Have you tested this with missing values? :-)

src/train-classifier/src/main/scala/TrainClassifier.scala
```diff
@@ -36,9 +37,13 @@ class TrainClassifier(override val uid: String) extends Estimator[TrainedClassif
   def getNumFeatures: Int = $(numFeatures)
   def setNumFeatures(value: Int): this.type = set(numFeatures, value)
 
-  val indexLabel = BooleanParam(this, "indexLabel", "index the label column", true)
-  def getIndexLabel: Boolean = $(indexLabel)
-  def setIndexLabel(value: Boolean): this.type = set(indexLabel, value)
+  val reindexLabel = BooleanParam(this, "reindexLabel", "reindex the label column", true)
+  def getReindexLabel: Boolean = $(reindexLabel)
```
drdarshan: Please add comments on what this parameter is :-)

src/train-classifier/src/main/scala/TrainClassifier.scala
```diff
@@ -156,7 +137,72 @@ class TrainClassifier(override val uid: String) extends Estimator[TrainedClassif
...
+        if (SparkSchema.isCategorical(dataframe, labelColumn)) {
+          throw new Exception("Column is already categorical, cannot set label values")
+        }
+        // Reindex is true, and labels are given, set levels, make column categorical given the levels
+        val labelDataType = dataset.schema(labelColumn).dataType
+        // Cast the labels to the given data type, validate labels match column type
+        labelValues.get.map(value => value.toDataType(labelDataType))
+        levels = labelValues
+        // Reindex the column to be categorical with given metadata
+        val reindexedDF = new ValueIndexerModel(Identifiable.randomUID("ValueIndexer"),
```
drdarshan: Add ValueIndexerModel constructor so you don't have to create GUIDs

src/core/schema/src/main/scala/Categoricals.scala
imatiach-msft: 1.) Used pattern matching
2.) Added throw in case duplicate types present in metadata (it is invalid in this case)

imatiach-msft: this seems very similar to the other method, I just combined them into one, and threw a better exception

src/value-indexer/src/main/scala/ValueIndexer.scala
imatiach-msft: Used UnsupportedOperationException exception type

src/train-classifier/src/main/scala/TrainClassifier.scala
imatiach-msft: added comments to the parameter and to the class to explain the interaction between the reindex labels and labels parameters

src/train-classifier/src/main/scala/TrainClassifier.scala
imatiach-msft: done

src/value-indexer/src/main/scala/IndexToValue.scala
imatiach-msft: added test

src/value-indexer/src/test/scala/VerifyValueIndexer.scala
imatiach-msft: instead of adding nulls to the levels, which didn't work out in a lot of ways, I added a special attribute to the metadata to specifying whether null levels exist

src/value-indexer/src/main/scala/ValueIndexer.scala
imatiach-msft: 1.) added handling of missing and nulls -> added special metadata flag is null value exists in metadata
2.) invalid values are added to a "garbage" bin, later on we can add more options for how the ValueIndexer/ValueIndexerModel behaves

PR #7, opened 2017-06-05
Rapid-Development

mhamilton723:
A way to quickly get code to the cluster for the demo


src/core/spark/src/main/scala/EstimatorParam.scala
```diff
@@ -34,3 +34,32 @@ class EstimatorParam(parent: String, name: String, doc: String, isValid: Estimat
...
+
+  def this(parent: Identifiable, name: String, doc: String) =
+    this(parent.uid, name, doc)
+
+  /** Creates a param pair with the given value (for Java). */
+  override def w(value: PipelineStage): ParamPair[PipelineStage] =
+    super.w(value)
+
+  override def jsonEncode(value: PipelineStage): String = {
+    throw new NotImplementedError("The transform cannot be encoded.")
```
imatiach-msft: these need to be implemented

PR #17, opened 2017-06-08
Add featurization of date, timestamp and image columns

imatiach-msft:
Add default featurization for date, timestamp spark types and our internal image type.
For featurization of date columns, convert column to double features:
year
day of week
month
day of month
For featurization of timestamp columns, same as date and in addition:
hour of day
minute of hour
second of minute
For featurization of image columns, use image data converted to double with width and height info


src/featurize/build.sbt
```diff
@@ -1,3 +1,4 @@
 //> DependsOn: core
 //> DependsOn: utils
 //> DependsOn: multi-column-adapter
+//> DependsOn: readers
```
drdarshan: Please avoid adding more dependencies unless really necessary. I wonder if ImageSchema can go into core.

drdarshan: Also, I don't think we should handle images implicitly like this..

imatiach-msft:
the dependency on readers was only in the tests, I've removed by generating the images instead of reading them from disk

imatiach-msft:
I've updated the code to only do image featurization for CNTK Learner and disabled it otherwise

elibarzilay:
I have rebased this branch on the current `master`.

PR #18, opened 2017-06-08
CNTKModel defaults to output node 0

RatanRSur:


src/cntk-model/src/main/scala/CNTKModel.scala
```diff
@@ -182,15 +183,12 @@ class CNTKModel(override val uid: String) extends Model[CNTKModel] with DefaultP
 
     val model = CNTKModel.loadModelFromBytes(getModel, device)
 
-    val setByName  = get(outputNodeName)
-    val setByIndex = get(outputNodeIndex)
-    if ((setByName.isDefined && setByIndex.isDefined) ||
-          (!setByName.isDefined && !setByIndex.isDefined))
-      throw new Exception("Must specify one and only one of outputNodeName or outputNodeIndex")
+    if ((isSet(outputNodeName) && isSet(outputNodeIndex)))
+      throw new Exception("Can't set both outputNodeName and outputNodeIndex")
```
drdarshan: Use require instead of throwing an explicit exception?

src/cntk-model/src/main/scala/CNTKModel.scala
```diff
@@ -182,15 +183,12 @@ class CNTKModel(override val uid: String) extends Model[CNTKModel] with DefaultP
...
 
-    val outputNode: Option[String] =
-      if (setByName.isDefined) setByName
-      else                     setByIndex.map(i => model.getOutputs.get(i).getName)
+    val outputNode =
+      if (isSet(outputNodeName)) get(outputNodeName)
+      else Option(model.getOutputs.get(getOrDefault(outputNodeIndex)).getName)
```
drdarshan: This seems like a harmless-enough change - why is the build failing?

mhamilton723:
This has already been done

PR #19, opened 2017-06-08
Delete the readers that are not in use.

dakirsa:


drdarshan:
Thank you, looks great!

PR #20, opened 2017-06-09
Minor script cleanup

elibarzilay:
No need for `LD_STRING`, and remove some unnecessary boilerplate in
`update_livy.py`.

### DIFF
--- a/tools/hdi/update_livy.py
+++ b/tools/hdi/update_livy.py
@@ -2,24 +2,18 @@
 # Copyright (C) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See LICENSE in project root for information.
 
-import sys
-import json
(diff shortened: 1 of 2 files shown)


tools/hdi/update_livy.py
```diff
@@ -2,24 +2,18 @@
...
+                    + " arguments.")
 
-if __name__ == "__main__":
-    try:
-        main()
-    except Exception as exn:
-        for line in str(exn).split("\n"):
-            print "[ERROR] {0}".format(line)
-        sys.exit(1)
+[_, config_file, maven_pkg] = sys.argv
```
drdarshan: List is unnecessary.. you can just do:
``_, config_file, maven_pkg = sys.argv``

PR #21, opened 2017-06-09
New example notebook to show transfer learning and DNN featurization

rastala:
Filling a gap in example Notebooks, to show DNN featurization together with SparkML model.

### DIFF
--- a/notebooks/samples/303 - Transfer Learning by DNN Featurization - Airplane or Automobile.ipynb
+++ b/notebooks/samples/303 - Transfer Learning by DNN Featurization - Airplane or Automobile.ipynb
@@ -1,0 +1,210 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 303 - Transfer Learning by DNN Featurization\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


notebooks/samples/303 - Transfer Learning by DNN Featurization - Airplane or Automobile.ipynb
```diff
@@ -0,0 +1,233 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 54,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
```
imatiach-msft: the outputs should be cleaned (removed)

drdarshan:
Hi Roope, 
Is there any way we could make this use the image reader API so we get a dataframe out more conveniently?

rastala:
Sudarshan - using ImageReader would require doing an extra step of copying the data from CDN to HDFS as images, which is not very convenient either, see: /notebooks/samples/302%20-%20Pipeline%20Image%20Transformations.ipynb  

So to address your comment we'd need a better mechanism to read images that aren't on HDFS to begin with.

Issue #11, opened 2017-06-06
Add support for missing value cleaner


rastala:
Add an Estimator that computes a missing value replacement value, such as mean, median or mode, for training data. The Estimator then produces a Model that can be applied to replace missing values.

The missing value cleaner should support one or more input columns. Different types should be supported as follows:

- Floating point numbers: mean, median, mode
- Integer numbers: median, mode
- Strings and categoricals: model
- Vectors and other composite types: not supported

The missing value cleaner should be a PipelineStage so it is compatible with SparkML pipelines.

drdarshan:
Closing since the basic missing value cleaner transform is now merged in #22

PR #22, opened 2017-06-09
Add clean missing data module

imatiach-msft:
Add clean missing data module with modes:
  1.) Replace with mean - replaces missings with mean of fit column 
  2.) Replace with median - replaces missings with approximate median of fit column 
  3.) Replace with custom value - replaces missings with custom value specified by user
As well as several smoke tests
Also added HasInputCols and HasOutputCols as traits for multiple columns (which will be needed when imputation will be added).
Resolved issue: Add support for missing value cleaner  #11

### DIFF
--- a/src/clean-missing-data/src/test/scala/VerifyCleanMissingData.scala
+++ b/src/clean-missing-data/src/test/scala/VerifyCleanMissingData.scala
@@ -1,0 +1,219 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.Estimator
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.types.StructType
+import java.lang.{Double => JDouble, Integer => JInt, Boolean => JBoolean}
+
+import org.scalactic.TolerantNumerics
+
+/**
+  * Tests to validate the functionality of Clean Missing Data estimator.
+  */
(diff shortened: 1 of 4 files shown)


rastala:
LGTM, let's remember to add example notebook once this gets into build.

src/clean-missing-data/src/test/scala/VerifyCleanMissingData.scala
```diff
@@ -0,0 +1,142 @@
...
+        }
+      }
+    }
+  }
+
+  override def createFitDataset: DataFrame = {
+    createMockDataset
+  }
+
+  override def schemaForDataset: StructType = ???
```
drdarshan: Why leave this unimplemented?

src/core/contracts/src/main/scala/Params.scala
```diff
@@ -121,6 +121,18 @@ trait HasOutputCol extends Wrappable {
   def getOutputCol: String = $(outputCol)
 }
 
+trait HasInputCols extends Wrappable {
```
drdarshan: This is great thanks but we should be very careful about versioning Core now that we are public. Is there a way to add a `@since` tag or would that be too superfluous?

imatiach-msft: yes, we can definitely do this later

imatiach-msft: sure, we can do this later

imatiach-msft:
discussed, we can add the "since" tag later, we should add it to all APIs, similar to Spark

imatiach-msft:
Updated PR - added support for string, boolean types for custom value and added test cases

PR #23, opened 2017-06-12
Fix pipeline round-trip serialization

mhamilton723:
- Made the pipeline stage param a new param type
- Fix a problem with the complexTypeConverter
- fix the spark hardcoded namespace problem

### DIFF
--- a/src/utils/src/main/python/java_params_patch.py
+++ b/src/utils/src/main/python/java_params_patch.py
@@ -1,0 +1,38 @@
+from pyspark.ml.wrapper import JavaParams
+
+@staticmethod
+def _mml_from_java(java_stage):
+    """
+    Given a Java object, create and return a Python wrapper of it.
+    Used for ML persistence.
+
+    Meta-algorithms such as Pipeline should override this method as a classmethod.
+    """
+    def __get_class(clazz):
+        """
+        Loads Python class from its name.
+        """
+        parts = clazz.split('.')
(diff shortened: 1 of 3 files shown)


src/utils/src/main/python/java_params_patch.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Meta-algorithms such as Pipeline should override this method as a classmethod.
+    """
+    def __get_class(clazz):
+        """
+        Loads Python class from its name.
+        """
+        parts = clazz.split('.')
+        module = ".".join(parts[:-1])
+        m = __import__(module)
```
drdarshan: Does this work in Python3? I thought you need to use imporlib?

drdarshan:
I don't see examples of how this is used? It looks like you need to implement a special params subtype for this logic to kick in?

src/utils/src/main/python/java_params_patch.py
mhamilton723: this is verbatim from spark, just with an additional line. I tested it in python3

mhamilton723:
I will have some stuff in the next set of PRs that uses the pipeline stage param. I figured it would be useful to have in general, if you would like it taken out until then  I can do that

src/utils/src/main/python/java_params_patch.py
drdarshan: Nice!

PR #25, opened 2017-06-12
Add a Default Serialization Mixin

mhamilton723:
- Added a serializer for arbitrary objects
- Reasonable behavior on MLLib objects
- Updated our modules to use it


src/utils/src/main/scala/Serializer.scala
```diff
@@ -0,0 +1,223 @@
...
+    val pipe = new Pipeline().setStages(Array(stage))
+    Serializer.writeMLWritable(pipe, outputPath, overwrite)
+  }
+
+  def read(path: Path): PipelineStage = {
+    Pipeline.load(path.toString).getStages(0)
+  }
+}
+
+class PipeilineArraySerializer extends Serializer[Array[PipelineStage]] {
```
RatanRSur: typo

mhamilton723: thanks!

src/core/env/src/main/scala/FileUtilities.scala
```diff
@@ -30,6 +30,16 @@ object FileUtilities {
     }
   }
 
+  def using[T <: AutoCloseable, U](disposable: T)(task: T => U): Try[U] = {
+    try {
```
drdarshan: Doesn't belong here

src/core/spark/src/main/scala/TransformParam.scala
```diff
@@ -53,4 +53,12 @@ class TransformerArrayParam(parent: String, name: String, doc: String, isValid:
   /** Creates a param pair with the given value (for Java). */
   def w(value: java.util.List[Transformer]): ParamPair[Array[Transformer]] = w(value.asScala.toArray)
 
+  override def jsonEncode(value: Array[Transformer]): String = {
+    throw new NotImplementedError("The transform cannot be encoded.")
```
drdarshan: Why override but throw not implemented?

drdarshan: Also, just setting to ??? should throw notimplemented

src/core/test/base/src/main/scala/TestBase.scala
```diff
@@ -3,19 +3,25 @@
...
 // Common test tags
 object TestBase {
+
   object Extended extends Tag("com.microsoft.ml.spark.test.tags.extended")
+
   object LinuxOnly extends Tag("com.microsoft.ml.spark.test.tags.linuxonly")
+
 }
 
 trait LinuxOnly extends TestBase {
```
drdarshan: I prefer to now have OS-specific flags in our tests - what's the use-case?

src/core/test/base/src/main/scala/TestBase.scala
```diff
@@ -153,3 +160,58 @@ abstract class TestBase extends FunSuite with BeforeAndAfterEachTestData with Be
...
+
+    val pipe = new Pipeline().setStages(Array(stageRoundTrip))
+    if (!ignoreEstimators) {
+      testRoundTripHelper(savePath + "/pipe", pipe, Pipeline, dfRoundTrip)
+    }
+    val fitPipe = pipe.fit(dfRoundTrip)
+    testRoundTripHelper(savePath + "/fitPipe", fitPipe, PipelineModel, dfRoundTrip)
+  }
+
+  override def afterAll(): Unit = {
```
drdarshan: Why not use this? https://commons.apache.org/proper/commons-io/javadocs/api-release/index.html?org/apache/commons/io/package-summary.html

src/utils/src/main/scala/Serializer.scala
```diff
@@ -0,0 +1,223 @@
...
+import org.apache.spark.ml.util.MLWritable
+import org.apache.spark.ml.{Pipeline, PipelineStage}
+import org.apache.spark.sql.{DataFrame, Dataset, SaveMode, SparkSession}
+import org.json4s.jackson.JsonMethods.{compact, render}
+import org.json4s.JsonDSL._
+import scala.reflect.runtime.universe._
+import scala.language.existentials
+import scala.reflect.ClassTag
+
+abstract class Serializer[O] {
```
drdarshan: Call it something more specific like ParamsSerializer or something? Also, this is pretty core piece of functionality so worth documenting better.

Thanks, this is really cool stuff 🥇 🥇

src/utils/src/main/scala/Serializer.scala
```diff
@@ -0,0 +1,223 @@
...
+    TypeTag(mirror, new reflect.api.TypeCreator {
+      def apply[U <: reflect.api.Universe with Singleton](m: reflect.api.Mirror[U]) = {
+        assert(m eq mirror, s"TypeTag[$tpe] defined in $mirror cannot be migrated to $m.")
+        tpe.asInstanceOf[U#Type]
+      }
+    })
+  }
+
+  def typeToSerializer[T](tpe: Type, sparkSession: SparkSession): Serializer[T] = {
+    val s = if (tpe <:< typeOf[PipelineStage]) {
```
drdarshan: Anyway you can use pattern matching?

src/core/test/base/src/main/scala/TestBase.scala
mhamilton723: This was from a while ago when OpenCV and CNTK tests were linux only. I did not add this though just reformatted the test base file because there were a lot of style errors

src/fuzzing/src/test/scala/Fuzzing.scala
```diff
@@ -45,7 +45,15 @@ class Fuzzing extends TestBase {
   // TODO verify that model UIDs match the class names, perhaps use a Trait
 
   test("Verify all estimators can be turned into pipelines, saved and loaded") {
-    estimators.foreach(est => {
+    val exemptions: Set[String] = Set(
```
drdarshan: Can just write: val exemptions = Set("foo", "bar")?

src/utils/src/main/scala/Serializer.scala
mhamilton723: This class is not tied to any Params classes so I'm not sold on ParamsSerializer. Do you have any other suggestions, or should I just add some docs?

src/utils/src/main/scala/Serializer.scala
mhamilton723: cannot use pattern matching because of erasure, I learned that the hard way!

src/fuzzing/src/test/scala/Fuzzing.scala
mhamilton723: Scalastyle complains because tests are macro expanded to something like classes and then public stuff needs to have type annotations

src/core/test/base/src/main/scala/TestBase.scala
mhamilton723: Done in the remove deltree commit

elibarzilay:
Um...  I didn't realize what's in this PR...  It would have helped to
have a *good* commit message that lists all the changes that are in it!

So, now that I looked at it:

1. The many (mostly fine, but some not) style improvements should
   definitely be done independently, in a different commit (and a
   different PR).

2. The same goes for a bunch of random changes that are included in
   this.  Like the change to `.gitignore` or the tweaking of
   `src/utils/src/main/scala/JarLoadingUtils.scala`.  These changes
   should be done separately, *maybe* with the style changes, but better
   if not.  In any case, this is very different from what this commit
   says it's doing.

3. Speaking about the hack in `JarLoadingUtils.scala` for finding
   `project-roots.txt` -- it is bad and this commit makes it worse.  If
   there's a need to find it, it should be done properly -- for example,
   when sbt makes the meta files it can also create a helper file with
   the paths.  Then it'll be a matter of a simple `import` instead of
   such hacks.  (But I don't know how intellij will cope with a source
   file that is generated.)

4. I don't get the `StreamUtilities` name -- `using` is definitely not
   something that should be in a `FileUtilities`, but it is also not
   really related to "streams".  But it is even questionable whether
   `ZipIterator` is something that should be in a "stream utilities".

5. Speaking about `ZipIterator`, there are many issues with it.  I've
   filed a separate issue about it.

6. Why does `src/core/test/base/build.sbt` depend on `core-env`?
   @akshaya-a -- I think that you had some issue with such a dependency?

mhamilton723:
Thanks for the comments here are a few replies:

1) Could you mark which style things you would like reverted? Could we also take care of style in this PR, it will be very time consuming to split it up because its been squashed. 

2) The gitignore changes are a matter of convenience, as it is annoying to work with files cluttering the repo. Again, can we just check that in, it has to happen at some point and its difficult to unsquash. The JarLoadingUtils is used in the fuzzing so all of these changes are needed for serialization.

3) Fixed (pushing a new version in a sec):)

4) Sudarshan asked that I split these up and name it like this. The zip iterator does work with streams so its been moved here, I didnt make it so I cant speak to its problems. This refactor was requested by Sudarshan.

5) It needs to depend on core-env because it  needs file and stream utilities.

PR #27, opened 2017-06-13
Update GitHub urls referenced in ARM template with the raw.githubusercontent.com link

tongwen11:
Azure can only works with the "raw.githubusercontent.com" links to GitHub items referenced in an ARM template. For example,
 https://github.com/Azure/azuremls/blob/master/tools/deployment/sparkClusterInVnetTemplate.json 
needs to be changed to
 https://raw.githubusercontent.com/Azure/mmlspark/master/tools/deployment/sparkClusterInVnetTemplate.json


tongwen11:
Against the wrong branch. Will close it.

PR #28, opened 2017-06-13
Update GitHub urls referenced in ARM templates with the raw.githubusercontent.com ones

tongwen11:
Azure can only works with the "raw.githubusercontent.com" links to GitHub items referenced in an ARM template. For example,
 https://github.com/Azure/azuremls/blob/master/tools/deployment/sparkClusterInVnetTemplate.json 
needs to be changed to
 https://raw.githubusercontent.com/Azure/mmlspark/master/tools/deployment/sparkClusterInVnetTemplate.json


tongwen11:
All the items referenced in the ARM templates are hosted in azure blob storage and the blob storage urls are used for testing.

elibarzilay:
Merging, but it will need to change to avoid using `githubusercontent`.

PR #31, opened 2017-06-14
Rename ImageTransform to ImageTransformer.

dakirsa:


dakirsa:
Talked to Maureen; we'll update docs in a separate commit.

PR #33, opened 2017-06-15
Docs part 1

elibarzilay:
This is the bigger part of the documentation work, the one that touches many files.  Should be merged soon to avoid painful rebases.


src/cntk-model/src/main/python/CNTKModel.py
```diff
@@ -12,8 +12,10 @@
 @inherit_doc
 class CNTKModel(_CNTKModel):
     """
-    :param SparkSession SparkSession: The SparkSession that will be used to find the model
-    :param str location: The location of the model, either on local or HDFS
+
+    Args:
```
drdarshan: Does Sphinx render these the same as `:param:`?

src/codegen/src/main/scala/PySparkWrapper.scala
```diff
@@ -34,113 +36,135 @@ abstract class PySparkWrapper(entryPoint: PipelineStage,
...
-    "        self.%1$s = Param(self, \"%1$s\", \"%2$s\")"
+  // Note: in the get/set with kwargs, there is an if/else that is due to the fact that since 2.1.1, kwargs is an
+  //       instance attribute. Once support for 2.1.0 is dropped, the else part of the if/else can be removed,
+  protected def classTemplate(importsString: String, inheritanceString: String,
+                              classParamsString: String,
+                              paramDefinitionsAndDefaultsString: String, paramGettersAndSettersString: String,
+                              classDocString: String, paramDocString: String,
+                              classParamDocString: String): String = {
+    s"""|$copyrightLines
+        |
```
drdarshan: Nice!

src/codegen/src/main/scala/PySparkWrapper.scala
```diff
@@ -238,39 +258,37 @@ abstract class PySparkWrapper(entryPoint: PipelineStage,
...
       val (pyParamDefault, autogenSuffix) = getParamDefault(param)
       paramsAndDefaults += pname + "=" + pyParamDefault
 
-      if (pyParamDefault != "None") {
-        paramDefinitionsAndDefaults += setParamDefaultTemplate.format(pname, pyParamDefault)
-      } else if (autogenSuffix != null) {
-        paramDefinitionsAndDefaults += setParamDefaultWithGuidTemplate.format(pname, autogenSuffix)
-      }
+      if (pyParamDefault != "None")
+        paramDefinitionsAndDefaults += s"""        self._setDefault($pname=$pyParamDefault)"""
```
drdarshan: The fixed number of spaces concerns me a bit especially since they're significant. Is it possible to explicitly specify the number of spaces?

src/codegen/src/main/scala/WrapperClassDoc.scala
```diff
@@ -0,0 +1,277 @@
...
+  * Where possible, there is also sample code to illustrate usage.
+  *
+  * The default case, TODO + classname will help identify missing class docs as more modules are added
+  * to the codebase. When new classes are added, please add a case and docstring here.
+  */
+
+object WrapperClassDoc {
+  def GenerateWrapperClassDoc(className: String): String = {
+    className match {
+      case "AssembleFeatures" =>
```
drdarshan: Can't we take these values from a file as opposed to encoding them in source? Each time we add a new class, it's very disruptive to have to change the wrapper generator.

src/downloader/src/main/python/ModelDownloader.py
```diff
@@ -81,19 +87,57 @@ def _wrap(self, iter):
...
     def remoteModels(self):
+        """
+
+        Downloads models stored remotely.
+
+        """
         return self._wrap(self._model_downloader.remoteModels())
 
     def downloadModel(self, model):
+        """
```
drdarshan: Do we really need so many new lines before and after?

src/image-transformer/src/main/python/ImageTransformer.py
```diff
@@ -23,9 +23,30 @@
     StructField(ImageFields[4], BinaryType(), True) ])   # OpenCV bytes: row-wise BGR in most cases
```
drdarshan: Not part of your PR but I wonder why we need to define ImageFields separately..

src/cntk-model/src/main/python/CNTKModel.py
mabusch: Yes. I found a sphinx extension that will take "Google Style" (also, "numpy style") and render them in the same format as :param: I opted to change the documents to this style because it makes them more readable as "classic python" for commandline help without losing the nicer API documentation style

src/codegen/src/main/scala/PySparkWrapper.scala
mabusch: Yes, it is done in other places, so we should be able to make that change here.

src/downloader/src/main/python/ModelDownloader.py
mabusch: double-checking which newlines are needed for correct formatting and which can be eliminated without causing problems; will remove any that are unnecessary for doc format.

PR #41, opened 2017-06-21
Documentation corrections and updates

mabusch:
- Parameterized indentation in formatted string that had multiple
spaces (i.e. "        some text" became "${ScopeDepth}some text")
- removed unnecessary blank lines in custom python scripts
- added docstrings and scaladocs for ValueIndexer
- corrected typos

### DIFF
--- a/src/codegen/src/main/scala/PySparkWrapper.scala
+++ b/src/codegen/src/main/scala/PySparkWrapper.scala
@@ -24,7 +24,7 @@ abstract class PySparkWrapper(entryPoint: PipelineStage,
                               entryPointName: String,
                               entryPointQualifiedName: String) {
 
-  private val ScopeDepth = "    "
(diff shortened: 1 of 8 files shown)


src/codegen/src/main/scala/PySparkWrapper.scala
```diff
@@ -24,7 +24,7 @@ abstract class PySparkWrapper(entryPoint: PipelineStage,
                               entryPointName: String,
                               entryPointQualifiedName: String) {
 
-  private val ScopeDepth = "    "
+  private val ScopeDepth = " " * 4
```
drdarshan: scopeDepth?

drdarshan: Please check for other Scala style messages..

PR #43, opened 2017-06-22
Add new pipeline-stages

mhamilton723:
Note this depends on the ConstuctorSerializable commit

Adding:
Meta Estimator
Timer
Ensemble By Key
Cacher

### DIFF
--- a/src/ensemble/src/main/scala/EnsembleByKey.scala
+++ b/src/ensemble/src/main/scala/EnsembleByKey.scala
@@ -1,0 +1,194 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.Transformer
+import org.apache.spark.ml.linalg.SQLDataTypes._
+import org.apache.spark.ml.linalg.{DenseVector, Vector, Vectors}
+import org.apache.spark.ml.param._
+import org.apache.spark.ml.util.{DefaultParamsReadable, Identifiable}
+import org.apache.spark.sql.expressions.{MutableAggregationBuffer, UserDefinedAggregateFunction}
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types._
+import org.apache.spark.sql.{Column, DataFrame, Dataset, Row}
+
+import scala.collection.mutable
+
+object EnsembleByKey extends DefaultParamsReadable[EnsembleByKey]
+
+class EnsembleByKey(val uid: String) extends Transformer with MMLParams {
+  def this() = this(Identifiable.randomUID("EnsembleByKey"))
+
+  val keys = new StringArrayParam(this, "keys", "keys to group by")
+
+  def getKeys: Array[String] = $(keys)
+
+  def setKeys(arr: Array[String]): this.type = set(keys, arr)
+
+  def setKeys(arr: String*): this.type = set(keys, arr.toArray)
+
+  def setKey(value: String): this.type = set(keys, Array(value))
+
+  val cols = new StringArrayParam(this, "cols", "Cols to ensemble")
+
+  def getCols: Array[String] = $(cols)
+
+  def setCols(arr: Array[String]): this.type = set(cols, arr)
+
+  def setCols(arr: String*): this.type = set(cols, arr.toArray)
+
+  def setCol(value: String): this.type = set(cols, Array(value))
+
+  val colNames = new StringArrayParam(this, "colNames", "Names of the result of each col")
+
+  def getColNames: Array[String] = $(colNames)
+
+  def setColNames(arr: Array[String]): this.type = set(colNames, arr)
+
+  def setColNames(arr: String*): this.type = set(colNames, arr.toArray)
+
+  def setColName(value: String): this.type = set(colNames, Array(value))
+
+  val allowedStrategies = Set("mean")
+  val strategy =
+    StringParam(this, "strategy", "How to ensemble the scores, ex: mean",
+                { x: String => allowedStrategies(x) })
+
+  def getStrategy: String = $(strategy)
+
+  def setStrategy(value: String): this.type = set(strategy, value)
+
+  setDefault(strategy -> "mean")
+
+  val collapseGroup =
+    BooleanParam(this, "collapseGroup", "whether to collapse all items in group to one entry")
+
+  def getCollapseGroup: Boolean = $(collapseGroup)
+
+  def setCollapseGroup(value: Boolean): this.type = set(collapseGroup, value)
+
+  setDefault(collapseGroup -> true)
+
+  override def transform(dataset: Dataset[_]): DataFrame = {
+
+    if (get(colNames).isEmpty) {
+      setDefault(colNames -> getCols.map(name => s"$getStrategy($name)"))
+    }
+
+    transformSchema(dataset.schema)
+
+    val strategyToFloatFunction = Map(
+      "mean" -> { (x: String, y: String) => mean(x).alias(y) }
+    )
+
+    val strategyToVectorFunction = Map(
+      "mean" -> { (x: String, y: String) =>
+        val dim = dataset.select(x).take(1)(0).getAs[DenseVector](0).size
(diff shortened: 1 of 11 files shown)


src/core/env/src/main/scala/FileUtilities.scala
```diff
@@ -30,6 +30,16 @@ object FileUtilities {
     }
   }
 
+  def using[T <: AutoCloseable, U](disposable: T)(task: T => U): Try[U] = {
```
drdarshan: This does not belong in FileUtilities?

src/core/spark/src/main/scala/TransformParam.scala
```diff
@@ -53,4 +53,12 @@ class TransformerArrayParam(parent: String, name: String, doc: String, isValid:
   /** Creates a param pair with the given value (for Java). */
   def w(value: java.util.List[Transformer]): ParamPair[Array[Transformer]] = w(value.asScala.toArray)
 
+  override def jsonEncode(value: Array[Transformer]): String = {
+    throw new NotImplementedError("The transform cannot be encoded.")
```
drdarshan: Why override but not implement?

src/core/test/base/build.sbt
```diff
@@ -1 +1 @@
-// nothing here
+//> DependsOn: core-env
```
drdarshan: Why do you need this?

src/core/env/src/main/scala/FileUtilities.scala
mhamilton723: Its right under the other using, but I can refactor it somewhere else if you see fit

src/core/spark/src/main/scala/TransformParam.scala
mhamilton723: This is the same pattern our other complex parameters have, unfortunately there is no straightforward way to json encode an arbitrary transfromer

src/core/test/base/build.sbt
mhamilton723: This comes from the serialization commit, which is a separate PR but this pr depends on it. In short, i needed some file utilities in the roundTripTestBase class

src/core/spark/src/main/scala/EstimatorParam.scala
```diff
@@ -34,3 +34,32 @@ class EstimatorParam(parent: String, name: String, doc: String, isValid: Estimat
   }
 
 }
+
+/**
+  * Param for PipelineStage.  Needed as spark has explicit params for many different types but not PipelineStage.
+  */
+class PipelineStageParam(parent: String, name: String, doc: String, isValid: PipelineStage => Boolean)
```
drdarshan: Rewrite this as a ``ComplexParam``

src/ensemble/src/main/scala/EnsembleByKey.scala
```diff
@@ -0,0 +1,187 @@
...
+  val collapseGroup = BooleanParam(this, "collapseGroup", "whether to collapse all items in group to one entry")
+
+  def getCollapseGroup: Boolean = $(collapseGroup)
+
+  def setCollapseGroup(value: Boolean): this.type = set(collapseGroup, value)
+
+  setDefault(collapseGroup -> true)
+
+  override def transform(dataset: Dataset[_]): DataFrame = {
+
```
drdarshan: Please ensure you call ``transformSchema`` to validate the incoming dataset.

src/ensemble/src/main/scala/EnsembleByKey.scala
```diff
@@ -0,0 +1,187 @@
...
+    if (get(colNames).isEmpty) {
+      setDefault(colNames -> getCols.map(name => s"avg($name)"))
+    }
+
+    val strategyToFloatFunction = Map(
+      "mean" -> { (x: String, y: String) => mean(x).alias(y) }
```
drdarshan: Please use ``avg`` or ``mean`` consistently

src/ensemble/src/main/scala/EnsembleByKey.scala
```diff
@@ -0,0 +1,187 @@
...
+    )
+
+    val newCols = getCols.zip(getColNames).map { case (inColName, outColName) =>
+      dataset.schema(inColName).dataType match {
+        case _: DoubleType =>
+          strategyToFloatFunction(getStrategy)(inColName, outColName)
+        case _: FloatType =>
+          strategyToFloatFunction(getStrategy)(inColName, outColName)
+        case v if v == VectorType && getStrategy == "mean" =>
```
drdarshan: Please create a ``Map`` similar to ``strategyToFloatFunction`` that can store the UDF so you don't have to special-case them here.

src/fuzzing/src/test/scala/Fuzzing.scala
```diff
@@ -79,33 +80,42 @@ class Fuzzing extends TestBase {
...
           (createDataSet, tr.copy(ParamMap()).asInstanceOf[PipelineStage])
         }
       tryRun(() => {
         val pipeline = new Pipeline().setStages(Array(pipelineStage))
         val pipelineModel = pipeline.fit(dataset)
         trySave(pipelineModel)
+        println(s"Sucessful round trip with $transformerName")
         ()
-      })
+      }, () => println(s"round trip failed with $transformerName"))
```
drdarshan: You can probably programatically generate a test case so failures show up atomically and get registered by JUnit... Please considering doing this when refactoring the fuzzing system.

src/meta-estimator/src/main/scala/MetaEstimator.scala
```diff
@@ -0,0 +1,65 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
```
drdarshan: This looks awfully similar to ``Pipeline`` - could you please look into reusing that if possible?

src/pipeline-stages/src/main/scala/Repartition.scala
```diff
@@ -35,13 +44,16 @@ class Repartition(val uid: String) extends Transformer with MMLParams {
...
     */
   override def transform(dataset: Dataset[_]): DataFrame = {
-
-    if (getN < dataset.rdd.getNumPartitions) {
-      dataset.coalesce(getN).toDF()
-    } else {
-      dataset.sqlContext.createDataFrame(
-        dataset.rdd.repartition(getN).asInstanceOf[RDD[Row]],
-        dataset.schema)
+    if (!getDisable){
```
drdarshan: Cleaner version:
if getDisable { 
dataset 
}
else {
..
}

src/pipeline-stages/src/main/scala/Timer.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  protected def log(str: String): Unit = {
+    if (str != "") {
+      println(str)
+    }
+  }
+
+}
+
+class Timer(val uid: String) extends Estimator[TimerModel]
```
drdarshan: As discussed, please consider only adding an enable/disable option instead of cache/count x before/after

src/ensemble/src/main/scala/EnsembleByKey.scala
```diff
@@ -0,0 +1,194 @@
...
+    val colToNewName = getCols.zip(getColNames).toMap
+
+    val newFields = schema.fields.flatMap { f =>
+      if (colSet(f.name)) {
+        val newField = StructField(colToNewName(f.name), f.dataType)
+        f.dataType match {
+          case _: DoubleType => Some(newField)
+          case _: FloatType => Some(newField)
+          case fdt if fdt == VectorType => Some(newField)
+          case t => throw new IllegalArgumentException(s"Cannot operate on type:$t with strategy:$getStrategy")
```
drdarshan: This looks like the same code in transform?
Why not follow the pattern in Spark? transform calls transformSchema to do validation.

mhamilton723: This already happens on line 77. I need to repeat some code, because actually applying the funtions requires knowing the dimensionality of the vector column which is only available at transform time

PR #44, opened 2017-06-22
Add the ImageAugmenter

mhamilton723:
Add a transformer that augments images with flips

### DIFF
--- a/src/codegen/src/main/scala/PySparkWrapperTest.scala
+++ b/src/codegen/src/main/scala/PySparkWrapperTest.scala
@@ -263,7 +263,7 @@ class SparkTransformerWrapperTest(entryPoint: Transformer,
         case "ValueIndexerModel" => valueIndexerModelString(entryPointName)
         case "_CNTKModel" | "FastVectorAssembler" | "MultiNGram" | "ImageFeaturizer" | "_ImageFeaturizer"
            | "_ImageTransformer" | "UnrollImage" | "HashTransform" | "Timer"
-           | "StopWordsRemoverTransform"
+           | "StopWordsRemoverTransform"  | "ImageSetAugmenter"
            => ""
         case _ =>
           tryFitSetupTemplate(entryPointName) + tryTransformString(entryPointName)
--- a/src/image-featurizer/build.sbt
+++ b/src/image-featurizer/build.sbt
@@ -3,3 +3,4 @@
 //> DependsOn: downloader
 //> DependsOn: cntk-model
 //> DependsOn: image-transformer
+//> DependsOn: pipeline-stages
--- a/src/image-featurizer/src/main/scala/ImageSetAugmenter.scala
+++ b/src/image-featurizer/src/main/scala/ImageSetAugmenter.scala
@@ -1,0 +1,69 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import com.microsoft.ml.spark.schema.ImageSchema
+import org.apache.spark.ml._
+import org.apache.spark.ml.param._
+import org.apache.spark.ml.util.{DefaultParamsReadable, DefaultParamsWritable, Identifiable}
+import org.apache.spark.sql._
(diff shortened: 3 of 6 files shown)


src/codegen/src/main/scala/PySparkWrapperTest.scala
```diff
@@ -261,7 +261,7 @@ class SparkTransformerWrapperTest(entryPoint: Transformer,
         case "ComputePerInstanceStatistics" => computeStatisticsString(entryPointName)
         case "IndexToValue" => indexToValueString(entryPointName)
         case "ValueIndexerModel" => valueIndexerModelString(entryPointName)
-        case "_CNTKModel" | "FastVectorAssembler" | "MultiNGram" | "ImageFeaturizer"
+        case "_CNTKModel" | "FastVectorAssembler" | "MultiNGram" | "ImageFeaturizer" | "ImageAugmenter"
            | "_ImageTransformer" | "UnrollImage" | "HashTransform" | "StopWordsRemoverTransform"
```
drdarshan: OK, I don't understand why we need to update the test each time we add a new transform. Having a single update point will be a big issue once we have more developers and each will constantly run into merge conflicts.

mhamilton723: Agreed, this is because of the python tests, hopefully we can fix this in a separate PR with a good fuzzing test framework

src/image-featurizer/src/main/scala/ImageAugmenter.scala
```diff
@@ -0,0 +1,73 @@
...
+
+class ImageAugmenter(val uid: String) extends Transformer
+  with HasInputCol with HasOutputCol with DefaultParamsWritable {
+
+  def this() = this(Identifiable.randomUID("ImageAugmenter"))
+
+  // if the problem is symmetric, the training dataset is supplemented with the "flipped" images
+  // by default, we assume the left-right symmetry, but not up-down
+  val flipLeftRight = BooleanParam(this, "flipLeftRight", "symmetric lr", false)
+
```
drdarshan: OK.. I would suggest structuring this similar to ImageTransformer where you can use fluent syntax to build the pipeline of transformations. 

Also, please add some parameters that are just not boolean (affine transforms for example).

src/image-featurizer/src/main/scala/ImageAugmenter.scala
```diff
@@ -0,0 +1,73 @@
...
+      None
+    }
+
+    val dfUD: Option[DataFrame] = if (getFlipUpDown) {
+      Some(flipImages(df, getInputCol, getOutputCol, Flip.flipUpDown))
+    } else {
+      None
+    }
+
+    List(dfLR, dfUD).flatten(x => x).foldLeft(dfID) { case (dfl, tdr) => dfl.union(tdr) }
```
drdarshan: Just making sure: does union actually move any data? Also, is pairwise concatenation the best way to combine rows of many dataframes (what if I had 10 different augmentation steps)
Finally, it looks like the training data will now be very structured: raw images, left-right flips, up-down flips.etc Can we structure so there is an option to interleave or randomize?

src/image-featurizer/src/main/scala/ImageAugmenter.scala
```diff
@@ -0,0 +1,73 @@
...
+package com.microsoft.ml.spark
+
+import com.microsoft.ml.spark.schema.ImageSchema
+import org.apache.spark.ml._
+import org.apache.spark.ml.param._
+import org.apache.spark.ml.util.{DefaultParamsReadable, DefaultParamsWritable, Identifiable}
+import org.apache.spark.sql._
+import org.apache.spark.sql.types._
+
+object ImageAugmenter extends DefaultParamsReadable[ImageAugmenter]
```
dakirsa: Sounds more like "ImageSetAugumenter"...

src/image-featurizer/src/main/scala/ImageAugmenter.scala
dakirsa: It sounds like we have several high-level question; seems like a good candidate for a (short) design review.

PR #49, opened 2017-06-23
Fix CNTKModel and ImageFeaturizer API

mhamilton723:


src/image-featurizer/src/test/scala/ImageFeaturizerSuite.scala
```diff
@@ -36,15 +36,15 @@ class ImageFeaturizerSuite extends LinuxOnly with CNTKTestUtils {
     val model = new ImageFeaturizer()
       .setInputCol(inputCol)
       .setOutputCol(outputCol)
-      .setModel(resNet)
+      .setModel(session, resNet)
```
drdarshan: Not a fan of passing in the session explicitly but don't see an alternative :-(

PR #53, opened 2017-06-26
Make deltree throw errors instead of returning a boolean

mhamilton723:


src/core/env/src/main/scala/FileUtilities.scala
```diff
@@ -30,10 +30,13 @@ object FileUtilities {
...
-  def delTree(file: File): Boolean =
-    if (!file.exists) true
-    else { if (file.isDirectory) file.listFiles.forall(delTree)
-           file.delete }
+  def delTree(file: File): Unit =
+    if (!file.exists) {
+      throw new IllegalArgumentException(s"File $file does not exist")
+    } else {
+      if (file.isDirectory) {file.listFiles.foreach(delTree)}
+      Files.delete(file.toPath)
```
drdarshan: Is this class private to our code? If not, can we make it?

src/core/env/src/main/scala/FileUtilities.scala
```diff
@@ -30,10 +30,12 @@ object FileUtilities {
     }
   }
 
-  def delTree(file: File): Boolean =
-    if (!file.exists) true
-    else { if (file.isDirectory) file.listFiles.forall(delTree)
-           file.delete }
+  private[microsoft] def delTree(file: File): Unit =
+    if (!file.exists) {
```
RatanRSur: I was just reminded of this too but it's more idiomatic to use require for IllegalArgumentExceptions

mhamilton723: can you write an example

RatanRSur:
```
private[microsoft] def delTree(file: File): Unit = {
  require(file.exists, s"File $file does not exist")
  FileUtils.deleteDirectory(file)
}
```

elibarzilay:
I don't get it: if it's a simple wrapper around `deleteDirectory` then why not use that directly?

(I thought that this was the agreed plan instead of fixing it.)

elibarzilay:
Closing -- use `forceDelete` instead of changing `delTree`, see #71.

PR #54, opened 2017-06-27
Clear the ivy cache of mmlspark when running a local build

elibarzilay:
I couldn't find any way to configure spark to avoid caching mmlspark in
ivy2, so instead just look for all files with the version string and
clear them out.  Do this in `mml-exec` since this is where such a spark
configuration would normally go.

Also fix some leftover code that resulted in two ".local"s in the
version.

### DIFF
--- a/tools/bin/mml-exec
+++ b/tools/bin/mml-exec
@@ -30,6 +30,17 @@ if [[ "$exe" == "jupyter-notebook" ]]; then
   exe="pyspark"; args=()
 fi
 
+if [[ "$MML_VERSION" = *".local"* ]]; then
+  # If we're running a locally built version, delete the cached versions in
+  # "$HOME/.ivy2".  This is a hack since it depends on this particular location.
+  # The reason that it's in this script instead of in the build script is that
+  # ideally we'd configure spark to disable caching of just mmlspark, and these
+  # kind of spark settings are done in this file.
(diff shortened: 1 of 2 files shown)


tools/bin/mml-exec
```diff
@@ -30,6 +30,17 @@ if [[ "$exe" == "jupyter-notebook" ]]; then
...
 fi
 
+if [[ "$MML_VERSION" = *".local"* ]]; then
+  # If we're running a locally built version, delete the cached versions in
+  # "$HOME/.ivy2".  This is a hack since it depends on this particular location.
+  # The reason that it's in this script instead of in the build script is that
+  # ideally we'd configure spark to disable caching of just mmlspark, and these
+  # kind of spark settings are done in this file.
+  ( cd "$HOME/.ivy2"; shopt -s globstar nullglob
+    # Note: use *$MML_VERSION* to avoid hard wiring specific paths
```
drdarshan: OK, just to confirm - this only clears the _local_ versions of our jar files? If I happen to have a released version, this will not delete them?
I'd prefer to have developers be responsible for clearing the cache and have the exec script itself be very simple.

elibarzilay: Yes, only local builds, and clears only the local cache.  I have an email that I'll send in a few minutes with a thorough explanation.

PR #56, opened 2017-06-27
RST file clean up

mabusch:
API documentation should not be generated for _ClassName.py because it is an
internal class for ClassName.py. For some of the more complex scala classes
of Transformers and Estimators, the wrapper _ClassName.py is inherited by
the custom ClassName.py, which provides the useful API documentation.


src/codegen/src/main/scala/DocGen.scala
```diff
@@ -38,12 +38,14 @@ object DocGen {
 
   def genRstFiles(): Unit = {
     // Generate a modules.rst file that lists all the .py files to be included in API documentation
-    val moduleString = allFiles(toZipDir, _.getName.endsWith(".py"))
-      .map(f => s"   ${f.getName.dropRight(3)}\n").mkString("")
+    val moduleString = allFiles(toZipDir, _.getName.endsWith(".py")).filterNot(_.getName.startsWith("_"))
+        .filterNot(_.getName.equals("java_params_patch.py"))
```
elibarzilay: Is this the file with all the class descriptions?  I prefer to have a single place with all of these hard-wired things which will serve as a checklist for things to move into the code (where this filter would get translated into some way to say in the source that no docs should be generated for it).

mabusch: No, the class descriptions is a different file/PR, which will not go out without more consultation with you. This PR addresses cleaning up modules.rst and the individual ClassName.rst files for the python API docs to deal with the questions I sent to Danil, Mark, and Ratan over the weekend regarding _CNTKModel.py vs CNTKModel.py, and the other _Class/Class pairs, as well as the java_params_patch.py file.

src/codegen/src/main/scala/DocGen.scala
```diff
@@ -38,12 +38,14 @@ object DocGen {
...
+        .map(f => s"   ${f.getName.dropRight(3)}\n").mkString("")
     writeFile(new File(inDocDir, "modules.rst"), rstFileLines(moduleString))
 
     // Generate .rst file for each PySpark wrapper - for documentation generation
-    allFiles(toZipDir, _.getName.endsWith(".py"))
+    allFiles(toZipDir, _.getName.endsWith(".py")).filterNot(_.getName.startsWith("_"))
```
drdarshan: Use regex?

src/codegen/src/main/scala/DocGen.scala
```diff
@@ -38,12 +38,14 @@ object DocGen {
...
+        .filterNot(_.getName.equals("java_params_patch.py"))
         .foreach{x => writeFile(new File(inDocDir, x.getName.dropRight(3) + ".rst"),
```
drdarshan: Minor suggestion: use org.apache.commons.io.FilenameUtils.getBaseName instead of dropRight(3).

PR #61, opened 2017-07-03
New docker documentation page

elibarzilay:
Also, a few minor tweaks in other docs.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -1,0 +1,276 @@
+# Using the MMLSpark Docker Image
+
+## Quickstart: install and run the Docker image
+
+Begin by installing [Docker for your OS][docker-products].  Then, to get the
+MMLSpark image and run it, open a terminal (powershell/cmd on Windows) and run
+
+    docker run -it -p 8888:8888 microsoft/mmlspark
+
+In your browser, go to <http://localhost:8888/> — you'll see the Docker image
+EULA, and once you accept it, the Jupyter notebook interface will start.  To
+skip this step, add `-e ACCEPT_EULA=yes` to the Docker command:
+
+    docker run -it -p 8888:8888 -e ACCEPT_EULA=y microsoft/mmlspark
+
+You can now select one of the sample notebooks and run it, or create your own.
+
+> Note: The EULA is needed only for running the MMLSpark Docker image; the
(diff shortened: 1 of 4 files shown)


docs/docker.md
```diff
@@ -0,0 +1,247 @@
...
+
+Begin by installing [Docker for your OS][docker-products].  Then, to get the
+MMLSpark image and run it, open a terminal (powershell/cmd on Windows) and run
+
+    docker run -it -p 8888:8888 microsoft/mmlspark
+
+In your browser, go to <http://localhost:8888/> — you'll see the docker image
+EULA, and once you accept it the jupyter notebook interface will start.  To skip
+this step, add `-e ACCEPT_EULA=yes` to the docker command (before the
+`microsoft/mmlspark`).
```
drdarshan: 1. Maybe clarify that EULA is for Docker only and source is released under the MIT license? 
1. Explicitly add the full like for docker run with the -e ACCEPT_EULA flag instead of writing before the microsoft/mmlspark line?
1. Spell "docker" and "jupyter" as "Docker" and "Jupyter" respectively?

docs/docker.md
```diff
@@ -0,0 +1,247 @@
...
+
+* And you can use `docker exec -it my-mmlspark bash` to start a shell in the
+  context of the running server container.  This is roughly equivalent to
+  starting a terminal in the jupyter interface.  This can also be used with a
+  docker image running as described in previous sections, though it's best to
+  add the `--name` flag so you know what the running image is called (otherwise
+  docker assigns it a random name).
+
+
+## Running other Spark executables
```
drdarshan: I've seen people ask to connect a bash shell to a running container so they can run `top`, `ps` etc. Is it worth documenting that?

docs/docker.md
elibarzilay: 1. Good point, I should have thought about adding it...  Done.
2. Done.
3. I was trying to avoid it since it looked kind of more formal...  Anyway, updated now with capitals, also for "spark" and "linux".

docs/docker.md
elibarzilay: I have added a new "Running other commands in an active container" section for that now.

PR #62, opened 2017-07-06
Move Python Class docs to text files

mabusch:
The class-level descriptions for the generated python classes
that wrap the scala classes were contained in the WrapperClassDoc.scala
file as formatted strings that were selected from a case statement
based on ClassName. There were two major disadvantages to this: (1) the
file has to be edited and Codegen has to be rebuilt every time a new
class is added and (2) developers are likely to skip this step because
the documentation does not live with the code. The text of the formatted
strings has been moved to text files that are part of the projects that
they live with. Now WrapperClassDoc.scala reads the text file to provide
the wrapper generator with the class doc string for the python.

Reformatted the docstring for ImageReader.py to keep most line to 80
characters of less for readability.


src/image-featurizer/src/main/scala/ImageFeaturizer.txt
```diff
@@ -0,0 +1,17 @@
+The ``ImageFeaturizer`` relies on a ``CNTKModel`` to do the featurization of
```
drdarshan: I see ``ClassName`` used in some places and ClassName (without the back ticks) used in others - could you please homogenize?

src/image-transformer/src/main/scala/UnrollImage.txt
```diff
@@ -0,0 +1,6 @@
+Converts the representation of an m X n pixel image to an m * n vector
+of double.
+
+The input column name is assumed to be "image", the output column name
```
drdarshan: ``image`` please..

elibarzilay:
@mabusch: why are the text files all copied to a single directory instead of reading them from the source directories?

mabusch:
The text files are copied to a single directory because the jar files have all been copied to a single directory, which is where codegen then processes them. Thus, when the python wrappers are being generated and this text is being inserted, codegen no longer has the paths from the root down to the text – it only has class names.

PR #65, opened 2017-07-10
Doc builds

elibarzilay:
Lots of changes to make doc builds.  See the commit logs for details.


src/project/build.scala
```diff
@@ -42,8 +42,12 @@ object Extras {
     "org.scalatest" %% "scalatest" % "3.0.0" % "provided"
     )
 
-  def artifactsDir = file(env("BUILD_ARTIFACTS", "../BuildArtifacts/packages/m2"))
+  def artifactsDir = file(env("BUILD_ARTIFACTS", "../BuildArtifacts"))
   def testsDir     = file(env("TEST_RESULTS", "../TestResults"))
+  def mavenDir     = artifactsDir / "packages" / "m2"
```
drdarshan: This is cool - I never knew you could concat paths like this in Scala!

tools/pydocs/src/conf.py
```diff
@@ -0,0 +1,132 @@
+# This file is execfile()d with the current directory set to its
```
drdarshan: Before you merge, it would be great to add a link to the original version this is based on.

src/project/build.scala
elibarzilay: I think that it's an sbt thing...

tools/pydocs/src/conf.py
elibarzilay: I added a comment that it is based on a file generated by `sphinx-autogen`.  (Also added a copyright line, since I forgot that earlier.)

PR #66, opened 2017-07-10
Refactor the Fuzzing System

mhamilton723:
- Added classes for experiment fuzzing
- Added classes for serialization fuzzing
- Added classes that can be used for pytest generation
- Everything requires only a few lines added to each test class


src/codegen/build.sbt
```diff
@@ -1,4 +1,5 @@
 //> DependsOn: core
+//> DependsOn: fuzzing
```
drdarshan: OK, given that we want to move codegen into its own tool, will adding fuzzing make that harder?

mhamilton723: This is so @mabusch can add python generated tests to codegen. Do you want to commit the python test generation to spark as well? If so it will need to depend on fuzzaing in some way. Alternatively python testing can be done in fuzzing

src/core/test/datagen/src/main/scala/GenerateDataset.scala
```diff
@@ -1,110 +0,0 @@
...
-/** Defines methods to generate a random spark DataFrame dataset based on given options. */
-object GenerateDataset {
-
-  /** Generates a random Spark DataFrame given a set of dataset generation constraints.
-    * @param sparkSession The spark session.
-    * @param datasetGenerationConstraints The dataset generation constraints to use.
-    * @param seed The random seed.
-    * @return A randomly generated dataset.
-    */
-  def generateDataset(sparkSession: SparkSession,
```
imatiach-msft: these files are needed to be able to generate datasets with random values based on constraints.  Please don't remove the files:
GenerateDataset.scala
GenerateDataType.scala
DatasetConstraints.scala
DatasetOptions.scala
GenerateRow.scala

I don't see how one could generate random datasets if these are removed.  I had several tests that used these methods, and I planned to add more in the future.

mhamilton723: Sorry for removing a lot of your code, right now its temporary as it needs to be re-worked. That fuzzing API was not general enough and required a lot of information at class extension time. All of the random test generation stuff can be easily added in the new API as a class that extends the ExperimentFuzzing trait. More specifically, the new API takes a sequence of stages, fittingDFs, and trainingDFs. These can be user supplied or in your case, generated with the generation options. This way its easy to either supply a simple test on one concrete dataframe, or in your case a large suite of random tests.

In the early stages i think we should focus on a simple and extensible API and then build the complex dataset generation stuff on top of it. We need a good way to quickly test the basic functionality with minimal overhead in the test class. Once we scale up our quality measures we can consider more rigorous schema based fuzzing.

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
```diff
@@ -0,0 +1,243 @@
...
+      "com.microsoft.ml.spark.EnsembleByKey",
+      "com.microsoft.ml.spark.Timer")
+    val applicableStages = pipelineStages.filter(t => !exemptions(t.getClass.getName))
+    val applicableClasses = applicableStages.map(_.getClass.asInstanceOf[Class[_]]).toSet
+    val classToFuzzer: Map[Class[_], ExperimentFuzzing[_ <: PipelineStage]] = experimentFuzzers.map(f =>
+      (f.experimentTestObjects().head.stage.getClass, f)).toMap
+    val classesWithFuzzers = classToFuzzer.keys
+    val classesWithoutFuzzers = applicableClasses.diff(classesWithFuzzers.toSet)
+    assertOrLog(classesWithoutFuzzers.isEmpty,
+      "These classes do not havce fuzzers: \n" +
```
imatiach-msft: minor: have

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
```diff
@@ -0,0 +1,243 @@
...
+      "com.microsoft.ml.spark.ValueIndexerModel",
+      "com.microsoft.ml.spark.PartitionSample",
+      "com.microsoft.ml.spark.MultiColumnAdapter"
+    )
+    val applicableStages = pipelineStages.filter(t => !exemptions(t.getClass.getName))
+    val applicableClasses = applicableStages.map(_.getClass.asInstanceOf[Class[_]]).toSet
+    val classToFuzzer: Map[Class[_], PyTestFuzzing[_ <: PipelineStage]] =
+      pytestFuzzers.map(f => (f.pyTestObjects().head.stage.getClass, f)).toMap
+    val classesWithFuzzers = classToFuzzer.keys
+    val classesWithoutFuzzers = applicableClasses.diff(classesWithFuzzers.toSet)
```
imatiach-msft: you might consider refactoring similarities in the tests

PR #81, opened 2017-07-19
Add ComplexParams readable and writable

mhamilton723:
- Add a set of serialization traits that work with arbitrary parameter types
- Fix CNTKModel/ImageFeaturizer so that byteArrays are not stored twice


src/cntk-model/src/main/scala/CNTKModel.scala
```diff
@@ -130,34 +131,31 @@ object CNTKModel extends DefaultParamsReadable[CNTKModel] {
     model
   }
 
-  override def load(path: String): CNTKModel = super.load(path)
 }
 
-trait CNTKModelParam extends Wrappable {
+@InternalWrapper
```
mhamilton723: Why is this here

src/codegen/src/main/scala/PySparkWrapperTest.scala
```diff
@@ -174,7 +174,6 @@ abstract class PySparkWrapperTest(entryPoint: PipelineStage,
         case "FastVectorAssembler" => "inputCols=\"col1\""
         case "MultiNGram"          => "inputColumns=np.array([ \"col5\", \"col6\" ])"
         case "SelectColumns"       => "cols=[\"col1\"]"
-        case "ImageFeaturizer"     => "modelSaveDir=\"file:///tmp\""
```
drdarshan: Please check why you needed to delete this..

src/core/serialize/src/main/scala/params/ArrayMapParam.scala
```diff
@@ -3,13 +3,16 @@
...
 
-import spray.json._
 import org.apache.spark.ml.util.Identifiable
+import spray.json.{DefaultJsonProtocol, _}
+
 import scala.collection.immutable.Map
 
+/**
+  * Created by marhamil on 7/21/17.
+  */
```
drdarshan: Remove?

PR #82, opened 2017-07-19
Update deployment scripts

tongwen11:
- Fix couple errors introduced during the last round of beautification
- Add a check on null resource group location when creating a new resource group
- Changes to make the coding style consistent


elibarzilay:
@tongwen11 I did some additional minor changes, can you check that it works fine?

tongwen11:
Validated. Thanks!

PR #86, opened 2017-07-25
Use Bytedeco OpenCV distribution.

imatiach-msft:
Updating Danil's PR:
https://mvnrepository.com/artifact/org.bytedeco.javacpp-presets/opencv

### DIFF
--- a/src/core/env/src/main/scala/NativeLoader.java
+++ b/src/core/env/src/main/scala/NativeLoader.java
@@ -6,6 +6,7 @@ package com.microsoft.ml.spark;
 import java.io.*;
 import java.nio.file.Files;
 import java.util.ArrayList;
+import java.util.List;
 
 /**
  * A helper class for loading native libraries from Java
@@ -41,7 +42,6 @@ public class NativeLoader {
         tempDir.deleteOnExit();
     }
 
-
     /**
      * Loads all native libraries from the jar file, if the jar contains a plain text file
      * named 'NATIVE_MANIFEST'.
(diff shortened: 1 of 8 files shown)


imatiach-msft:
We had issues with using JavaCpp opencv in our repo. The maven project structure for opencv is: http://central.maven.org/maven2/org/bytedeco/javacpp-presets/opencv/3.2.0-1.3/

We need to get the jars:
opencv-3.2.0-1.3.jar
opencv-3.2.0-1.3-linux-x86.jar 
From maven:
http://central.maven.org/maven2/org/bytedeco/javacpp-presets/opencv/3.2.0-1.3/
But I was unable to get both. One contains the .class files needed for opencv and the other contains the .so files.
In the POM file, when I specify:
   <dependency>

        <groupId>org.bytedeco.javacpp-presets</groupId>

        <artifactId>opencv</artifactId>

        <version>3.2.0-1.3</version>

    </dependency>
I am able to get the file opencv-3.2.0-1.3.jar.
And when specifying:
<dependency> 
        <groupId>org.bytedeco.javacpp-presets</groupId>

        <artifactId>opencv</artifactId>

        <version>3.2.0-1.3</version>

        <classifier>linux-x86_64</classifier>

    </dependency>
I am able to retrieve opencv-3.2.0-1.3-linux-x86.jar.
However, adding both only seems to retrieve one of the jars.
Using two classifiers doesn't seem to work either.  Using opencv-platform, as suggested on Gitter, did not work also.

elibarzilay:
Note: alternative solution in #89.

PR #89, opened 2017-07-28
Using opencv from openpnp

imatiach-msft:
Using opencv from openpnp

### DIFF
--- a/src/core/env/src/main/scala/NativeLoader.java
+++ b/src/core/env/src/main/scala/NativeLoader.java
@@ -5,7 +5,6 @@ package com.microsoft.ml.spark;
 
 import java.io.*;
 import java.nio.file.Files;
-import java.util.ArrayList;
 
 /**
  * A helper class for loading native libraries from Java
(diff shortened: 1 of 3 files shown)


src/core/env/src/main/scala/NativeLoader.java
```diff
@@ -116,6 +116,13 @@ private void extractNativeLibraries() throws IOException{
         extractionDone = true;
     }
 
+    private void extractNativeLibraries(String libName) throws IOException{
```
drdarshan: This code is not thread-safe. Is there a way you can call it inside a static block or use a lock? In Java, you can do this inside a ``synchronized`` block.

src/project/build.scala
```diff
@@ -35,7 +35,7 @@ object Extras {
     // should include these things in the distributed jar
     "io.spray"         %% "spray-json"  % "1.3.2",
     "com.microsoft.CNTK" % "cntk_jni"   % "2.0rc3",
-    "org.opencv"         % "opencv_jni" % "3.2.0"
+    "org.openpnp"         % "opencv" % "3.2.0-1"
```
drdarshan: Also, could you please make it use the CNTK jar that Ratan uploaded to Maven:
(http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22com.microsoft.cntk%22)
and make sure we don't need our Maven repo anymore? That'll completely unblock our move to Spark packages.

src/core/env/src/main/scala/NativeLoader.java
imatiach-msft: done

src/core/env/src/main/scala/NativeLoader.java
```diff
@@ -90,9 +90,9 @@ public void loadLibraryByName(String libName){
         }
         catch (UnsatisfiedLinkError e){
             try{
-                extractNativeLibraries();
```
imatiach-msft: done!

src/project/build.scala
imatiach-msft: I am getting an error when running tests when using the newest jar, maybe we can do this as part of a separate review as it might involve more changes:
Native code library failed to load.
java.lang.UnsatisfiedLinkError: no Cntk.Core.JavaBinding-2.0d in java.library.path
Exception in thread "Thread-72" java.io.EOFException
        at java.io.ObjectInputStream$BlockDataInputStream.peekByte(ObjectInputStream.java:2917)
        at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1502)
        at java.io.ObjectInputStream.readObject(ObjectInputStream.java:422)
        at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1$React.react(Framework.scala:798)
        at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1.run(Framework.scala:787)
        at java.lang.Thread.run(Thread.java:748)

imatiach-msft:
thanks, I've updated based on comments

imatiach-msft:
We had issues with using JavaCpp opencv in our repo. The maven project structure for opencv is: http://central.maven.org/maven2/org/bytedeco/javacpp-presets/opencv/3.2.0-1.3/

We need to get the jars:
opencv-3.2.0-1.3.jar
opencv-3.2.0-1.3-linux-x86.jar 
From maven:
http://central.maven.org/maven2/org/bytedeco/javacpp-presets/opencv/3.2.0-1.3/
But I was unable to get both. One contains the .class files needed for opencv and the other contains the .so files.
In the POM file, when I specify:
   <dependency>

        <groupId>org.bytedeco.javacpp-presets</groupId>

        <artifactId>opencv</artifactId>

        <version>3.2.0-1.3</version>

    </dependency>
I am able to get the file opencv-3.2.0-1.3.jar.
And when specifying:
<dependency> 
        <groupId>org.bytedeco.javacpp-presets</groupId>

        <artifactId>opencv</artifactId>

        <version>3.2.0-1.3</version>

        <classifier>linux-x86_64</classifier>

    </dependency>
I am able to retrieve opencv-3.2.0-1.3-linux-x86.jar.
However, adding both only seems to retrieve one of the jars.
Using two classifiers doesn't seem to work either.  Using opencv-platform, as suggested on Gitter, did not work also.

PR #92, opened 2017-08-01
Updating CNTKLearner to work with latest gpu branch updated with latest master

imatiach-msft:
Updating CNTKLearner to work with latest gpu branch updated with latest master.
Removed setting owner, fixed ByteArrayParam transfer from java to python, fixed path to model.
Verified full end-to-end scenario.

### DIFF
--- a/src/cntk-train/src/main/scala/CommandBuilders.scala
+++ b/src/cntk-train/src/main/scala/CommandBuilders.scala
@@ -8,13 +8,12 @@ import java.net.URI
 import scala.collection.mutable.ListBuffer
 import scala.sys.process._
 
-import com.microsoft.ml.spark.schema._
 import FileUtilities._
 import org.apache.hadoop.fs.Path
 import org.apache.spark.sql.SparkSession
 import org.slf4j.Logger
 
-import scala.sys.process._
+import StreamUtilities.using
 
 abstract class CNTKCommandBuilderBase(log: Logger) {
   val command: String
@@ -132,10 +131,8 @@ class MPICommandBuilder(log: Logger,
     val identityPath = new Path(identity)
     val outputPath = new Path(s"file:///${identityDir.getAbsolutePath}/identity")
     // Copy from wasb to local file
-    using(Seq(inputPath.getFileSystem(hadoopConf))) { fs =>
-      log.info(s"Copying from wasb: $inputPath to local: $identityPath")
-      fs(0).copyToLocalFile(inputPath, identityPath)
-      fs(0).setOwner(outputPath, System.getProperty("user.name"), null)
+    using(inputPath.getFileSystem(hadoopConf)) { fs =>
+      fs.copyToLocalFile(inputPath, identityPath)
     }.get
     var localDir = workingDir.toString.replaceFirst("file:/+", "/")
     val modelPath = new Path(localDir, new Path(outputDir, "Models")).toString
--- a/src/core/serialize/src/main/scala/params/ByteArrayParam.scala
+++ b/src/core/serialize/src/main/scala/params/ByteArrayParam.scala
@@ -7,9 +7,9 @@ package org.apache.spark.ml.param
   * types but not ByteArray.
   */
 class ByteArrayParam(parent: Params, name: String, doc: String, isValid: Array[Byte] => Boolean)
-  extends ComplexParam[Array[Byte]](parent, name, doc, isValid) {
+    extends ComplexParam[Array[Byte]](parent, name, doc, isValid) {
 
-    def this(parent: Params, name: String, doc: String) =
-      this(parent, name, doc, ParamValidators.alwaysTrue)
+  def this(parent: Params, name: String, doc: String) =
+    this(parent, name, doc, ParamValidators.alwaysTrue)
 
-  }
+}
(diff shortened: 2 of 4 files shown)


src/cntk-train/src/main/scala/CommandBuilders.scala
```diff
@@ -15,6 +15,7 @@ import org.apache.spark.sql.SparkSession
 import org.slf4j.Logger
 
 import scala.sys.process._
```
tongwen11: This import line is duplicated, there is one above.

imatiach-msft: removed duplicate

imatiach-msft:
updated to remove duplicate

src/cntk-train/src/main/python/CNTKLearner.py
```diff
@@ -8,11 +8,31 @@
...
 
 @inherit_doc
 class CNTKLearner(_CNTKLearner):
     """
+    Transforms the params from java object, with special case for byte array
+
+    Args:
+        java_model (CNTKModel): The python CNTKModel to transfer parameters to from java
+    """
+    def _custom_transfer_params_from_java(self, model):
```
mhamilton723: Is there a way to edit the main transfer params from java so we can make it more robust to these things. There will be many more custom parameters in the future

src/cntk-train/src/main/scala/CNTKLearner.scala
```diff
@@ -154,9 +154,10 @@ class CNTKLearner(override val uid: String) extends Estimator[CNTKModel] with CN
     val output = cb.runCommand
     logInfo("CNTK training finished")
     conformedData.unpersist()
+    val modelPath = "file://" + config.getModelPath
```
mhamilton723: SHould this be automatic? What about when the model is on Wasb?

src/cntk-train/src/main/scala/CommandBuilders.scala
```diff
@@ -133,9 +132,7 @@ class MPICommandBuilder(log: Logger,
     val outputPath = new Path(s"file:///${identityDir.getAbsolutePath}/identity")
     // Copy from wasb to local file
     using(Seq(inputPath.getFileSystem(hadoopConf))) { fs =>
```
mhamilton723: I added a using with one object so no need for the Seq wrapper

src/core/serialize/src/main/scala/params/ByteArrayParam.scala
```diff
@@ -12,4 +14,15 @@ class ByteArrayParam(parent: Params, name: String, doc: String, isValid: Array[B
     def this(parent: Params, name: String, doc: String) =
       this(parent, name, doc, ParamValidators.alwaysTrue)
 
-  }
+    /** Creates a param pair with a java.util.List of values (for Java and Python). */
+    def w(value: java.util.List[java.lang.Byte]): ParamPair[Array[Byte]] =
+      w(value.asScala.map(_.asInstanceOf[Byte]).toArray)
+
```
mhamilton723: I dont want to json encode these as I think it causes memory problems with large models

tongwen11:
Validated the cifar10 notebook run in a freshly deployed environment with success.

src/cntk-train/src/main/scala/CNTKLearner.scala
imatiach-msft: for CNTKLearner it is currently always local, but you're right eventually we want to support both once we have the parquet reader working.

src/cntk-train/src/main/python/CNTKLearner.py
imatiach-msft: discussed, the code is a bit more convoluted than we thought, we will keep this as is for now

src/cntk-train/src/main/scala/CommandBuilders.scala
imatiach-msft: nice, I've removed the Seq()

src/core/serialize/src/main/scala/params/ByteArrayParam.scala
imatiach-msft: removed the jsonEncode/Decode as they were not being used anyway to fix this specific issue

tongwen11:
Also validated that in the client mode, the set owner operation is not needed.

src/cntk-train/src/main/python/CNTKLearner.py
```diff
@@ -8,11 +8,31 @@
...
+        for param in model.params:
+            if model._java_obj.hasParam(param.name):
+                java_param = model._java_obj.getParam(param.name)
+                if model._java_obj.isSet(java_param):
+                    param_val = model._java_obj.getOrDefault(java_param)
+                    if isinstance(param_val, (bytearray, bytes)):
```
drdarshan: See my other comment on Mark's PR #98 - checking ``isinstance`` is generally not preferred in Python. 
Also, does this code work in both 2.7 and 3.6?

src/cntk-train/src/main/scala/CNTKLearner.scala
drdarshan: +1. Can't we make config.getModelPath return the correct URI (local vs. HDFS vs. WASB etc)

PR #98, opened 2017-08-04
Fix Pyspark Wrapper Generation

mhamilton723:
-fix a problem where internal wrappers could not be serialized
-fix a problem where byte array params were not transferred to python properly

### DIFF
--- a/src/codegen/src/main/scala/PySparkWrapper.scala
+++ b/src/codegen/src/main/scala/PySparkWrapper.scala
@@ -35,8 +35,9 @@ abstract class PySparkWrapper(entryPoint: PipelineStage,
     ""
   }
 
-  // Note: in the get/set with kwargs, there is an if/else that is due to the fact that since 2.1.1, kwargs is an
-  //       instance attribute. Once support for 2.1.0 is dropped, the else part of the if/else can be removed,
+  // Note: in the get/set with kwargs, there is an if/else that is due to the fact that since 2.1.1,
+  //   kwargs is an instance attribute.  Once support for 2.1.0 is dropped, the else part of the
+  //   if/else can be removed
   protected def classTemplate(importsString: String, inheritanceString: String,
                               classParamsString: String,
                               paramDefinitionsAndDefaultsString: String, paramGettersAndSettersString: String,
@@ -149,7 +150,8 @@ abstract class PySparkWrapper(entryPoint: PipelineStage,
         |    @staticmethod
         |    def _from_java(java_stage):
         |        module_name=${entryPointName}.__module__
-        |        module_name=module_name.rsplit(".", 1)[0] + ".${entryPointName}"
+        |        module_name=module_name.rsplit(".", 1)[0] + ".${
+      if (entryPointName.startsWith("_")) entryPointName.tail else entryPointName}"
         |        return from_java(java_stage, module_name)
         |""".stripMargin
   }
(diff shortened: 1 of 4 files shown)


src/cntk-model/src/main/scala/CNTKModel.scala
```diff
@@ -142,6 +142,8 @@ class CNTKModel(override val uid: String) extends Model[CNTKModel] with ComplexP
   val model: ByteArrayParam =
     new ByteArrayParam(this, "model", "Array of bytes containing the serialized CNTKModel")
 
+  model.getClass
```
imatiach-msft: should this be removed?

src/codegen/src/main/scala/PySparkWrapper.scala
```diff
@@ -164,7 +165,8 @@ abstract class PySparkWrapper(entryPoint: PipelineStage,
 
   val psType: String
   private lazy val objectBaseClass: String = "Java" + psType
-  private lazy val autoInheritedClasses = Seq("JavaMLReadable", "JavaMLWritable", objectBaseClass)
+  private lazy val autoInheritedClasses =
+    Seq("ComplexParamsMixin", "JavaMLReadable", "JavaMLWritable", objectBaseClass)
```
imatiach-msft: nice fix!

mhamilton723: thx legend

src/cntk-model/src/main/python/CNTKModel.py
```diff
@@ -8,6 +8,7 @@
 
 from mmlspark._CNTKModel import _CNTKModel
 from pyspark.ml.common import inherit_doc
+from mmlspark.Utils import *
```
drdarshan: Please avoid ``import *``. Also, I'd prefer to keep casing of modules consistent so please consider ``mmlspark.utils`` instead of ``mmlspark.Utils``.

src/core/schema/src/main/python/TypeConversionUtils.py
```diff
@@ -28,16 +30,20 @@ def complexTypeConverter(name, value, cache):
...
-            java_value.append(v._java_obj)
-        return java_value
-    if hasattr(value, "_transfer_params_to_java"):
-        value._transfer_params_to_java()
-    if hasattr(value, "_java_obj"):
-        return value._java_obj
-    else:
-        return value._to_java()
+    try:
+        if isinstance(value, list):
```
drdarshan: OK, do you really need ``value`` to be a list or can it be any iterable object? I generally prefer to not check for concrete instances since it's not really idiomatic in a duck typed language.

mhamilton723: You are probably right, I just wanted to keep any alterations to this code to a minimum for fear of inadvertently breaking something. After whacking through this jungle, I think we need to refactor codegen and how we abstract the python-scala handshaking. It should probabbly be planned out and discussed. However this is a fix for now.

mhamilton723:
Because we dont really have python tests that are full enough to catch these things. The fuzzing architecture and new pytest architecture will hopefully catch alot of these and give us a very robust python functionality check

src/core/schema/src/main/python/TypeConversionUtils.py
drdarshan: OK..

PR #102, opened 2017-08-11
Add `hadoop-azure` to make wasb access work

elibarzilay:
Also adds `azure-storage` as a dependency.

### DIFF
--- a/src/project/build.scala
+++ b/src/project/build.scala
@@ -26,13 +26,17 @@ object Extras {
   def sparkVer = env("SPARK_VERSION", null)
 
   def commonLibs = Seq(
-    "org.apache.spark"   %% "spark-core"  % sparkVer % "provided",
-    "org.apache.spark"   %% "spark-mllib" % sparkVer % "provided",
-    "org.scalatest"      %% "scalatest"   % "3.0.0"  % "provided",
+    "org.apache.spark"   %% "spark-core"   % sparkVer % "provided",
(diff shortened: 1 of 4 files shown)


src/project/build.scala
```diff
@@ -26,13 +26,15 @@ object Extras {
...
+    "org.scalatest"      %% "scalatest"    % "3.0.0"  % "provided",
     // should include these things in the distributed jar
-    "io.spray"           %% "spray-json"  % "1.3.2",
-    "com.microsoft.cntk" %  "cntk"        % "2.1",
-    "org.openpnp"        %  "opencv"      % "3.2.0-1"
+    "io.spray"           %% "spray-json"   % "1.3.2",
+    "com.microsoft.cntk"  % "cntk"         % "2.1",
+    "org.openpnp"         % "opencv"       % "3.2.0-1",
+    // needed for wasb access
+    "org.apache.hadoop"   % "hadoop-azure" % "2.7.3"
```
drdarshan: This is only for the Spark distribution inside Docker.. why do we need to add it to our compile-time dependencies?

elibarzilay: 1. It's not a big thing on one hand, and it makes managing this dependency much easier than downloading from a hard-wired url and filtering out just these two (the full thing is huge).

2. Wouldn't it be possible for our code (or users code) to run into this problem when using the library as is (outside of docker) and not in an hdicluster context?

PR #106, opened 2017-08-14
Update deployment template and setup document with the updated links

tongwen11:
- In the master ARM template, update the version of install-mmlspark.sh to 0.6
- In the setup document, correct/update with the new name for the script to setup password-less ssh connection
- In the setup document, update the example to open an ARM template in the portal with the GitHub url

### DIFF
--- a/docs/azure-setup.md
+++ b/docs/azure-setup.md
@@ -138,34 +138,34 @@ the Azure CLI 2.0 using the instruction found in [Install Azure CLI
 
 CNTK training using MMLSpark requires passwordless ssh connection from the
 HDInsight spark cluster to the GPU VM.  The setup of this connection is done
-through a [script](../tools/deployment/gpuvmsetup.sh) and it needs to be done
+through a [script](../tools/hdi/setup-ssh-keys.sh) and it needs to be done
 once.  The script takes two parameters:
 
-    ./setup-ssh.sh -m <vmName> -u <userName>
+    ./setup-ssh-keys.sh <vm-name> [<username>]
 
-The parameter userName is optional if the HDI spark cluster and the GPU VM share
-the same ssh user name.
+The `<username>` parameter is optional if the HDI spark cluster and the GPU VM
+share the same ssh user name.
 
 ## Shutdown GPU VM to save money
 
-Azure will stop billing if a VM is in the "Stopped (Deallocated)" state.  Please
-note that Billing will continue if a VM is in the "Stopped" state.  In Azure
-portal, clicking on the Stop button will make the VM in the "Stopped
-(Deallocated)" state and clicking on the Start button to bring the VM back.
-Please refer to this
-[article](https://buildazure.com/2017/03/16/properly-shutdown-azure-vm-to-save-money/)
+Azure will stop billing if a VM is in the "Stopped (**Deallocated**)" state.
+Please note that billing will continue if a VM is in the "Stopped" state (so
+make sure it is *Deallocated* to avoid billing).  In Azure portal, clicking the
+"Stop" button will get VM into a "Stopped (Deallocated)" state and clicking the
+"Start" button brings the VM back.  See "[Properly Shutdown Azure VM to Save
+Money](https://buildazure.com/2017/03/16/properly-shutdown-azure-vm-to-save-money/)"
 for more details.
 
-Here is an example of the steps to Stop and Start a VM in Azure PowerShell:
-
-    Login-AzureRmAccount
-    Select-AzureRmSubscription -SubscriptionID "90856715-5231-42a2-9069-4bf974b1xxxx"
-    Stop-AzureRmVM -ResourceGroupName "MyResourceGroupName" -Name "mygpuvm"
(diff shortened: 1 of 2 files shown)


docs/azure-setup.md
```diff
@@ -138,12 +138,12 @@ the Azure CLI 2.0 using the instruction found in [Install Azure CLI
 
 CNTK training using MMLSpark requires passwordless ssh connection from the
 HDInsight spark cluster to the GPU VM.  The setup of this connection is done
-through a [script](../tools/deployment/gpuvmsetup.sh) and it needs to be done
+through a [script](../tools/hdi/setup-ssh-keys/.sh) and it needs to be done
```
tongwen11: A typo here will update.

docs/azure-setup.md
```diff
@@ -79,7 +79,7 @@ An ARM template can be opened within Azure Portal via the following REST API:
 
 The URI can be one for either an Azure Blob or a GitHub file.  For example,
 
-    https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Ftongtest.blob.core.windows.net%2Fcntk%2FazureDeployMainTemplate.json
+    https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fmmlspark%2Fmaster%2Ftools%2Fdeployment%2FazureDeployMainTemplate.json
```
elibarzilay: It is not a good idea to depend on a path from github in this way.  I'd much rather see some solution that doesn't depend on GH at all.  (And I prefer to keep the `tongtest` since it's the magic string of all places that need to be updated when everything is sorted out.)

docs/azure-setup.md
```diff
@@ -138,12 +138,12 @@ the Azure CLI 2.0 using the instruction found in [Install Azure CLI
...
 HDInsight spark cluster to the GPU VM.  The setup of this connection is done
-through a [script](../tools/deployment/gpuvmsetup.sh) and it needs to be done
+through a [script](../tools/hdi/setup-ssh-keys.sh) and it needs to be done
 once.  The script takes two parameters:
 
-    ./setup-ssh.sh -m <vmName> -u <userName>
+    ./setup-ssh-keys.sh <vm-name> [<username>]
 
-The parameter userName is optional if the HDI spark cluster and the GPU VM share
+The parameter username is optional if the HDI spark cluster and the GPU VM share
```
elibarzilay: Should be: `The username parameter`

tools/deployment/sparkClusterInVnetTemplate.json
```diff
@@ -98,7 +98,7 @@
    "id": "[resourceId('Microsoft.Network/virtualNetworks', concat(parameters('clusterName'),'-vnet'))]",
    "subnet": "[concat(resourceId('Microsoft.Network/virtualNetworks', concat(parameters('clusterName'),'-vnet')), concat('/subnets/', parameters('virtualNetworkSubnetName')))]"
   },
-  "scriptActionUri": "https://mmlspark.azureedge.net/buildartifacts/0.5/install-mmlspark.sh"
+  "scriptActionUri": "https://mmlspark.azureedge.net/buildartifacts/0.6/install-mmlspark.sh"
```
elibarzilay: Also over here: this should not rely on a specific build artifact.  Maybe the template file itself should be something that gets generated to not depend on a specific version like this?

tongwen11: I can make scriptActionUri a parameter so that user have the option to override it at deployment time.

docs/azure-setup.md
tongwen11: So the concern here is the file can be relocated so that the url is broken? This is the doc that describes the usage of those deployment templates and scripts. In this particular example, the url is what Azure API requires. (All Azure ARM templates are hosted in GitHub and referenced using GitHub url in the published documents.) Don't see a way to avoid it. At other places in this doc, the relative path is used for linkage. If a file is relocated, the relative path still needs to be updated. 

The other usage of my blob storage (tongtest) is to host the prebuilt hdfs-mount and cntk. We can move them to https://mmlspark.azureedge.net I assume. This looks like a different problem than this one though.

tongwen11: Talked more about the two issues mentioned here with Eli:
- GitHub urls used in the doc example  and in the master deployment template 
- Url to a particular version of script action hardcoded in the deployment templates.

The current plan is to have the build process generate the final version of the templates using the checked-in version as the template where the build process will fill in the latest version of the script action and references to the child deployment templates. Since the templates are hosted at  https://mmlspark.azureedge.net, internal and external reference will not involve any GitHub urls and the urls to the script actions are updated.

I will undo the change here.

docs/azure-setup.md
tongwen11: updated

tools/deployment/sparkClusterInVnetTemplate.json
tongwen11: Please see the comment above.

PR #115, opened 2017-08-18
Add Medical Entity extraction Notebook

mhamilton723:
- also fix the CNTKmodel setOutputINodendex

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Python 3.5+.  See the API documentation
 * Easily ingest images from HDFS into Spark `DataFrame` ([example:301])
 * Pre-process image data using transforms from OpenCV ([example:302])
 * Featurize images using pre-trained deep neural nets using CNTK ([example:301])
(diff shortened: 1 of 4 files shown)


rastala:
Could you add a source attribution for the example text. I think it's from this article: 
http://acrabstracts.org/abstract/baricitinib-methotrexate-or-baricitinib-plus-methotrexate-in-patients-with-early-rheumatoid-arthritis-who-had-received-limited-or-no-treatment-with-disease-modifying-anti-rheumatic-drugs-dmards-p/

rastala:
Also, please add more verbiage about what's the problem that's being solved, and what's happening at each step. Here's a good example to follow:
https://github.com/Microsoft/CNTK/blob/master/Tutorials/CNTK_202_Language_Understanding.ipynb

mhamilton723:
Yes, do I need to rebuild to do that or is there some slick way to avoid it?

elibarzilay:
@mhamilton723, build fails in the e2e tests/

PR #116, opened 2017-08-23
Optimize CNTK Model

mhamilton723:
- Cache broadcasted model
- Default to byte array load

### DIFF
--- a/src/cntk-model/src/main/scala/CNTKModel.scala
+++ b/src/cntk-model/src/main/scala/CNTKModel.scala
@@ -3,11 +3,9 @@
 
 package com.microsoft.ml.spark
 
-import java.io.File
-
-import com.microsoft.CNTK.{DataType => CNTKDataType, Function => CNTKFunction, _}
+import com.microsoft.CNTK.CNTKExtensions._
+import com.microsoft.CNTK.{DataType => CNTKDataType, SerializableFunction => CNTKFunction, _}
 import com.microsoft.ml.spark.schema.DatasetExtensions
-import org.apache.commons.io.FileUtils._
 import org.apache.spark.broadcast._
 import org.apache.spark.ml.Model
 import org.apache.spark.ml.linalg.SQLDataTypes.VectorType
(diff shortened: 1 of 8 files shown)


src/cntk-model/src/main/scala/CNTKModel.scala
```diff
@@ -139,19 +125,30 @@ class CNTKModel(override val uid: String) extends Model[CNTKModel] with ComplexP
...
+  private var broadcastedModelOption: Option[Broadcast[CNTKFunction]] = None
 
   /** @group setParam */
-  def setModel(bytes: Array[Byte]): this.type = set(model, bytes)
+  def setModel(value: CNTKFunction): this.type = {
+    // Free up memory used by the previous model
+    // TODO: investigate using destroy()
+    broadcastedModelOption.foreach {m =>
+      m.unpersist()
+      print("unpersisting")
```
drdarshan: Remove?

src/cntk-model/src/main/scala/CNTKModel.scala
```diff
@@ -139,19 +125,30 @@ class CNTKModel(override val uid: String) extends Model[CNTKModel] with ComplexP
...
+  val model: CNTKFunctionParam =
+    new CNTKFunctionParam(this, "model", "Array of bytes containing the serialized CNTKModel")
+
+  private var broadcastedModelOption: Option[Broadcast[CNTKFunction]] = None
 
   /** @group setParam */
-  def setModel(bytes: Array[Byte]): this.type = set(model, bytes)
+  def setModel(value: CNTKFunction): this.type = {
+    // Free up memory used by the previous model
+    // TODO: investigate using destroy()
```
drdarshan: This worries me - shouldn't the GC handle freeing up the model?

src/cntk-model/src/main/scala/SerializableFunction.scala
```diff
@@ -0,0 +1,82 @@
...
+object CNTKUtils {
+
+  def loadModelFromBytes(bytes: Array[Byte],
+                         device: DeviceDescriptor =
+                         DeviceDescriptor.useDefaultDevice): Function = {
+    try {
+      Function.load(bytes, device)
+    } catch {
+      case _: IllegalArgumentException =>
+        println("WARNING: model is a legacy model," +
```
drdarshan: Consider using Log4J warnings so the user can control/disable them.

PR #128, opened 2017-09-08
Notebook style scan and verification scripts

elibarzilay:
Some additional style improvements.

### DIFF
--- a/notebooks/samples/301 - CIFAR10 CNTK CNN Evaluation.ipynb
+++ b/notebooks/samples/301 - CIFAR10 CNTK CNN Evaluation.ipynb
@@ -4,15 +4,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## 301 - Ingesting CIFAR Images into Spark DataFrames and Evaluating Pre-Trained CNTK Models "
+    "## 301 - Ingesting CIFAR Images into Spark DataFrames and Evaluating Pre-Trained CNTK Models"
    ]
   },
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": false
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "from mmlspark import CNTKModel, ModelDownloader\n",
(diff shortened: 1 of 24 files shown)


notebooks/samples/301 - CIFAR10 CNTK CNN Evaluation.ipynb
```diff
@@ -183,7 +174,7 @@
    "execution_count": null,
    "metadata": {
     "mml-deploy": "hdinsight",
-    "collapsed": false
+    "collapsed": true
```
mhamilton723: why are these collapsed now?

tools/misc/check-notebooks.js
```diff
@@ -0,0 +1,137 @@
...
+              {cell_type: "code", execution_count: null,
+               metadata: _cellMetadata, outputs: [],
+               source: _srcStrings})),
+   metadata: _obj({"anaconda-cloud": _obj({}),
+                   kernelspec: _obj({display_name: "Python [default]",
+                                     language: "python",
+                                     name: "python3"}),
+                   language_info: _}),
+   nbformat: 4,
+   nbformat_minor: 0});
```
mhamilton723: Will this just block notebooks if they have funky fields? If so, is there a way to make it a script that normalizes it to take the tedium out of it?

notebooks/samples/301 - CIFAR10 CNTK CNN Evaluation.ipynb
elibarzilay: I made the verification thing enforce a rule that `mml-deploy` with a `local` has `false` and with `hdinsight` it should have `true`.  The idea is to see just the local version which is more presentable since it doesn't have the hdinsight hacks.  I'm not sure about it though, so that's a question for @drdarshan whether there is a technical reason to change this.

tools/misc/check-notebooks.js
elibarzilay: @mhamilton723, no -- it's a script that I'll use to verify and fix new notebooks when they come in, since I got tired of doing the fixes manually and missing things over and over again...  It'll also be useful in making changes to the normalized versions in case of changes -- eg, if @drdarshan thinks that `collapsed` should not be changed, then it'll be easy to do that over all notebooks quickly.

drdarshan: We can just use nbformat to do all this. I'm not sure adding a dependency on node is really needed.

PR #131, opened 2017-09-10
Optimize OpenCV native loading

mhamilton723:
- The code previously loaded the natives for each partition which took a lot of memory, it now loads it once per machine
- This makes the build approx 15 min faster :)

### DIFF
--- a/src/core/env/src/main/scala/NativeLoader.java
+++ b/src/core/env/src/main/scala/NativeLoader.java
@@ -25,16 +25,15 @@ import java.nio.file.Files;
  * the libraries will be loaded in the order specified in the manifest. The native libraries should be
  * in folders describing the OS they run on: linux, windows, mac. </p>
  * */
-public class NativeLoader {
+public class NativeLoader implements Serializable {
 
     private String resourcesPath;
     private Boolean extractionDone = false;
     private File tempDir;
-    private Object lock = new Object();
 
-    public NativeLoader(String topLevelResourcesPath) throws IOException{
+    public NativeLoader(String topLevelResourcesPath) throws IOException {
         this.resourcesPath = getResourcesPath(topLevelResourcesPath);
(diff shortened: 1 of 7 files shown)


src/core/env/src/main/scala/NativeLoader.java
```diff
@@ -25,16 +25,15 @@
...
 
     private String resourcesPath;
     private Boolean extractionDone = false;
     private File tempDir;
-    private Object lock = new Object();
 
     public NativeLoader(String topLevelResourcesPath) throws IOException{
         this.resourcesPath = getResourcesPath(topLevelResourcesPath);
-        tempDir = Files.createTempDirectory("tmp").toFile();
+        tempDir = Files.createTempDirectory("mml-natives").toFile();
```
elibarzilay: Whew -- thanks!!  It'll help when I clear out future messes on the build machine.

mhamilton723: yeah having that be the same prefix as cntk's natives was a fat 0

src/core/schema/src/main/scala/ImageSchema.scala
```diff
@@ -35,9 +34,18 @@ object ImageSchema {
   def isImage(df: DataFrame, column: String): Boolean =
     df.schema(column).dataType == columnSchema
 
-  private[spark] def loadLibraryForAllPartitions[T:ClassTag](rdd: RDD[T], lib: String):RDD[T] = {
-    def perPartition(it: Iterator[T]):Iterator[T] = {
-      new NativeLoader("/nu/pattern/opencv").loadLibraryByName(lib); it }
-    rdd.mapPartitions(perPartition, preservesPartitioning = true)
+  /**
```
elibarzilay: @mhamilton723 This is a private thing, so shouldn't it be a simple comment rather than a documentation one?  (Tell me if so and I'll fix it.)

PR #135, opened 2017-09-19
Changed save code to not modify the current working directory by removing the code that disposed the current file system

imatiach-msft:
This PR fixes the issue here:
https://github.com/Azure/mmlspark/issues/133
It looks like disposing the file system object was removing the currently set working directory and all other temporary settings, and a new file system object was getting created from the Hadoop configuration.

### DIFF
--- a/src/core/serialize/src/main/scala/Serializer.scala
+++ b/src/core/serialize/src/main/scala/Serializer.scala
@@ -83,10 +83,8 @@ object Serializer {
   def writeToHDFS[O](sc: SparkContext, obj: O, outputPath: Path, overwrite: Boolean)
                     (implicit ttag: TypeTag[O]): Unit = {
     val hadoopConf = sc.hadoopConfiguration
-    using(outputPath.getFileSystem(hadoopConf)) { fs =>
-      using(fs.create(outputPath, overwrite)) { os =>
-        write[O](obj, os)(ttag)
-      }.get
+    using(outputPath.getFileSystem(hadoopConf).create(outputPath, overwrite)) { os =>
+      write[O](obj, os)(ttag)
     }.get
   }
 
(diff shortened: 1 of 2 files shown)


src/core/serialize/src/main/scala/Serializer.scala
```diff
@@ -140,6 +140,7 @@ object Serializer {
     val modelPath = new Path(path)
     val hadoopConf = sc.hadoopConfiguration
     // Note: to get correct working dir, must use root path instead of root + part
+    // Also, do not dispose file system in case user modifier workdir
```
akshaya-a: typo in a couple places :)

src/train-classifier/src/test/scala/VerifyTrainClassifier.scala
```diff
@@ -121,22 +130,34 @@ class VerifyTrainClassifier extends EstimatorFuzzingTest {
...
-    lazy val dir = new File(myModelName)
+    lazy val dir = new File(subfolder + "/" + myModelName)
     try {
       model.write.overwrite().save(myModelName)
       // write a second time with overwrite flag, verify still works
       model.write.overwrite().save(myModelName)
       // assert directory exists
       assert(dir.exists())
 
+      val fileSystem = getFileSystem
```
akshaya-a: Awesome thanks for the quick turnaround!

src/core/serialize/src/main/scala/Serializer.scala
imatiach-msft: done, fixed

src/train-classifier/src/test/scala/VerifyTrainClassifier.scala
imatiach-msft: Thanks!

PR #136, opened 2017-09-20
Updating CNTK version in script action to 2.2

imatiach-msft:
Updating CNTK version in script action to 2.2


RatanRSur:
This should update it to 2.1 to match the CNTK received from maven central. Once 2.2 is up both the dependency on maven central and the script action can be bumped in parallel

elibarzilay:
Closing: superseded by #139 which makes the HDI scripts use the same `$CNTK_VERSION` so there's a single version.  Now we need to wait until the cntk maven package for 2.2 is up.

PR #138, opened 2017-09-22
Fix string column preprocessing for sample notebook 102

eedeleon:


imatiach-msft:
Nice catch!  Signed off.

eedeleon:
Thanks!

PR #143, opened 2017-09-29
Add a pipelineStage for calculating class balancing weights

mhamilton723:
- This pipeline stages calculates weights for estimators like LogisticRegression, so that one can correct for unbalanced datasets

### DIFF
--- a/src/pipeline-stages/src/main/scala/ClassBalancer.scala
+++ b/src/pipeline-stages/src/main/scala/ClassBalancer.scala
@@ -1,0 +1,81 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.param.{BooleanParam, ParamMap}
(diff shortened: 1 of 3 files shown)


src/pipeline-stages/src/main/scala/ClassBalancer.scala
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.spark.ml.util.{DefaultParamsReadable, DefaultParamsWritable, Identifiable}
+import org.apache.spark.ml.{Estimator, Model}
+import org.apache.spark.sql.functions.{col, lit, max}
+import org.apache.spark.sql.types.{DoubleType, StructType}
+import org.apache.spark.sql.{DataFrame, Dataset}
+import org.apache.spark.sql.functions.broadcast
+
+import scala.reflect.runtime.universe.{TypeTag, typeTag}
+
+class ClassBalancer(val uid: String) extends Estimator[ClassBalancerModel]
```
drdarshan: Documentation please? :-)

mhamilton723:
Logistic Regression in spark supports these weights. Also plan on making one that is classifier independent by augmenting the dataset, but that is a slower and less reliable solution when compared to estimators that naturally support weighting.

PR #147, opened 2017-10-10
Add file name to notebook JSON decoding errors

eedeleon:
When there is any error message in the loading of notebooks from notebooks/samples/ the error message was related to the contents of the file instead of the file name.

The change improves the error message when failing to Decode notebook JSON by adding the file name to the error message.

### DIFF
--- a/tools/notebook/postprocess.py
+++ b/tools/notebook/postprocess.py
@@ -90,7 +90,15 @@ def postprocess_notebooks(input_dir, output_base_dir):
     import os
     import glob
     from nbformat import read, write, NO_CONVERT
-    notebooks = [(os.path.split(nbfile)[-1], read(nbfile, NO_CONVERT))
+
+    def _read(nbfile, to_conv):
+        try:
+            return read(nbfile, to_conv)
+        except Exception as e:
+            e.args += ("File name: {}".format(os.path.split(nbfile)[-1]),)
+            raise e
+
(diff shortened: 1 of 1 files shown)


tools/notebook/postprocess.py
```diff
@@ -90,7 +90,15 @@ def postprocess_notebooks(input_dir, output_base_dir):
     import os
     import glob
     from nbformat import read, write, NO_CONVERT
-    notebooks = [(os.path.split(nbfile)[-1], read(nbfile, NO_CONVERT))
+
+    def _read(nbfile, to_conv):
+        try:
+            return read(nbfile, to_conv)
+        except ValueError as e:
+            e.args += ("File name: {}".format(os.path.splut(nbfile)[-1]),)
```
RatanRSur: splut?

tools/notebook/postprocess.py
```diff
@@ -90,7 +90,15 @@ def postprocess_notebooks(input_dir, output_base_dir):
     import os
     import glob
     from nbformat import read, write, NO_CONVERT
-    notebooks = [(os.path.split(nbfile)[-1], read(nbfile, NO_CONVERT))
+
+    def _read(nbfile, to_conv):
+        try:
+            return read(nbfile, to_conv)
+        except ValueError as e:
```
drdarshan: Might as well catch ``Exception as e`` since you might run into errors besides just ``ValueError`` :-)

eedeleon: Makes sense. The added information would be relevant regardless.

PR #149, opened 2017-10-12
gpu training with parquet reader

imatiach-msft:
Adding GPU training with Parquet reader

### DIFF
--- a/src/cntk-train/src/main/scala/BrainscriptBuilder.scala
+++ b/src/cntk-train/src/main/scala/BrainscriptBuilder.scala
@@ -22,10 +22,17 @@ class BrainScriptBuilder {
 
   var rootDir: String = ""
   var outDir: String = ""
-  var weightPrecision: String = "float"
+  var weightPrecision: String = CNTKLearner.floatPrecision
 
   var commands = ListBuffer[String]("trainNetwork")
   var testModel = false
+
+  var activeNameNode: String = ""
+
+  def setWeightPrecision(precision: String): this.type = {
+    weightPrecision = precision
+    this
+  }
 
   def setInputFile(path: String, format: String, shapes: Map[String, InputShape]): this.type = {
     inData = Some(InputData(format, path, shapes))
@@ -34,6 +41,13 @@ class BrainScriptBuilder {
 
   def setModelName(n: String): this.type = {
     modelName = n
+    this
+  }
+
+  def setHDFSPath(hdfsPath: Option[(String, String, String)]): this.type = {
+    if (hdfsPath.isDefined) {
+      activeNameNode = hdfsPath.get._3
+    }
     this
   }
 
@@ -74,13 +88,39 @@ class BrainScriptBuilder {
     this
   }
 
+  def textReaderConfig(loc: String, ipstring: String): String = {
+    s"""reader = [ readerType = CNTKTextFormatReader; file = "$loc" ; $ipstring ]"""
+  }
+
+  def parquetReaderConfig(loc: String, ipstring: String): String = {
+    s"""    reader = {
+      |        deserializers = (
+      |                        [
+      |                            type = "DataFrameDeserializer"
+      |                            module = "Cntk.Deserializers.DF"
+      |                            $ipstring
+      |                            hdfs = {
+      |                                host = "$activeNameNode";
+      |                                port = "${CNTKLearner.rpcPortNumber}";
+      |                                filePath = "$loc";
+      |                                format = "Parquet"
+      |                            }
+      |                        ]
+      |                    )
+      |        prefetch = false
+      |        randomize = true
+      |        keepDataInMemory = true     # cache all data in memory
+      |    }
+    """.stripMargin
+  }
+
   def toReaderConfig: String = {
     val ipstring = getInputString
     val loc = inData.get.path
-    val form = inData.get.format match {
-      case "text" => "CNTKTextFormatReader"
+    inData.get.format match {
+      case CNTKLearner.textDataFormat => textReaderConfig(loc, ipstring)
+      case CNTKLearner.parquetDataFormat => parquetReaderConfig(loc, ipstring)
     }
-    s"""reader = [ readerType = $form ; file = "$loc" ; $ipstring ]"""
   }
 
   def toOverrideConfig: Seq[String] = {
--- a/src/cntk-train/src/main/scala/CNTKLearner.scala
+++ b/src/cntk-train/src/main/scala/CNTKLearner.scala
@@ -16,43 +16,42 @@ import org.apache.spark.sql._
 import org.apache.spark.sql.types._
 
 trait CNTKParams extends MMLParams {
-
-  // This is only needed until Train* accepts CNTKLearner instead of CL acting like Train*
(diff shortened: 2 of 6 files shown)


src/cntk-train/src/main/scala/CommandBuilders.scala
```diff
@@ -181,8 +184,9 @@ class MPICommandBuilder(log: Logger,
         Seq("scp", "-i", identity, "-r", "-o", "StrictHostKeyChecking=no",
           fileInputPath, s"$gpuUser:$fileInputPath"))
     }
+    val runMPI = "''" + s"$exportClasspath $command $mpiArgs $cntkArgs" + "''"
     // Run the mpi command
-    printOutput(Seq("ssh", "-i", identity, gpuUser, command, mpiArgs, cntkArgs, "parallelTrain=true"))
+    printOutput(Seq("ssh", "-i", identity, gpuUser, runMPI))
```
tongwen11: Just want to make sure we are not losing "parallelTrain=true"

src/cntk-train/src/main/scala/DataConversion.scala
```diff
@@ -106,6 +103,74 @@ object DataTransferUtils {
...
+      }
+    // Note: order matters, features column needs to come first
+    val cols = data.select(feats, label).columns.flatMap(name => {
+      if ((name == label && labelForm == CNTKLearner.denseForm) ||
+        (name == feats && featureForm == CNTKLearner.denseForm)) {
+        Seq(toDenseArray(data.col(name)).alias(name))
+      } else {
+        Seq(getNZs(data.col(name)).alias(name + "_size"),
+          getIdxes(data.col(name)).alias(name + "_indices"),
+          getVals(data.col(name)).alias(name) + "_values")
```
tongwen11: Are we missing a closing ")" to match "Seq("

imatiach-msft: good catch, the string:
+ "_values"
needs to be inside alias() I believe

tongwen11:
Ilya,

Do you want to make add the top-level try-catch for the final cleanup in this check-in or a separate one?

Another touch up item is to handle cases where the DF passed to CNTKLearner has columns which are not Vector type.

src/cntk-train/src/main/scala/CommandBuilders.scala
imatiach-msft: it didn't seem to make a difference for me, I believe it's already in the override config file

imatiach-msft:
I can make the top level try catch in a separate checkin

src/cntk-train/src/main/scala/BrainscriptBuilder.scala
```diff
@@ -74,13 +88,39 @@ class BrainScriptBuilder {
...
+  }
+
+  def parquetReaderConfig(loc: String, ipstring: String): String = {
+    s"""    reader = {
+      |        deserializers = (
+      |                        [
+      |                            type = "DataFrameDeserializer"
+      |                            module = "Cntk.Deserializers.DF"
+      |                            $ipstring
+      |                            hdfs = {
```
drdarshan: Nice! is the port number fixed or does it need to be made a parameter.

src/cntk-train/src/main/scala/CNTKLearner.scala
```diff
@@ -16,12 +16,10 @@ import org.apache.spark.sql._
...
 
 trait CNTKParams extends MMLParams {
-
-  // This is only needed until Train* accepts CNTKLearner instead of CL acting like Train*
-  val labelsColumnName = StringParam(this, "labelsColumnName", "Label col", "labels")
+  val labelsColumnName = StringParam(this, "labelsColumnName", "label column name", "labels")
   def getLabelsColumnName: String = $(labelsColumnName)
 
-  val featuresColumnName = StringParam(this, "featuresColumnName", "feats col", "features")
+  val featuresColumnName = StringParam(this, "featuresColumnName", "features column name", "features")
```
drdarshan: Sentence case for docs please :-)

src/cntk-train/src/main/scala/CommandBuilders.scala
```diff
@@ -161,14 +165,13 @@ class MPICommandBuilder(log: Logger,
 
     // Run commands below only if writing input data to HDFS
     var fileDirStr: String = ""
-    if (hdfsPath.isDefined) {
+    if (hdfsPath.isDefined && dataFormat == CNTKLearner.textDataFormat) {
```
drdarshan: I think the HDFS mounting is orthogonal to text format vs. Parquet. At some point, we can add libhdfs support to the CNTK text reader for example.

src/cntk-train/src/main/scala/BrainscriptBuilder.scala
imatiach-msft: it needs to be fixed according to Tong

src/cntk-train/src/main/scala/CNTKLearner.scala
imatiach-msft: done, Eli has already modified the other params

src/cntk-train/src/main/scala/BrainscriptBuilder.scala
tongwen11: We can create a constant like CNTKLearner.rpcPortNumber to represent RPC protocol port number and use it to replace the port number in the template.

src/cntk-train/src/main/scala/CommandBuilders.scala
```diff
@@ -114,10 +120,8 @@ class MPICommandBuilder(log: Logger,
...
   def runCommand(): Unit = {
-    val pathEnvVar = "PATH=/usr/bin/cntk/cntk/bin:$PATH"
-    val ldLibraryPathEnvVar =
-      "LD_LIBRARY_PATH=/usr/bin/cntk/cntk/lib:/usr/bin/cntk/cntk/dependencies/lib:$LD_LIBRARY_PATH"
-    val mpiArgs = s" -x $pathEnvVar -x $ldLibraryPathEnvVar --npernode ${nodeConfig.head._2} "
+    val exportClasspath = "export CLASSPATH=$($HADOOP_HOME/bin/hadoop classpath --glob); "
+    val mpiArgs = s" --npernode ${nodeConfig.head._2} "
     val cntkArgs = "cntk " + configs
       .map(c => if (fileBased) s"configFile=${configToFile(c)} " else c.text.mkString(" "))
       .mkString(" ")
```
drdarshan: Please hoist the location of the SSH key pair into a constant somewhere. We might want to make it configurable.

src/cntk-train/src/main/scala/BrainscriptBuilder.scala
imatiach-msft: done

src/cntk-train/src/main/scala/CommandBuilders.scala
imatiach-msft: done

imatiach-msft:
I added the top level try/finally based on Tong's recommendation and updated the code to use several constants

PR #154, opened 2017-10-16
Properly highlight scala code in README

Utsav2:


elibarzilay:
Merged with slight modifications.

Note: the CLA thing was triggered because of these modifications, I overrode it because they are my modifications...

PR #156, opened 2017-10-16
Update MultiColumnAdapter to support Estimators

eedeleon:


src/multi-column-adapter/src/main/scala/MultiColumnAdapter.scala
```diff
@@ -4,67 +4,74 @@
...
+  }
+
+  /** Base pipeline stage to apply to every column in the input column list.
     * @group param
     */
-  val baseTransformer: TransformerParam =
-    new TransformerParam(this,
-                         "baseTransformer",
-                         "base transformer to apply to every column")
+  val basePipelineStage: PipelineStageParam =
```
mhamilton723: nit: perhaps rename to just stage, this is what the pipeline and timer APIs look like

src/multi-column-adapter/src/test/scala/MultiColumnAdapterSpec.scala
```diff
@@ -3,47 +3,65 @@
...
 import com.microsoft.ml.spark.schema.DatasetExtensions._
-import org.apache.spark.ml.Transformer
+import org.apache.spark.ml.{Estimator, PipelineModel}
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.types.StructType
 
-class MultiColumnAdapterSpec extends TransformerFuzzingTest {
+import scala.collection.mutable
 
+class MultiColumnAdapterSpec extends EstimatorFuzzingTest {
```
mhamilton723: This should extend a different fuzzer. do a git fetch upstream then git rebase upstream/master to see

src/multi-column-adapter/src/test/scala/MultiColumnAdapterSpec.scala
```diff
@@ -3,47 +3,65 @@
...
+    assert(lines1 === trueLines1)
 
-  override def createDataset: DataFrame = wordDF
+    val lines2 = tokenizedDF.getColAs[Array[String]]("output2")
+    val trueLines2 = mutable.ArraySeq(1, 0, 0, 2)
+    assert(lines2 === trueLines2)
+  }
 
-  override def schemaForDataset: StructType = ???
+  override def createTransformDataset: DataFrame = wordDF
```
mhamilton723: These will become unecessary with the new fuzzer

src/multi-column-adapter/src/main/scala/MultiColumnAdapter.scala
eedeleon: In the Pipeline, there is a distinction with stages. My only concern is that we use this param to initialize stages, it is not used as a stage itself. baseStage may take this into account more so than stage. What do you think?

src/codegen/src/main/scala/PySparkWrapperTest.scala
```diff
@@ -101,6 +101,10 @@ abstract class PySparkWrapperTest(entryPoint: PipelineStage,
           |        model = my$entryPointName.fit(data)
           |        self.assertNotEqual(model, None)""".stripMargin
 
+  protected def tryMultiColumnFitTemplate(entryPointName: String, model: String) =
```
mhamilton723: Could you explain why this code is needed? Shouldnt the code ion the case statement be sufficient? Thanks

src/featurize/src/main/scala/AssembleFeatures.scala
```diff
@@ -442,12 +442,13 @@ class AssembleFeaturesModel(val uid: String,
         val ohe = new OneHotEncoder()
         val inputColsKeys = columnNamesToFeaturize.categoricalColumns.keys
         val outputColsKeys = columnNamesToFeaturize.categoricalColumns.values
-        val inputCols = inputColsKeys.mkString(",")
-        val outputCols = outputColsKeys.mkString(",")
+        val inputCols = inputColsKeys.toArray
```
mhamilton723: thank god :)

src/multi-column-adapter/src/test/scala/MultiColumnAdapterSpec.scala
eedeleon: Done

src/multi-column-adapter/src/test/scala/MultiColumnAdapterSpec.scala
eedeleon: Done, it was great to delete some code

src/codegen/src/main/scala/PySparkWrapperTest.scala
eedeleon: The model param does not exist for MultiColumnAdapter. One option was to change tryTemplate to take "model" or "baseStage" as input and handle it accordingly. I decided this fix would avoid short term errors more easily, and since I am pushing to rewrite this portion of the code base with Maureen and Eli, that would be preferred.

eedeleon: Thanks for the snappy review!

PR #161, opened 2017-10-20
Fixing reader to display full path to image

imatiach-msft:
Fixing image reader so that path in schema would contain the full path to image (or file) instead of just the zip file and added a test to validate

### DIFF
--- a/src/readers/src/main/scala/KeyValueReaderIterator.scala
+++ b/src/readers/src/main/scala/KeyValueReaderIterator.scala
@@ -1,0 +1,54 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package org.apache.spark.binary
+
+import java.io.Closeable
+import org.apache.hadoop.mapreduce.RecordReader
+
+// Based on:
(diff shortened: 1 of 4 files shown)


src/readers/src/main/scala/KeyValueReaderIterator.scala
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
```
drdarshan: I'm still a bit concerned that we have to copy-paste ~50 lines from Spark in order to return a tuple instead of a single value.
In any event, I would suggest renaming the type params K and V instead of K and T since that seems more idiomatic to code I've seen.

imatiach-msft: done

elibarzilay:
@imatiach-msft: Merging this with a comment pointing at the original code.

elibarzilay:
@imatiach-msft: the build fails with

    ##[error]Executor: Exception in task 5.0 in stage 22.0 (TID 81)
    java.lang.ClassCastException: org.apache.hadoop.io.BytesWritable cannot be cast to [B
      at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow$class.getBinary(rows.scala:47)
      ...

PR #165, opened 2017-10-23
Fixes to the beautified gpu vm setup script

tongwen11:
- syntax fix to case-in
- zulu can only be installed with "apt-get" and with "apt-get update"
- Add back the LD_LIBRARY_PATH setting for MPI


elibarzilay:
Closing, will force-push a re-re-updated version of the script.

PR #168, opened 2017-10-25
Adding randomized grid search hyperparameter tuner + tests

imatiach-msft:
Adding randomized grid search hyperparameter tuner + tests
Resolves https://github.com/Azure/mmlspark/issues/14
Also resolves the JIRA:
https://issues.apache.org/jira/browse/SPARK-18755


src/core/test/benchmarks/src/main/scala/Benchmarks.scala
```diff
@@ -0,0 +1,77 @@
...
+
+abstract class Benchmarks extends TestBase {
+  val moduleName: String
+  val targetDirectory = new File("target")
+  val thisDirectory = {
+    // intellij runs from a different directory
+    val d = new File("src/test/scala")
+    if (d.isDirectory) d else new File(s"$moduleName/src/test/scala")
+  }
+  assert(thisDirectory.isDirectory, "-- the test should run in the sub-project root level")
```
mhamilton723: Can we try to get rid of this dependency, it comes back to haunt us in fuzzing

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def compareBenchmarkFiles(): Unit = {
+    try writeFile(benchmarkMetricsFile, accuracyResults.mkString("\n") + "\n")
+    catch {
+      case e: java.io.IOException => throw new Exception("Not able to process benchmarks file")
+    }
+    val historicMetrics = readFile(historicMetricsFile, _.getLines.toList)
+    if (historicMetrics.length != accuracyResults.length)
+      throw new Exception(s"Mis-matching number of lines in new benchmarks file: $benchmarkMetricsFile")
+    for (((hist,acc),i) <- (historicMetrics zip accuracyResults).zipWithIndex) {
```
mhamilton723: Might want to use json parser for this work to eliminate this handmade text file parsing

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
```diff
@@ -26,6 +26,7 @@ class FuzzingTest extends TestBase {
     val exemptions: Set[String] = Set(
       "com.microsoft.ml.spark.TrainClassifier",
       "com.microsoft.ml.spark.TrainRegressor",
+      "com.microsoft.ml.spark.TuneHyperparameters",
```
mhamilton723: Can we try to get this passing the fuzzing tests? It will be easier if you dont have the directory running issue

src/find-best-model/src/main/scala/EvaluationUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+    }
+    (evaluationMetricColumnName, operator)
+  }
+
+  def getModelParams(model: Transformer): ParamMap = {
+    model match {
+      case reg: TrainedRegressorModel => reg.getParamMap
+      case cls: TrainedClassifierModel => cls.getParamMap
+      case evm: BestModel => getModelParams(evm.getBestModel)
+      case _ => throw new Exception("Model type not supported for evaluation")
```
mhamilton723: Can we try to generalize this to all sparkml estimators using the existing traits

src/tune-hyperparameters/src/main/scala/HyperparamBuilder.scala
```diff
@@ -0,0 +1,81 @@
...
+    * Adds a param to the search space.
+    */
+  def addHyperparam[T](param: Param[T], values: HyperParam[T]): this.type = {
+    hyperparams.put(param, values)
+    this
+  }
+
+  /**
+    * Builds the search space of hyperparameters.
+    * @return The map of hyperparameters to search through.
```
mhamilton723: This is a good start, we should think about this problem as sampling from distributions over different domains. In this light these would all be uniform distributions see:

http://scikit-learn.org/stable/modules/grid_search.html#randomized-parameter-search

For an example of a good API

src/tune-hyperparameters/src/main/scala/Hyperparams.scala
```diff
@@ -0,0 +1,88 @@
...
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.classification._
+import org.apache.spark.ml.param.Param
+
+/**
+  * Provides good default hyperparameter ranges and values for sweeping.
+  * Publicly visible to users so they can easily select the parameters for sweeping.
+  */
+object Hyperparams {
```
mhamilton723: Is it possible to use the tuner on other models than these?

src/tune-hyperparameters/src/main/scala/TuneHyperparameters.scala
```diff
@@ -0,0 +1,200 @@
...
+      val evaluator = new ComputeModelStatistics()
+      evaluator.set(evaluator.evaluationMetric, getEvaluationMetric)
+
+      // Evaluate the models
+      val foldMetricFutures = modelFutures.zip(modelFuturesParams).map { case (modelFuture, paramMap) =>
+        modelFuture.map { model =>
+          val scoredDataset = model.transform(validationDataset, paramMap)
+          val metrics = evaluator.transform(scoredDataset)
+          val metric = metrics.select(evaluationMetricColumnName).first()(0).toString.toDouble
+          logDebug(s"Got metric $metric for model trained with $paramMap.")
```
mhamilton723: We should just use the cross validator from spark with a custom set of randomly sampled params

https://spark.apache.org/docs/2.2.0/ml-tuning.html

imatiach-msft: the cross validator does not use our model stats transform, which is faster, pipelineable and supports more metrics (and has better logging).  It also does not allow the user to specify parallelism in the 2.1 release, doesn't allow the user to specify multiple estimators (only a single one) and doesn't take in hyperparams.  I also plan to add more tricks to choosing the next parameter which wouldn't be possible if we called spark's crossvalidator directly.

src/tune-hyperparameters/src/main/scala/Hyperparams.scala
imatiach-msft: Yes!  These are just defaults that the user can choose conveniently.  The user can build any hyperparams they wish.

src/tune-hyperparameters/src/main/scala/TuneHyperparameters.scala
imatiach-msft: We can also add more termination criteria besides numRuns in the future, which won't be possible if we use their cross validator

src/tune-hyperparameters/src/main/scala/HyperparamBuilder.scala
imatiach-msft: Yes!  I want to add more distributions in the future.  It can be put on the hyperparams class constructor.

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
imatiach-msft: can you discuss this with Eli?  I originally output the data to csv in TrainClassifier, but he preferred this way.

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
imatiach-msft: sure, the directory issue is independent of the fuzzing tests

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
elibarzilay: I definitely think that JSON is an overkill for simple cases like this...

src/find-best-model/src/main/scala/EvaluationUtils.scala
imatiach-msft: done!

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
imatiach-msft: (I believe this is resolved - let me know if it still needs to be taken care of and we can discuss)

src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
```diff
@@ -113,11 +109,50 @@ class ComputeModelStatistics(override val uid: String) extends Transformer with
   /** @group setParam */
   def setEvaluationMetric(value: String): this.type = set(evaluationMetric, value)
 
-  lazy val logger = Logger.getLogger(this.getClass.getName)
+  // Parameter required only is using native Spark ML Estimator instead of MMLSpark Estimator
+  val modelName: Param[String] =
+    StringParam(this, "modelName", "Model name, only required if using SparkML estimators", "SparkML")
```
mhamilton723: Using a String type to represent an object this is brittle, could you use use a pipeline stage param?

src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
```diff
@@ -113,11 +109,50 @@ class ComputeModelStatistics(override val uid: String) extends Transformer with
...
+
+  def getModelName: String = $(modelName)
+
+  def setModelName(value: String): this.type = set(modelName, value)
+
+  /** Parameter required only is using native Spark ML Estimator instead of MMLSpark Estimator
+    * Set to true if computing metrics for the output of a classification model, set to false if regression model.
+    */
+  val classifier: Param[Boolean] =
+    BooleanParam(this, "classifier", "Is classifier scored data, only required if using SparkML estimators", true)
```
mhamilton723: This would not be needed if the above parameter was a pipeline stage because you could case match

src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
```diff
@@ -113,11 +109,50 @@ class ComputeModelStatistics(override val uid: String) extends Transformer with
...
+
+  def getClassifier: Boolean = $(classifier)
+
+  def setClassifier(value: Boolean): this.type = set(classifier, value)
+
+  val scoredLabelsCol =
+    StringParam(this, "scoredLabelsCol", "Scored labels column name, only required if using SparkML estimators")
```
mhamilton723: Could you elaborate on the differences between scores and scoredLabels? I'm having a hard time understanding why these need to both be specified

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
mhamilton723: Still seems to be in the file, could you remove it? Thanks!

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
mhamilton723: CSVs allow users to manipulate these in more complicated ways down the line. Its simple now but in the future someone might want to build on top of this. Dont depend on any huge parsing libraries, but if you can find a lightweight csv reader and writer (or use spark's!) then that would be much easier to work with and understand

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
mhamilton723: Can you remove this line?

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
```diff
@@ -67,6 +68,7 @@ class FuzzingTest extends TestBase {
...
```
mhamilton723: THis one too

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
```diff
@@ -112,6 +114,7 @@ class FuzzingTest extends TestBase {
...
```
mhamilton723: this one too

src/find-best-model/src/main/scala/EvaluationUtils.scala
```diff
@@ -0,0 +1,128 @@
...
+    *     - r2
+    *     - mae
+    *
+    *   Or, for either type of classifier:
+    *     - all - This will report all the relevant metrics
+    *
+    * @group param
+    */
+  val evaluationMetric: Param[String] = StringParam(this, "evaluationMetric", "Metric to evaluate models with",
+    (s: String) => Seq(ComputeModelStatistics.MseSparkMetric,
```
mhamilton723: Nit: Use a set

src/find-best-model/src/main/scala/EvaluationUtils.scala
```diff
@@ -0,0 +1,128 @@
...
+  def getEstimatorModelType(model: Estimator[_]): String = {
+    model match {
+      case _: TrainRegressor => SchemaConstants.RegressionKind
+      case _: TrainClassifier => SchemaConstants.ClassificationKind
+      case _: Classifier[_, _, _] => SchemaConstants.ClassificationKind
+      case _: LinearRegression => SchemaConstants.RegressionKind
+      case _: DecisionTreeRegressor => SchemaConstants.RegressionKind
+      case _: GBTRegressor => SchemaConstants.RegressionKind
+      case _: RandomForestRegressor => SchemaConstants.RegressionKind
+      case _ => throw new Exception(modelTypeUnsupportedErr)
```
mhamilton723: Is there any way to find a cleaner way to figure this out for regression?

src/find-best-model/src/main/scala/EvaluationUtils.scala
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  def getMetricWithOperator(model: Transformer, evaluationMetric: String): (String, Ordering[Double]) = {
+    val modelType = getTransformerModelType(model)
```
mhamilton723: Both of these should be in one function that takes in pipeline stages and case matches internally, likewise for getXModelType

src/tune-hyperparameters/src/main/scala/Hyperparams.scala
mhamilton723: Okay can we rename this object to something like DefaultHyperparameters then?

src/tune-hyperparameters/src/main/scala/Hyperparams.scala
```diff
@@ -0,0 +1,88 @@
...
+  /**
+    * Defines the default hyperparameter range for logistic regression.
+    * @return The hyperparameter search space for logistic regression.
+    */
+  def logisticRegression(learner: LogisticRegression): Array[(Param[_], Dist[_])] = {
```
mhamilton723: All of this should be a case match with a single function as interface, or of these functions should have the same name, scala overrides will kick in from there.

src/tune-hyperparameters/src/main/scala/TuneHyperparameters.scala
```diff
@@ -0,0 +1,221 @@
...
+          evaluator.set(evaluator.evaluationMetric, getEvaluationMetric)
+          model match {
+            case _: TrainedRegressorModel => {
+              logDebug("Evaluating trained regressor model.")
+            }
+            case _: TrainedClassifierModel => {
+              logDebug("Evaluating trained classifier model.")
+            }
+            case classificationModel: ClassificationModel[_, _] => {
+              logDebug(s"Evaluating SparkML ${model.uid} classification model.")
```
mhamilton723: This kind of structure is what worries me, there are very large case matches throughout, which points to the need to re-abstract some of this

src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
imatiach-msft: agreed, it is brittle, but it's the only way to make the case work when bringing in scored data outside of spark

src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
imatiach-msft: This is applicable in classification.  The score is the raw value or probability.  The scored label is the "label" that is assigned based on the score, eg if you had a probability of 0.9 to determine whether the image is of a dog (1) or cat (0), the label would be "dog"

src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
imatiach-msft: I've removed the param now

src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
imatiach-msft: I've removed the param now

src/find-best-model/src/main/scala/EvaluationUtils.scala
imatiach-msft: done

src/find-best-model/src/main/scala/EvaluationUtils.scala
imatiach-msft: unfortunately regressor is private to spark (in Regressor.scala in spark ML):
private[spark] abstract class Regressor
but, for some bizarre reason, Classifier is not (in Classifier.scala in spark ML):
abstract class Classifier
Do you see any way to make it cleaner?

src/tune-hyperparameters/src/main/scala/Hyperparams.scala
imatiach-msft: done!

src/tune-hyperparameters/src/main/scala/Hyperparams.scala
imatiach-msft: renamed all functions to "defaultRange" so they would have the same name

src/tune-hyperparameters/src/main/scala/TuneHyperparameters.scala
imatiach-msft: how can I improve the case matches?  It seems like it is necessary in some cases.

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
imatiach-msft: can you discuss this more with Eli?  I don't have one specific preference or another.  Originally I wrote it with CSV read/write but Eli changed the code to check raw files output to disk, and I liked his new approach.  I could try to go back to my original changes, it would be an easy change, but I think Eli's approach is actually nicer.

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
imatiach-msft: removed

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
imatiach-msft: for the python tests I need to add a corresponding parameter space python file first

src/find-best-model/src/main/scala/EvaluationUtils.scala
imatiach-msft: done!

src/find-best-model/src/main/scala/EvaluationUtils.scala
imatiach-msft: I made the private regressor function in spark ML public through a method in the same namespace

src/compute-model-statistics/src/test/scala/VerifyComputeModelStatistics.scala
```diff
@@ -47,17 +63,15 @@ class VerifyComputeModelStatistics extends TestBase {
...
     val evaluatedData = new ComputeModelStatistics().transform(datasetWithScores)
     val firstRow = evaluatedData.first()
-    assert(firstRow.get(0) == 0.0)
-    assert(firstRow.get(1) == 0.0)
-    assert(firstRow.get(2) == 1.0)
-    assert(firstRow.get(3) == 0.0)
+    assert(tolEq.areEquivalent(firstRow.get(0).asInstanceOf[Double], 0.0))
+    assert(tolEq.areEquivalent(firstRow.get(1).asInstanceOf[Double], 0.0))
+    assert(tolEq.areEquivalent(firstRow.get(2).asInstanceOf[Double], 1.0))
+    assert(tolEq.areEquivalent(firstRow.get(3).asInstanceOf[Double], 0.0))
```
mhamilton723: Is there a test for using CMS without the Schema manipulation API?

src/compute-model-statistics/src/test/scala/VerifyComputeModelStatistics.scala
```diff
@@ -81,10 +95,10 @@ class VerifyComputeModelStatistics extends TestBase {
 
     val evaluatedData = new ComputeModelStatistics().transform(datasetWithScores)
     val firstRow = evaluatedData.first()
-    assert(firstRow.get(0) == 0.0)
-    assert(firstRow.get(1) == 0.0)
-    assert(firstRow.get(2) == 1.0)
-    assert(firstRow.get(3) == 0.0)
+    assert(tolEq.areEquivalent(firstRow.get(0).asInstanceOf[Double], 0.0))
```
mhamilton723: can replace with "===" if you make tolEq implicit

src/compute-model-statistics/src/test/scala/VerifyComputeModelStatistics.scala
```diff
@@ -101,41 +115,25 @@ class VerifyComputeModelStatistics extends TestBase {
...
+    val linearRegressor = createLinearRegressor(otherLabelColumn)
     val scoredDataset =
-      TrainRegressorTestUtilities.trainScoreDataset(labelColumn, datasetWithAddedColumn, linearRegressor)
+      TrainRegressorTestUtilities.trainScoreDataset(otherLabelColumn, datasetWithAddedColumn, linearRegressor)
 
     val evaluatedData = new ComputeModelStatistics().transform(scoredDataset)
     val firstRow = evaluatedData.first()
-    assert(firstRow.get(0) == 0.0)
-    assert(firstRow.get(1) == 0.0)
+    assert(tolEq.areEquivalent(firstRow.get(0).asInstanceOf[Double], 0.0))
```
mhamilton723: replace with ===

src/core/serialize/src/main/scala/params/EstimatorArrayParam.scala
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package org.apache.spark.ml.param
+
+import org.apache.spark.ml.Estimator
+import scala.collection.JavaConverters._
+
+/** Param for Array of Models. */
+class EstimatorArrayParam(parent: Params, name: String, doc: String, isValid: Array[Estimator[_]] => Boolean)
```
mhamilton723: Please make sure to add this to codegen so it can be correctly picked up

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
mhamilton723: Still here lol

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
mhamilton723: Still here

src/compute-model-statistics/src/test/scala/VerifyComputeModelStatistics.scala
imatiach-msft: yep, test("Verify computing statistics on generic spark ML estimators is supported") {

src/compute-model-statistics/src/test/scala/VerifyComputeModelStatistics.scala
imatiach-msft: done

src/compute-model-statistics/src/test/scala/VerifyComputeModelStatistics.scala
imatiach-msft: done

src/core/serialize/src/main/scala/params/EstimatorArrayParam.scala
imatiach-msft: done

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
imatiach-msft: (as discussed I re-added this back for now due to a different issue in fuzzing)

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
imatiach-msft: (as discussed I re-added this back for now due to a different issue in fuzzing)

mhamilton723:
If you check this in, can you open an issue around adding Python Wrappers for this to keep this on the agenda?

PR #171, opened 2017-10-27
Add TextPreprocessor Transformer

eedeleon:
Adding a Transformer that performs a replace all from left to right, preferring longer matches over all key value pairs in a dictionary. It does not replace replacement texts

### DIFF
--- a/src/pipeline-stages/src/main/scala/TextPreprocessor.scala
+++ b/src/pipeline-stages/src/main/scala/TextPreprocessor.scala
@@ -1,0 +1,148 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.Transformer
+import org.apache.spark.ml.param.{MapParam, Param, ParamMap}
+import org.apache.spark.ml.util.{ComplexParamsReadable, ComplexParamsWritable, Identifiable}
+import org.apache.spark.sql.types.{StringType, StructField, StructType}
+import org.apache.spark.sql.{DataFrame, Dataset}
+import org.apache.spark.sql.functions.udf
+import spray.json.DefaultJsonProtocol._
+
+class Trie(map: Map[Char, Trie] = Map.empty,
+           textValue: Seq[Char] = Seq.empty,
+           normFunction: Char => Char = identity) extends Serializable {
+  val trieMap: Map[Char, Trie] = map
+  val value: Seq[Char] = textValue
+  val normFunc: Char => Char = normFunction
+  def getValueString: String = this.value.toString
+
+  def getNodeValue(key: Seq[Char], value: Seq[Char]): Seq[Char] = (key, value) match {
+    case (k, v) if k.tail.isEmpty => v
+    case (_, _) => Seq.empty[Char]
+  }
+  def put(key: String, value: String): Trie = this.put(key.toCharArray, value)
+
+  def put(key: Seq[Char], value: String, isNormalized: Boolean = false): Trie = (key, value) match {
+    case (_, _) if key.isEmpty => this.copy()
+    case (k, v) =>
+      val normalizedKey = if (isNormalized) k else k.map(this.normFunc)
+      val newLeaf = new Trie(textValue = getNodeValue(normalizedKey, v.toCharArray), normFunction = this.normFunc)
+      val next = this.trieMap.getOrElse(normalizedKey.head, newLeaf).put(normalizedKey.tail, v, true)
+      val newMap = this.trieMap + (normalizedKey(0) -> next)
+      new Trie(newMap, this.value, this.normFunc)
+  }
+
+  def get(letter: Char): Option[Trie] = this.trieMap.get(this.normFunc(letter))
+
+  def putAll(stringMap: Map[String, String]): Trie = {
+    stringMap.foldLeft(this){
+      case (out, (k, v)) => out.put(k, v)
+    }
+  }
+
+  def copy(): Trie = new Trie(this.trieMap, this.value, this.normFunc)
+
+  def mapText(inputText: String): String = {
+    val chars = inputText.toCharArray
+    val outputText: StringBuilder = new StringBuilder("")
+    def isAlpha(char: Char): Boolean = {
+      char.isLetterOrDigit || char.equals("_".charAt(0))
+    }
+    def skipAlphas(rest: Seq[Char]): Unit = (rest) match {
+      case _ if rest.isEmpty || !isAlpha(rest.head)=> scan(rest)
+      case _  => skipAlphas(rest.tail)
+    }
+
+    def findMatch(rest: Seq[Char],
+                  matched: Seq[Char],
+                  hasMatch: Boolean,
+                  chars: Seq[Char],
+                  trie: Option[Trie]): Unit = (rest, matched, hasMatch, chars, trie) match {
+      case (_) if (trie == None || chars.isEmpty) => {
+        outputText ++= matched
+        if (hasMatch) skipAlphas(rest) else scan(rest)
+      }
+      case (_)  if (trie.get.value.isEmpty) => findMatch(rest, matched, false, chars.tail, trie.get.get(chars.head))
+
+      case (_) => findMatch(chars, trie.get.value, true, chars.tail, trie.get.get(chars.head))
+    }
+
+    def scan(chars: Seq[Char]): Unit = {
+      if (!chars.isEmpty) findMatch(chars.tail, Array(chars.head), false, chars.tail, this.get(chars.head))
+    }
+
+    scan(chars)
+    outputText.mkString
+  }
+
+  override def toString: String = s"Val: $getValueString Rest:\n $trieMap"
+}
+
+object Trie {
+  def apply(map: Map[String, String]): Trie = {
+    new Trie().putAll(map)
+  }
+}
+
+object TextPreprocessor extends ComplexParamsReadable[TextPreprocessor]
+
+/** <code>TextPreprocessor</code> takes a dataframe and a dictionary
+  * that maps (text -> replacement text), scans each cell in the input col
+  * and replaces all substring matches with the corresponding value.
+  * Priority is given to longer keys and from left to right.
+  */
+class TextPreprocessor(val uid: String) extends Transformer
+  with HasInputCol with HasOutputCol with Wrappable with ComplexParamsWritable {
+  def this() = this(Identifiable.randomUID("TextPreprocessor"))
+  val IDENTITY = """identity"""
+
+  val normFuncs: Map[String, Char => Char] = Map[String, Char => Char] (
+    IDENTITY         -> identity,
+    """lowerCase"""  -> Character.toLowerCase,
+    """upperCase"""  -> Character.toUpperCase
+  )
+
+  val map = new MapParam[String, String](this, "map", "Map of substring match to replacement")
+
+  /** @group getParam */
+  def getMap: Map[String, String] = get(map).getOrElse(Map())
+
+  /** @group setParam */
+  def setMap(value: Map[String, String]): this.type = set("map", value)
+
+  def isValidNormFunc(normFuncName: String): Boolean = normFuncs.contains(normFuncName)
+
+  val normFunc = new Param[String](this, "normFunc", "Name of normalization function to apply", isValidNormFunc(_))
+
+  /** @group getParam */
+  def getNormFunc: String = get(normFunc).getOrElse("identity")
+
+  /** @group setParam */
+  def setNormFunc(value: String): this.type = set("normFunc", value)
+
+  /** @param dataset - The input dataset, to be transformed
+    * @return The DataFrame that results from column selection
+    */
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val spark = dataset.sparkSession
+    val inputIndex = dataset.columns.indexOf(getInputCol)
+    val trie = new Trie(normFunction = normFuncs(getNormFunc)).putAll(getMap)
+    val broadcastedTrie = spark.sparkContext.broadcast(trie)
+
+    require(inputIndex != -1, s"Input column $getInputCol does not exist")
+
+    val mapText: String => String = broadcastedTrie.value.mapText
+    val textMapper = udf(mapText)
+    dataset.withColumn(getOutputCol, textMapper(dataset(getInputCol)).as(getOutputCol))
+  }
+
+  def transformSchema(schema: StructType): StructType = {
+    schema.add(StructField(getOutputCol, StringType))
+  }
+
+  def copy(extra: ParamMap): TextPreprocessor = defaultCopy(extra)
+
+}
--- a/src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
+++ b/src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
@@ -1,0 +1,120 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.util.MLReadable
+
+class TextPreprocessorSuite extends TestBase with TransformerFuzzing[TextPreprocessor] {
+  val toMap1 = "The happy sad boy drank sap"
+  val toMap2 = "The hater sad doy drank sap"
+  val toMap3 = "The hater sad doy"
+  val INPUT_COL = "words1"
+  val OUTPUT_COL = "out"
+
+  val expectedResult = session.createDataFrame(Seq(
+    (toMap1, "The sad sap boy drank sap"),
+    (toMap2, "The sap sap drank sap"),
+    ("foo", "foo"),
+    (s"$toMap3 aABc0123456789Zz_", "The sap sap")))
+    .toDF(INPUT_COL, OUTPUT_COL)
+
+  val wordDF = expectedResult.drop(OUTPUT_COL)
+
+  val testMap = Map[String, String] (
+    "happy"   -> "sad",
+    "hater"   -> "sap",
+    "sad"     -> "sap",
+    "sad doy" -> "sap"
+  )
+
+  val testTrie1 = Trie(Map[String, String]("happy" -> "sad", "hater" -> "sap"))
+  var testTrie1Pivot: Trie = testTrie1
+  for (letter <- "ha") testTrie1Pivot = testTrie1Pivot.get(letter).get
+
+  test("Check for value of words with intersection in trie") {
+      var copyHappy: Trie = testTrie1
+      for (letter <- "happy") copyHappy = copyHappy.get(letter).get
+      assert(copyHappy.value.mkString("") == "sad")
+
+      var copyHater: Trie = testTrie1
+      for (letter <- "hater") copyHater = copyHater.get(letter).get
+      assert(copyHater.value.mkString("") == "sap")
+    }
+
+    test("Check continuation from child node in trie") {
+      var copyHaHater: Trie = testTrie1Pivot
+      for (letter <- "ppy") copyHaHater = copyHaHater.get(letter).get
+      assert(copyHaHater.value.mkString("") == "sad")
+
+      var copyHaHappy: Trie = testTrie1Pivot
+      for (letter <- "ter") copyHaHappy = copyHaHappy.get(letter).get
+      assert(copyHaHappy.value.mkString("") == "sap")
+  }
+
+  test("Check trie put vs putAll equality") {
+    val test2 = Trie(Map[String,String]("happy" -> "sad", "hater" -> "sap"))
+    val testString = "happy hater"
+    assert(testTrie1.mapText(testString).equals(test2.mapText(testString)))
+  }
+
+  test("Check trie text mapper") {
+    val mappings = Map[String, String]("happy" -> "sad", "hater" -> "sap", "sad" -> "sap", "sad doy" -> "sap")
+    val test = Trie(mappings)
+    print(test.mapText(toMap1))
+    assert(test.mapText(toMap1).equals("The sad sap boy drank sap"))
+    assert(test.mapText(toMap2).equals("The sap sap drank sap"))
+  }
+
+  test("Check trie multiple puts") {
+    var test = new Trie(normFunction = Character.toUpperCase)
+    test = test.put("happy", "sad").put("hater", "sap")
+    test = test.put("sad", "sap").put("sad doy", "sap")
+    test = test.put("the", "sat")
+    val item = test.mapText(toMap1)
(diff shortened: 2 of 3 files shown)


src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.spark.ml.param.{MapParam, ParamMap}
+import org.apache.spark.ml.util.{ComplexParamsReadable, ComplexParamsWritable, Identifiable}
+import org.apache.spark.sql.types.{StringType, StructField, StructType}
+import org.apache.spark.sql.{DataFrame, Dataset}
+import org.apache.spark.sql.functions.udf
+import spray.json.DefaultJsonProtocol._
+
+import scala.collection.mutable
+
+class Trie(normFunc: String => String = x => x) extends Serializable {
```
RatanRSur: `identity instead of x => x`

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,134 @@
...
+      val normK: String = normFunction(k)
+      val first = normK.substring(0, 1)
+      val next: Trie = trieMap.getOrElse(first, new Trie()).put(normK.substring(1), v)
+      if (normK.length() == 1) next.value = v
+      this.trieMap(normK.substring(0, 1)) = next
+      this
+    }
+  }
+
+  def get(letter:String): Trie = trieMap.getOrElse(letter, null)
```
RatanRSur: don't use `null` use `Option[String]` intstead

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
RatanRSur: I only just skimmed this but I don't see a reason this can't be immutable. `put` etc. can just return new instances

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,134 @@
...
+  /** @param dataset - The input dataset, to be transformed
+    * @return The DataFrame that results from column selection
+    */
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val spark = dataset.sparkSession
+    val inputIndex = dataset.columns.indexOf(getInputCol)
+    val trie = new Trie().putAll(getMap)
+    val broadcastedTrie = spark.sparkContext.broadcast(trie)
+
+    if (inputIndex == -1)
```
RatanRSur: more idiomatic to use `require` when throwing `IllegalArgumentExceptions`

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,134 @@
...
+      throw new IllegalArgumentException(s"Input column $getInputCol does not exist")
+
+    val df = dataset.toDF()
+    val mapText: String => String = broadcastedTrie.value.mapText
+    val textMapper = udf(mapText, StringType)
+    dataset.withColumn(getOutputCol, textMapper(dataset(getInputCol)).as(getOutputCol))
+  }
+
+  def transformSchema(schema: StructType): StructType = {
+    val newField = StructField(getOutputCol, StringType)
```
RatanRSur: inline newField

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,134 @@
...
+    schema.add(newField)
+  }
+
+  def copy(extra: ParamMap): TextPreprocessor = defaultCopy(extra)
+
+  private def verifySchema(schema: StructType): Unit = {
+    val outputCol = getOutputCol
+    assert(schema.fieldNames contains outputCol)
```
RatanRSur: inline outputCol

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: The map is mutable to speed up the process of creation. Importing mutable is for mutable.Map. Otherwise, each put call would make a new copy of the internal map, and then runtime is more like n^2*(word length) than n*(word length)

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: Require for what exactly? Can you link me to the require you are referring to?

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
RatanRSur: Performance would be a legit reason to make it mutable, however, did you benchmark it?
This http://docs.scala-lang.org/overviews/collections/performance-characteristics.html leads me to believe the O would be the same

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
RatanRSur: it's in the Predef, which is already imported in all scala scopes

https://github.com/scala/scala/blob/2.11.x/src/library/scala/Predef.scala#L222

eedeleon: Thanks! I added the change. Much cleaner now.

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: done

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: Hecho

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: done

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: Made the change. The get function for Options was a bit odd, but the solution is more robust now. Thanks

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: I made the change, scala seems to handle immutable trees really well.

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: done

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: done

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.spark.ml.param.{MapParam, Param, ParamMap}
+import org.apache.spark.ml.util.{ComplexParamsReadable, ComplexParamsWritable, Identifiable}
+import org.apache.spark.sql.types.{StringType, StructField, StructType}
+import org.apache.spark.sql.{DataFrame, Dataset}
+import org.apache.spark.sql.functions.udf
+import spray.json.DefaultJsonProtocol._
+
+class Trie(trieMap: Map[String, Trie] = Map[String, Trie](),
+           textValue: String = "",
+           normFunc: String => String = (text:String) => text.toLowerCase()) extends Serializable {
```
RatanRSur: nit: put spaces after all colons

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,157 @@
...
+import breeze.linalg.max
+import org.apache.spark.ml.Transformer
+import org.apache.spark.ml.param.{MapParam, Param, ParamMap}
+import org.apache.spark.ml.util.{ComplexParamsReadable, ComplexParamsWritable, Identifiable}
+import org.apache.spark.sql.types.{StringType, StructField, StructType}
+import org.apache.spark.sql.{DataFrame, Dataset}
+import org.apache.spark.sql.functions.udf
+import spray.json.DefaultJsonProtocol._
+
+class Trie(trieMap: Map[String, Trie] = Map[String, Trie](),
```
RatanRSur: `Map.empty` instead of `Map[String, Trie]()`

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,157 @@
...
+  val value: String = textValue
+
+  def getNodeValue(key: String, value:String): String = (key, value) match {
+    case (k, v) if k.length() == 1 => v
+    case (_, _) => ""
+  }
+
+  def getNode(key: String, value:String): Trie = {
+    val newLeaf = new Trie(textValue = getNodeValue(key, value), normFunc = this.normFunc)
+    this.trieMap.getOrElse(key.substring(0, 1), newLeaf)
```
RatanRSur: you should drop `this` unless it's unclear what something is referring to

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  def put(key:String, value:String, isNormalized:Boolean = false): Trie = (key, value) match {
+    case (_, _) if key.length() == 0 => this.copy()
+    case (k, v) =>
+      var normalizedKey:String = k
+      if (!isNormalized) normalizedKey = this.normFunc(k)
```
RatanRSur: really try to avoid using `var` in general, in this case, you would make this

`val normalizedKey = if (isNormalized) k else normFunk(k)`

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,157 @@
...
+      val next = this.getNode(normalizedKey, v).put(normalizedKey.substring(1), v, isNormalized = true)
+      val newMap = this.trieMap + (normalizedKey.substring(0, 1) -> next)
+      new Trie(newMap, this.value, this.normFunc)
+  }
+
+  def get(letter:String): Option[Trie] = this.trieMap.get(letter)
+
+  def putAll(stringMap: Map[String, String]): Trie = {
+    var out = this
+    stringMap.foreach {case (k, v) => out = out.put(k, v) }
```
RatanRSur: `foldLeft` is the immutable way to iterate through a collection applying a 2 arg function like `put`

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,157 @@
...
+    out
+  }
+
+  def copy(): Trie = new Trie(this.trieMap, this.value, this.normFunc)
+
+  def searchForWord(inputText:String, index: Int): (Int, String) = {
+    val normText = this.normFunc(inputText)
+    var current = this.get(normText(index).toString)
```
RatanRSur: challenge you to remove the vars here

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,157 @@
...
+        (normText(index + longestWordKeyLength).toString matches """[a-zA-Z0-9_]""")) {
+        longestWordKeyLength += 1
+      }
+    }
+    (max(index + 1, index + longestWordKeyLength), longestWord)
+  }
+
+  def mapText(inputText:String): String = {
+    var outputText: StringBuilder = new StringBuilder("")
+    var i = 0
```
RatanRSur: same

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,157 @@
...
+        outputText ++= inputText.substring(i, newIndex)
+      } else {
+        outputText ++= word
+      }
+      i = newIndex
+    } while (i != inputText.length())
+    outputText.toString()
+  }
+
+  override def toString: String = "Val:" + this.value + " Rest:\n" + this.trieMap.toString()
```
RatanRSur: Probably looks better as an interpolated string

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: Good call on one lining. However, I think "this." is useful since parameters could be similarly named. Is there any strong opinion on this from the team?

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: It may be possible, but I am not clear on what the net benefit would be. It would require rewriting in a style that falls less in line with the problem to be solved. Is this a team preference?

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: same as before.

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
elibarzilay: Should make it a trie of chars instead of one-length strings

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,141 @@
...
+
+import org.apache.spark.ml.Transformer
+import org.apache.spark.ml.param.{MapParam, Param, ParamMap}
+import org.apache.spark.ml.util.{ComplexParamsReadable, ComplexParamsWritable, Identifiable}
+import org.apache.spark.sql.types.{StringType, StructField, StructType}
+import org.apache.spark.sql.{DataFrame, Dataset}
+import org.apache.spark.sql.functions.udf
+import spray.json.DefaultJsonProtocol._
+
+class Trie(map: Map[Char, Trie] = Map.empty,
```
drdarshan: Please check what the accessibility is by default.. make ``private[namespace]`` if necessary.

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,141 @@
...
+           textValue: Array[Char] = Array.empty,
+           normFunction: Char => Char = identity) extends Serializable {
+  val trieMap: Map[Char, Trie] = map
+  val value: Array[Char] = textValue
+  val normFunc: Char => Char = normFunction
+  def getValueString: String = this.value.toString
+
+  def getNodeValue(key: Array[Char], value: Array[Char]): Array[Char] = (key, value) match {
```
drdarshan: Please consider using ``Seq`` instead of ``Array``

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,141 @@
...
+  def put(key: String, value: String): Trie = this.put(key.toCharArray, value)
+
+  def put(key: Array[Char], value: String, isNormalized: Boolean = false): Trie = (key, value) match {
+    case (_, _) if key.length() == 0 => this.copy()
+    case (k, v) =>
+      val normalizedKey = if (isNormalized) k else k.map(this.normFunc)
+      val newLeaf = new Trie(textValue = getNodeValue(normalizedKey, v.toCharArray), normFunction = this.normFunc)
+      val next = this.trieMap.getOrElse(normalizedKey.head, newLeaf).put(normalizedKey.tail, v, true)
+      val newMap = this.trieMap + (normalizedKey(0) -> next)
+      new Trie(newMap, this.value, this.normFunc)
```
drdarshan: Nice! Functional Trie!

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  /** @group setParam */
+  def setMap(value: Map[String, String]): this.type = set("map", value)
+
+  def isValidNormFunc(normFuncName: String): Boolean = normFuncs.contains(normFuncName)
+
+  val normFunc = new Param[String](this, "normFunc", "Name of normalization function to apply", isValidNormFunc(_))
+
+  /** @group getParam */
+  def getNormFunc: String = get(normFunc).getOrElse("identity")
```
drdarshan: Make a named constant for ``identity`` instead of a string?

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,141 @@
...
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val spark = dataset.sparkSession
+    val inputIndex = dataset.columns.indexOf(getInputCol)
+    val trie = new Trie(normFunction = normFuncs(getNormFunc)).putAll(getMap)
+    val broadcastedTrie = spark.sparkContext.broadcast(trie)
+
+    require(inputIndex != -1, s"Input column $getInputCol does not exist")
+
+    val mapText: String => String = broadcastedTrie.value.mapText
+    val textMapper = udf(mapText, StringType)
```
drdarshan: StringType is the default for udf?

src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  val testTrie1 = new Trie().put("happy", "sad").put("hater", "sap")
+  var testTrie1Pivot: Trie = testTrie1
+  for (letter <- "ha") testTrie1Pivot = testTrie1Pivot.get(letter).get
+
+  test("Check basic string mapper") {
+    {
+      var copyHappy: Trie = testTrie1
+      for (letter <- "happy") copyHappy = copyHappy.get(letter).get
+      assert(copyHappy != null)
```
drdarshan: Use ``None`` instead of ``null``.

src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+      assert(copyHappy.value.mkString("") == "sad")
+    }
+
+    {
```
drdarshan: Add a new test case for each instead of adding one big test. This will allow you to track failures in a more granular manner.

src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  test("Check trie put vs putAll equality") {
+    val test2 = new Trie().putAll(Map[String,String]("happy" -> "sad", "hater" -> "sap"))
+    val testString = "happy hater"
+    assert(testTrie1.mapText(testString).equals(test2.mapText(testString)))
+  }
+
+  test("Check trie text mapper") {
+    val mappings = Map[String, String]("happy" -> "sad", "hater" -> "sap", "sad" -> "sap", "sad doy" -> "sap")
+    val test = new Trie().putAll(mappings)
```
drdarshan: Please consider adding a ``object Trie`` version so you can write something like this: ``test = Trie(mappings)``

src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+      .setMap(testMap).setInputCol("words1")
+      .setOutputCol("out")
+      .setNormFunc("identity")
+    new TextPreprocessor()
+      .setMap(testMap)
+      .setInputCol("words1")
+      .setOutputCol("out")
+      .setNormFunc("lowerCase")
+    var errorThrown = false
+    try {
```
drdarshan: Please check - there is possibly an existing macro that asserts failure.

src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+    assert(errorThrown)
+  }
+
+  test("Check trie text df") {
+    val textPreprocessor = new TextPreprocessor()
+      .setNormFunc("lowerCase")
+      .setMap(testMap)
+      .setInputCol("words1")
+      .setOutputCol("out")
+    val processed = textPreprocessor.transform(wordDF)
```
drdarshan: Pretty sure there is a pre-existing function to compare two data frames for equality in our codebase (it's likely in one of the ``utils`` packages)

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
```diff
@@ -0,0 +1,141 @@
...
+      val next = this.trieMap.getOrElse(normalizedKey.head, newLeaf).put(normalizedKey.tail, v, true)
+      val newMap = this.trieMap + (normalizedKey(0) -> next)
+      new Trie(newMap, this.value, this.normFunc)
+  }
+
+  def get(letter: Char): Option[Trie] = this.trieMap.get(this.normFunc(letter))
+
+  def putAll(stringMap: Map[String, String]): Trie = {
+    stringMap.foldLeft(this){
+      case (out, (k, v)) => out.put(k.toCharArray, v)
```
RatanRSur: this can be `out.put(k, v)` right? Since you already have a `put` which takes `String`?

src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
RatanRSur: look into `assertThrows`

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: I made the fix. Only needed length like information for one location, i used empty tail as a idiomatic alternative to length == 1

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: Default is package level visibility. http://www.jesperdj.com/2016/01/08/scala-access-modifiers-and-qualifiers-in-detail/. That means I can leave as is correct?

src/pipeline-stages/src/test/scala/TextPreprocessorSuite.scala
eedeleon: Found two. Used verifyResults

eedeleon: the other was asserDataFrameEq

src/pipeline-stages/src/main/scala/TextPreprocessor.scala
eedeleon: I made the change, the initial idea was to have an external and internal contract. This was for convenience of the user of Trie.

PR #177, opened 2017-10-31
Add Work from Connect() Demo

mhamilton723:
- Add ability to write images back to storage
- Add ability to read images from a dataframe of paths


src/plot/src/main/python/plot.py
```diff
@@ -13,12 +13,12 @@
 from pyspark.sql import DataFrame
 from pyspark.sql.types import *
 
-import matplotlib.pyplot as plt
 import numpy as np
 from sklearn.metrics import confusion_matrix, roc_curve
 import itertools
 
 def confusionMatrix(df, y_col, y_hat_col, labels):
+    import matplotlib.pyplot as plt
```
mhamilton723: Catch inport error

src/plot/src/main/python/plot.py
```diff
@@ -13,12 +13,12 @@
 from pyspark.sql import DataFrame
 from pyspark.sql.types import *
```
mhamilton723: dont import star

src/plot/src/main/python/plot.py
```diff
@@ -0,0 +1,58 @@
...
+from pyspark.ml.param.shared import *
+from pyspark.sql import DataFrame
+from pyspark.sql.types import *
+
+import numpy as np
+from sklearn.metrics import confusion_matrix, roc_curve
+import itertools
+
+def confusionMatrix(df, y_col, y_hat_col, labels):
+    import matplotlib.pyplot as plt
```
drdarshan:
```python
try:
  import matplotlib.pyplot as plt
except ImportError:
  raise Error("Please install matplotlib in order to use plotting") ##.. or some such thing.
```

src/udf/src/main/scala/UtilityUDFs.scala
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
```
mhamilton723: put in different package

src/readers/src/main/scala/ImageReader.scala
```diff
@@ -97,4 +120,47 @@ object ImageReader {
...
+        val fs = path.getFileSystem(hconf.value)
+        val is = fs.open(path)
+        val imageRow = ImageReader.decode(path.toString,IOUtils.toByteArray(is)).getOrElse(Row(None))
+        val ret = Row.merge(Seq(row, Row(imageRow)):_*)
+        ret
+      }
+    }(encoder)
+  }
+
+  def write(df: DataFrame,
```
drdarshan: Split into a Writer class?

src/udf/src/main/scala/UtilityUDFs.scala
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License. See LICENSE in project root for information.
+
```
drdarshan: Move into a sub-package

src/plot/src/main/python/plot.py
mhamilton723: done

src/udf/src/main/scala/UtilityUDFs.scala
mhamilton723: made a udfs object. It is imported like a package.

src/plot/src/main/python/plot.py
mhamilton723: done

src/plot/src/main/python/plot.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
```
drdarshan: Plotting should go into its own sub-module.. please file an issue if you are not able to refactor for this check-in.

src/plot/src/main/python/plot.py
```diff
@@ -0,0 +1,61 @@
...
+        plt.text(j, i, cm[i, j],
+                 horizontalalignment="center",
+                 fontsize=18,
+                 color="white" if cmn[i, j] > thresh else "black")
+
+    plt.colorbar()
+    plt.xlabel("Predicted Label", fontsize=18)
+    plt.ylabel("True Label", fontsize=18)
+
+def roc(df, y_col, y_hat_col, thresh=.5):
```
drdarshan: Like `sklearn.metrics`, I would split the creation of the ROC curve from the plotting itself.

src/readers/src/main/python/PowerBIWriter.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
```
drdarshan: Put it into a submodule so it doesn't looks like:
``py
import mmlspark;
mmlspark.PowerBIWriter(...)
``

src/readers/src/main/scala/PowerBIWriter.scala
```diff
@@ -0,0 +1,101 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
```
drdarshan: Add a sub-package for visualization?

src/plot/src/main/python/plot.py
mhamilton723: This is in its own submodule. If a python class is not explicitly imported in the __init__.py its addressable by the filename aka mmlspark.plot

src/plot/src/main/python/plot.py
mhamilton723: I had wanted to provide a quick function that both computes and plots (though does not display allowing for composability and subplot embedding). If a user needed the extra control they could use the underlying sklearn.metrics function. Do you think that's reasonable?

src/readers/src/main/python/PowerBIWriter.py
mhamilton723: Again this should look like mmlspark.PowerBIWriter.writeToPowerBI(....)
or df.writeToPowerBI(....)

PR #181, opened 2017-11-08
Add RenameColumn Transformer

eedeleon:
Add a transformer for withColumnRenamed

### DIFF
--- a/src/pipeline-stages/src/main/scala/RenameColumn.scala
+++ b/src/pipeline-stages/src/main/scala/RenameColumn.scala
@@ -1,0 +1,39 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.Transformer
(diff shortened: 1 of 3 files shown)


src/pipeline-stages/src/main/scala/RenameColumn.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  private def verifyInputSchema(schema: StructType): Unit = {
+    val hasInputCol = schema.fields.map(_.name).toSet.contains(getInputCol)
+    if (!hasInputCol) {
+      throw new Exception (s"DataFrame does not contain specified input column: $getInputCol")
+    }
+  }
+  private def verifiedOutputSchema(schema: StructType): StructType = {
+    val hasOutputCol = schema.fields.map(_.name).toSet.contains(getOutputCol)
+    if (!hasOutputCol) {
```
drdarshan: Please use ``require`` instead of throwing ``Exception`` and consider using ``validateAndTransformSchema`` from Spark, (https://github.com/apache/spark/blob/v2.2.0/mllib/src/main/scala/org/apache/spark/ml/feature/StringIndexer.scala)

PR #189, opened 2017-11-21
Add UDFTransformer and fix minor PipelineStage bugs

eedeleon:


src/core/serialize/src/main/scala/params/UDFParam.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import org.apache.spark.sql.expressions.UserDefinedFunction
+
+/** Param for UserDefinedFunction.  Needed as spark has explicit params for many different
+  * types but not UserDefinedFunction.
+  */
+class UDFParam(parent: Params, name: String, doc: String, isValid: UserDefinedFunction => Boolean)
+  extends ComplexParam[UserDefinedFunction](parent, name, doc, isValid) {
+
+  def this(parent: Params, name: String, doc: String) =
```
RatanRSur: make this a default arg in the class constructor?

src/pipeline-stages/src/main/scala/UDFTransformer.scala
```diff
@@ -0,0 +1,49 @@
...
+  def getUDF: UserDefinedFunction = $(udf)
+
+  /** @group setParam */
+  def setUDF(value: UserDefinedFunction): this.type = set("udf", value)
+
+  /** @param dataset - The input dataset, to be transformed
+    * @return The DataFrame that results from applying the udf to the inputted dataset
+    */
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    transformSchema(dataset.schema, logging = true)
```
RatanRSur: unless there's a side effect I'm not noticing, this line doesn't do anything right?

eedeleon: HasInputCol? It sets up the inoutCol and outputCol params.

src/core/serialize/src/main/scala/params/UDFParam.scala
eedeleon: What would tat change look like? This is how PipelineStageParam works.

src/pipeline-stages/src/main/scala/UDFTransformer.scala
RatanRSur: are we looking at the same lines? I'm seeing

`transformSchema(dataset.schema, logging = true)`

eedeleon: This is done to check the schema early as a fail fast.

src/core/serialize/src/main/scala/params/UDPythonFParam.scala
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package org.apache.spark.ml.param
+
+import org.apache.spark.sql.execution.python.UserDefinedPythonFunction
+
+/** Param for UserDefinedFunction.  Needed as spark has explicit params for many different
+  * types but not UserDefinedFunction.
+  */
+class UDPythonFParam(parent: Params, name: String, doc: String, isValid: UserDefinedPythonFunction => Boolean)
```
drdarshan: Minor: `UDPyFParam`?

src/core/test/fuzzing/src/test/scala/FuzzingTest.scala
```diff
@@ -163,7 +163,8 @@ class FuzzingTest extends TestBase {
   }
 
   test("Verify all pipeline stage values match their param names") {
-    val exemptions: Set[String] = Set()
+    val exemptions: Set[String] = Set[String](
```
eedeleon: Get rid of exemptions

src/pipeline-stages/src/main/python/UDFTransformer.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+
+        Returns:
+
+            str: The name of the output column
+        """
+        return self.getOrDefault(self.outputCol)
+
+    def setUDF(self, udf):
+        func = None if not hasattr(udf, "func") else udf.func
```
drdarshan: replace with `func = getattr(udf, "func", None)`

src/pipeline-stages/src/main/python/UDFTransformer.py
```diff
@@ -0,0 +1,139 @@
...
+        returnType = None if not hasattr(udf, "returnType") else udf.returnType
+        userDefinedFunction = UserDefinedFunction(func, returnType = returnType, name = __name__)
+        self._java_obj = self._java_obj.setUDF(userDefinedFunction._judf)
```
drdarshan: Please just pass the UDF through to the `UserDefinedFunction` constructor. If it fails, that's consistent with Python error semantics..

src/pipeline-stages/src/main/python/UDFTransformer.py
```diff
@@ -0,0 +1,139 @@
...
+
+from pyspark.ml.common import inherit_doc
+from pyspark.sql.functions import UserDefinedFunction
+from pyspark.ml.common import inherit_doc
+from mmlspark.Utils import *
+
+@inherit_doc
+class UDFTransformer(ComplexParamsMixin, JavaMLReadable, JavaMLWritable, JavaTransformer):
+    """
+
```
elibarzilay: Why are there so many empty lines in docstrings in this file?

elibarzilay:
See my single comment about the zillion newlines, I'll merge it then.

src/pipeline-stages/src/main/python/UDFTransformer.py
eedeleon: I removed them but that is how the auto generated code looks.

PR #191, opened 2017-12-04
Structured Serving: REST endpoints with Structured Streaming

mhamilton723:


src/readers/src/test/scala/LoadBalancerSuite.scala
```diff
@@ -0,0 +1,36 @@
...
+import com.microsoft.ml.spark.FileUtilities.File
+import org.apache.spark.sql.functions.{length, col}
+class LoadBalancerSuite extends TestBase with FileReaderUtils {
+
+  test("Creation of a load balancer", TestBase.Extended) {
+    val auth = Azure.authenticate(AzureCliCredentials.create)
+
+    val setup = new AzureLoadBalancer(
+      auth,
+      "MAML Performance and Accuracy",
```
imatiach-msft: can the MAML perf & acc be removed here?  It's an internal only subscription.

PR #232, opened 2018-01-31
Adding python wrappers and notebook example for TuneHyperparameters

imatiach-msft:
Adding python wrappers and notebook example for TuneHyperparameters

### DIFF
--- a/src/tune-hyperparameters/src/main/python/HyperparamBuilder.py
+++ b/src/tune-hyperparameters/src/main/python/HyperparamBuilder.py
@@ -1,0 +1,88 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in project root for information.
+
+import sys
+
+if sys.version >= '3':
+    basestring = str
+
+import pyspark
+from pyspark import SparkContext
+from pyspark.ml.common import inherit_doc
+from pyspark.sql.types import *
+
+class HyperparamBuilder(object):
+    """
+    Specifies the search space for hyperparameters.
+    """
+    def __init__(self):
+        ctx = SparkContext.getOrCreate()
+        self.jvm = ctx.getOrCreate()._jvm
+        self.hyperparamBuilder = self.jvm.com.microsoft.ml.spark.HyperparamBuilder()
+
+    def addHyperparam(self, est, param, hyperParam):
+        """
+        Add a hyperparam to the builder
+
+        Args:
+            param (Param): The param to tune
+            dist (Dist): Distribution of values
+
+        """
+        javaParam = est._java_obj.getParam(param.name)
+        self.hyperparamBuilder.addHyperparam(javaParam, hyperParam.get())
+        return self
+
+    def build(self):
+        """
+        Builds the search space of hyperparameters, returns the map of hyperparameters to search through.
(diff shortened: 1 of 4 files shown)


src/tune-hyperparameters/src/main/python/HyperparamBuilder.py
```diff
@@ -0,0 +1,98 @@
...
+        return self.hyperparamBuilder.build()
+
+class DiscreteHyperParam(object):
+    """
+    Specifies a discrete list of values.
+    """
+    def __init__(self, values, seed=0):
+        ctx = SparkContext.getOrCreate()
+        self.jvm = ctx.getOrCreate()._jvm
+        valuesList = self.jvm.scala.collection.JavaConversions.asScalaBuffer(values).toList()
```
mhamilton723: Nit: add ArrayList constructor

src/tune-hyperparameters/src/main/python/HyperparamBuilder.py
```diff
@@ -0,0 +1,98 @@
...
+
+class RangeHyperParam(object):
+    """
+    Specifies a range of values.
+    """
+    def __init__(self, min, max, seed=0, isDouble=False):
+        ctx = SparkContext.getOrCreate()
+        self.jvm = ctx.getOrCreate()._jvm
+        if (isinstance(min, int)):
+            self.rangeParam = self.jvm.com.microsoft.ml.spark.IntRangeHyperParam(min, max, seed)
```
mhamilton723: is there any way to make sure this logic happens on the scala side?

src/tune-hyperparameters/src/main/python/HyperparamBuilder.py
imatiach-msft: done

src/tune-hyperparameters/src/main/python/HyperparamBuilder.py
imatiach-msft: done!

src/tune-hyperparameters/src/main/scala/HyperparamBuilder.scala
```diff
@@ -54,6 +54,36 @@ class DoubleRangeHyperParam(min: Double, max: Double, seed: Long = 0)
...
+object HyperParamUtils {
+  /**
+    * Returns a range hyper param by matching to the given input type.
+    * @param min The min value of the range
+    * @param max The max value of the range
+    * @param seed The random number seed.
+    * @return A RangeHyperParam matched to the given type for min and max values.
+    */
+  def getRangeHyperParam(min: Any, max: Any, seed: Long = 0): RangeHyperParam[_] = {
+    min match {
```
mhamilton723: You can use scala pattern matching to your advantage here: try something like

(min, max) match {
    case (a: Float, b: Float) => floatRange(a,b)
   ......

mhamilton723:
Very Close, just one Nit then feel free to merge

src/tune-hyperparameters/src/main/scala/HyperparamBuilder.scala
imatiach-msft: good idea, done!

PR #233, opened 2018-02-01
Updating per instance stats module to work for spark ML estimators in addition to TrainClassifier and TrainRegressor

imatiach-msft:
Updating per instance stats module to work for spark ML estimators in addition to TrainClassifier and TrainRegressor.
Also doing some minor refactoring and removing duplicate code.

### DIFF
--- a/src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
+++ b/src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
@@ -4,12 +4,13 @@
 package com.microsoft.ml.spark
 
 import com.microsoft.ml.spark.contracts.MetricData
+import com.microsoft.ml.spark.metrics.{MetricConstants, MetricUtils}
 import com.microsoft.ml.spark.schema.SchemaConstants._
 import com.microsoft.ml.spark.schema.{CategoricalUtilities, SchemaConstants, SparkSchema}
 import org.apache.spark.ml.Transformer
 import org.apache.spark.ml.linalg.Vector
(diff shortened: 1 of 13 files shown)


src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
```diff
@@ -18,83 +19,11 @@ import org.apache.spark.sql.functions._
...
-    else if (classificationMetrics.contains(metric)) true
-    else throw new Exception("Invalid metric specified")
-  }
-}
+object ComputeModelStatistics extends DefaultParamsReadable[ComputeModelStatistics]
 
-trait ComputeModelStatisticsParams extends MMLParams {
-  /** Metric to evaluate the models with. Default is "all"
+trait ComputeModelStatisticsParams extends MMLParams
+  with HasLabelCol with HasScoresCol with HasScoredLabelsCol with HasEvaluationMetric {
```
mhamilton723: Nice!

src/find-best-model/src/main/scala/FindBestModel.scala
```diff
@@ -52,6 +78,16 @@ class FindBestModel(override val uid: String) extends Estimator[BestModel]
     if (trainedModels.isEmpty) {
       throw new Exception("No trained models to evaluate.")
     }
+    if (!(Set(MetricConstants.MseSparkMetric,
```
mhamilton723: Nit: might want to add this set to MetricConstants so that users who add a metric can have it show up

imatiach-msft: done!

PR #235, opened 2018-02-06
New lightGBM learner

imatiach-msft:
New lightGBM learner

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMBooster.scala
+++ b/src/lightgbm/src/main/scala/LightGBMBooster.scala
@@ -1,0 +1,78 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import com.microsoft.ml.lightgbm.{SWIGTYPE_p_void, lightgbmlib, lightgbmlibConstants}
+import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector}
+
+/** Represents a LightGBM Booster learner
+  * @param model The string serialized representation of the learner
+  */
+class LightGBMBooster(val model: String) extends Serializable {
+  /**
+    * Transient variable containing local machine's pointer to native booster
+    */
+  @transient
+  var boosterPtr: SWIGTYPE_p_void = null
+
+  def numClasses(): Int = {
+    if (boosterPtr == null) {
+      LightGBMUtils.initializeNativeLibrary()
+      boosterPtr = getModel()
+    }
+    val numClasses = lightgbmlib.new_intp()
+    LightGBMUtils.validate(lightgbmlib.LGBM_BoosterGetNumClasses(boosterPtr, numClasses), "Booster GetNumClasses")
+    lightgbmlib.intp_value(numClasses)
+  }
+
+  def scoreRaw(features: Vector): Double = {
+    // Reload booster on each node
+    if (boosterPtr == null) {
+      LightGBMUtils.initializeNativeLibrary()
+      boosterPtr = getModel()
+    }
+    val numRows = 1
+    val rowsAsDoubleArray = features match {
+      case dense: DenseVector => dense.toArray
+      case sparse: SparseVector => sparse.toDense.toArray
+    }
+    val dataPtr = LightGBMUtils.generateData(numRows, Array(rowsAsDoubleArray))
+    val data64bitType = lightgbmlibConstants.C_API_DTYPE_FLOAT64
+
+    val numRowsIntPtr = lightgbmlib.new_intp()
+    lightgbmlib.intp_assign(numRowsIntPtr, numRows)
+    val numRows_int32_tPtr = lightgbmlib.int_to_int32_t_ptr(numRowsIntPtr)
+    val scoredDataLengthLongPtr = lightgbmlib.new_longp()
+    lightgbmlib.longp_assign(scoredDataLengthLongPtr, numRows)
+    val scoredDataLength_int64_tPtr = lightgbmlib.long_to_int64_t_ptr(scoredDataLengthLongPtr)
+
+    val numCols = rowsAsDoubleArray.length
+    val isRowMajor = 1
+    val numColsIntPtr = lightgbmlib.new_intp()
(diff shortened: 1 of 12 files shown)


src/lightgbm/src/main/scala/LightGBMClassifier.txt
```diff
@@ -0,0 +1,2 @@
+Trains a LightGBM Binary Classification model, a fast, distributed, high performance gradient boosting
```
drdarshan: Maybe add a link to the GitHub page for LightGBM?

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -0,0 +1,117 @@
...
+  def translate(labelColumn: String, featuresColumn: String, parallelism: String, log: Logger,
+                trainParams: TrainParams, inputRows: Iterator[Row]): Iterator[LightGBMBooster] = {
+    if (!inputRows.hasNext)
+      List[LightGBMBooster]().toIterator
+
+    val rows = inputRows.toArray
+    val numRows = rows.length
+    val rowsAsDoubleArray = rows.map(row => (row.get(row.fieldIndex(featuresColumn)) match {
+      case dense: DenseVector => dense.toArray
+      case sparse: SparseVector => sparse.toDense.toArray
```
drdarshan: It seems strange that you are converting sparse features to dense. LightGBM supports sparse features natively, see:  LGBM_DatasetCreateFromCSR and  LGBM_DatasetCreateFromCSC. Making a dense matrix from sparse rows will be very memory inefficient.

src/lightgbm/src/main/scala/LightGBMClassifier.txt
imatiach-msft: done

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: yes, I was planning to add sparse support later, but I can add it now.
In the case where there are mixed sparse and dense vectors I will just convert the rest to the format of the first row, as an intelligent computation may be expensive.

imatiach-msft:
@drdarshan I've updated the code to use CSR for rows with SparkML SparseVectors

src/lightgbm/src/main/scala/LightGBMBooster.scala
```diff
@@ -0,0 +1,121 @@
...
+      LightGBMUtils.initializeNativeLibrary()
+      boosterPtr = getModel()
+    }
+    val numClasses = lightgbmlib.new_intp()
+    LightGBMUtils.validate(lightgbmlib.LGBM_BoosterGetNumClasses(boosterPtr, numClasses), "Booster GetNumClasses")
+    lightgbmlib.intp_value(numClasses)
+  }
+
+  def scoreRaw(features: Vector): Double = {
+    // Reload booster on each node
```
imatiach-msft: I think we can eventually optimize scoring more by doing it in batch instead of on a row by row basis (in order to reduce the translation costs from Java to C++), although I'm honestly not sure how big the performance gains would be, since scoring seems to be quite fast already

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -0,0 +1,216 @@
...
+    val datasetParams = "max_bin=255 is_pre_partition=True"
+    val dataInt32bitType = lightgbmlibConstants.C_API_DTYPE_INT32
+    val data64bitType = lightgbmlibConstants.C_API_DTYPE_FLOAT64
+
+    // Generate the dataset for features
+    LightGBMUtils.validate(CSRUtils.LGBM_DatasetCreateFromCSR(indptrNative._1, dataInt32bitType,
+      indexes._1, values._1, data64bitType, intToPtr(indptr.length), intToPtr(valuesArray.length), intToPtr(numCols),
+      datasetParams, null, datasetOutPtr), "Dataset create")
+    // Delete the input rows
+    lightgbmlib.delete_doubleArray(values._2)
```
drdarshan: deletes must always be inside try..finally otherwise you will leak memory in the presence of exceptions. Could you please verify other places where you have delete_foo or delete_bar and ensure they are inside a try..finally guard?

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -0,0 +1,216 @@
...
+    if (nodes.isEmpty) {
+      // Running in local[*]
+      getNodesFromPartitions(data, defaultListenPort)
+    } else {
+      // Running on cluster, include all workers with driver excluded
+      (nodes.sorted.distinct.reduceLeft((val1, val2) => val1 + "," + val2), nodes.size)
+    }
+  }
+
+  def getNodesFromPartitions(processedData: DataFrame, defaultListenPort: Int): (String, Int) = {
```
drdarshan: I thought you did not need this anymore?

imatiach-msft: I only need it for the local[*] case, all of the executor id's are driver when using local[*] (which seems kind of like a bug)

drdarshan: For local[*], can't you just use localhost?

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -0,0 +1,233 @@
...
+      val data64bitType = lightgbmlibConstants.C_API_DTYPE_FLOAT64
+
+      // Generate the dataset for features
+      LightGBMUtils.validate(CSRUtils.LGBM_DatasetCreateFromCSR(indptrNative.get._1, dataInt32bitType,
+        indexes.get._1, values.get._1, data64bitType, intToPtr(indptr.length), intToPtr(valuesArray.length),
+        intToPtr(numCols), datasetParams, null, datasetOutPtr), "Dataset create")
+      lightgbmlib.voidpp_value(datasetOutPtr)
+    } finally {
+      // Delete the input rows
+      if (values.isDefined) {
```
drdarshan: Pretty sure delete_doubleArray would be a no-op if the pointer is null but I guess it doesn't hurt to check..

Issue #186, opened 2017-11-11
Find the problem with ComputeModelStatistics in NB 104


elibarzilay:
Find why the two commented cells with ComputeModelStatistics take so much time and fix it in some better way.

imatiach-msft:
closing as PR #237 has been merged

PR #237, opened 2018-02-08
Fixing random performance bug in CMS due to catalyst issue

imatiach-msft:
CMS sometimes gets randomly stuck in catalyst when dropping missing values if a specific sequence of estimators is run before it.  Caching the data resolves the issues with catalyst.  This looks like a bug in spark's catalyst which might have been already resolved in future versions of spark.

This resolves issue #186

### DIFF
--- a/src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
+++ b/src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
@@ -257,7 +257,10 @@ class ComputeModelStatistics(override val uid: String) extends Transformer with 
   private def selectAndCastToDF(dataset: Dataset[_],
                                 predictionColumnName: String,
                                 labelColumnName: String): DataFrame = {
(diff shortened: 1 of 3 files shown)


mabusch:
Thank you for finding a solution to the timing problem so that the more meaningful/interesting version of the notebook can be run.

src/compute-model-statistics/src/main/scala/ComputeModelStatistics.scala
```diff
@@ -381,7 +381,10 @@ class ComputeModelStatistics(override val uid: String) extends Transformer with
   private def selectAndCastToDF(dataset: Dataset[_],
                                 predictionColumnName: String,
                                 labelColumnName: String): DataFrame = {
+    // Note: We call cache in order to avoid a bug with catalyst where CMS seems to get stuck in a loop
+    // For future spark upgrade past 2.2.0, we should try to see if the cache() call can be removed
```
drdarshan: Please add TODO: tag so it's easier to grep down the line.

imatiach-msft:
done - updated based on comments, thanks!

PR #248, opened 2018-02-17
Refactor HTTP Client Library APIs

mhamilton723:


src/io/http/src/main/scala/HTTPTransformer.scala
```diff
@@ -3,149 +3,103 @@
...
-    val df = dataset.toDF()
-    val enc = RowEncoder(transformSchema(df.schema))
-    df.mapPartitions { it =>
-      clientHolder.get.send(it)
-    }(enc)
-  }
-
-  def copy(extra: ParamMap): HTTPTransformer[T] = defaultCopy(extra)
+  /** @group getParam */
+  def getAdvancedHandling: Boolean = $(advancedHandling)
```
mhamilton723: Maybe generalize pre-emptively

src/io/http/src/main/scala/SimpleHTTPTransformer.scala
```diff
@@ -0,0 +1,3 @@
+class SimpleHTTPTransformer {
```
mhamilton723: Whaaa

PR #249, opened 2018-02-21
Adding LightGBM regressor and refactoring common code

imatiach-msft:
Adding LightGBM regressor and refactoring common code


src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -71,9 +71,17 @@ object LightGBMUtils {
   def getNumExecutors(dataset: Dataset[_], defaultListenPort: Int): Int = {
     val executors = getExecutors(dataset, defaultListenPort)
     if (executors.isEmpty) {
-      // Case when run locally in local[*]
-      val blockManager = BlockManagerUtils.getBlockManager(dataset)
-      blockManager.master.getMemoryStatus.size
+      val master = dataset.sparkSession.sparkContext.master
```
drdarshan: Nice!

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -71,9 +71,17 @@ object LightGBMUtils {
...
+      // Case when run locally in local[_]
+      if (master.contains("local")) {
```
drdarshan: Minor nit, I would match a regex `local[number-or-star]`. If you master happens to have a name `yarnlocalized`  or something that just happens to match `local`, you shouldn't enter this branch.

src/lightgbm/src/main/scala/RegressorUtils.scala
```diff
@@ -0,0 +1,8 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
```
drdarshan: This is great but does it break backwards compatibility? Can I load models saved using older versions of MMLSpark?

src/lightgbm/src/main/scala/TrainParams.scala
```diff
@@ -0,0 +1,48 @@
...
+/** Defines the Booster parameters passed to the LightGBM classifier.
+  */
+class ClassifierTrainParams(parallelism: String, numIterations: Int, learningRate: Double, numLeaves: Int)
+  extends TrainParams(parallelism, numIterations, learningRate, numLeaves) {
+  override def toString(): String = {
+    s"objective=binary metric=binary_logloss,auc ${super.toString()}"
+  }
+}
+
+object ClassifierTrainParams {
```
drdarshan: Unless I'm mistaken, you could get the same effect by making `ClassifierTrainParams` into a case-class.

imatiach-msft: done

src/lightgbm/src/main/scala/RegressorUtils.scala
imatiach-msft: this only affects LightGBM so no backwards compat issues here

src/lightgbm/src/main/scala/LightGBMUtils.scala
imatiach-msft: good idea, done

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -71,9 +71,17 @@ object LightGBMUtils {
   def getNumExecutors(dataset: Dataset[_], defaultListenPort: Int): Int = {
     val executors = getExecutors(dataset, defaultListenPort)
     if (executors.isEmpty) {
+      val master = dataset.sparkSession.sparkContext.master
+      if (master == "local") 1
       // Case when run locally in local[*]
-      val blockManager = BlockManagerUtils.getBlockManager(dataset)
-      blockManager.master.getMemoryStatus.size
+      else if (master.matches("local\\[[*\\d+]\\]")) {
+        val numCores = master.replace("local[", "").replace("]", "").split(",").head
```
drdarshan: You can actually use regex match groups to just extract the element.

imatiach-msft: done!

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -71,9 +71,25 @@ object LightGBMUtils {
   def getNumExecutors(dataset: Dataset[_], defaultListenPort: Int): Int = {
     val executors = getExecutors(dataset, defaultListenPort)
     if (executors.isEmpty) {
-      // Case when run locally in local[*]
-      val blockManager = BlockManagerUtils.getBlockManager(dataset)
-      blockManager.master.getMemoryStatus.size
+      val master = dataset.sparkSession.sparkContext.master
+      if (master == "local") 1
+      else {
+        val capture = "local\\[([*\\d+])\\]".r
```
elibarzilay: 1. If `capture` here refers to the fact that the regexp is capturing,
   then it's a bad name, similar to using "variable" in a variable name.
   Same for `numCoresCapture` below.

2. The regexp seems broken: I think that what you want to use is

       "local\\[(\\*|\\d+)\\]".r

3. Also, there's some weird thing below about splitting
   `numCoresCapture` which seems wrong: I don't see any commas that
   could appear there.

imatiach-msft: Ah, good point, I think the regex can just be:
"local\\[(*)\\]".r
The '*' part can be any of:

local : Run Spark locally with one worker thread (i.e. no parallelism at all).

local[K] : Run Spark locally with K worker threads (ideally, set this to the number of cores on your machine).

local[K,F] : Run Spark locally with K worker threads and F maxFailures (see spark.task.maxFailures for an explanation of this variable)

local[*] : Run Spark locally with as many worker threads as logical cores on your machine.

local[*,F] : Run Spark locally with as many worker threads as logical cores on your machine and F maxFailures.

Reference:
https://jaceklaskowski.gitbooks.io/mastering-apache-spark/content/spark-local.html
https://stackoverflow.com/questions/32356143/what-does-local-mean-in-spark

imatiach-msft: Otherwise, it looks like this might work, I'll test it out (and then I won't need to do the split by comma):
"local\\[(\\*|\\d+)(:?,\\d+)?\\]".r

PR #253, opened 2018-03-01
Bump CNTK to 2.4

mhamilton723:


mhamilton723:
Changing env variable requires uploading all CNTK stuff. We should really decouple the eval and training libraries

elibarzilay:
Superseded by #255.

PR #259, opened 2018-03-08
Add MultiNGram Transformer

mhamilton723:


src/core/spark/src/main/scala/NamespaceInjections.scala
```diff
@@ -7,8 +7,8 @@ import org.apache.spark.ml.util.Identifiable
 
 object NamespaceInjections {
 
-  def pipelineModel(stages: Array[Transformer]): PipelineModel = {
-    new PipelineModel(Identifiable.randomUID("PipelineModel"), stages)
+  def pipelineModel[T <: Transformer](stages: Array[T]): PipelineModel = {
```
drdarshan: nice one :-)

src/text-featurizer/src/test/scala/MultiNGramSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+    .transform(dfRaw)
+
+  lazy val t = new MultiNGram().setLengths(Array(1, 3, 4)).setInputCol("tokens").setOutputCol("ngrams")
+  lazy val dfNgram = t.transform(dfTok)
+
+  test("operate on tokens ") {
+    val grams = dfNgram.collect().last.getAs[mutable.WrappedArray[String]]("ngrams").toSet
+    assert(grams("1 2 3 4"))
+    assert(grams("4"))
+    assert(grams("2 3 4"))
```
drdarshan: Is there a contains in Scala collections? so you can check for membership without checking for Nulls?

mhamilton723: most collection types have a contains method, but i thought it would be faster to convert to a set and use the apply syntax for inclusion checking

PR #264, opened 2018-03-15
Plotting Bugfix

mhamilton723:


src/plot/src/main/python/plot.py
```diff
@@ -18,7 +18,7 @@ def confusionMatrix(df, y_col, y_hat_col, labels):
     from sklearn.metrics import confusion_matrix
     import matplotlib.pyplot as plt
 
-    if isinstance(sdf, pyspark.sql.dataframe.DataFrame):
+    if isinstance(df, pyspark.sql.dataframe.DataFrame):
```
drdarshan: :-) We need more tests..

PR #266, opened 2018-03-19
Fix dealing with UTF-8 in the docker image

elibarzilay:


tools/docker/Dockerfile
```diff
@@ -13,10 +13,14 @@ RUN : '==== create a user (and also hide the random hostname in the prompt)' \
  && useradd -c "Microsoft ML for Apache Spark" -U -G root -d "$HOME" -m "$USER" \
  && : '==== install needed packages' \
  && apt-get update --fix-missing \
- && apt-get install -y curl unzip bzip2 libopenmpi1.10 libgomp1 libunwind8 \
```
mhamilton723: I dont think we need these "libopenmpi1.10 libgomp1 libunwind8  libtiff5 libpng12-0 libjasper1"

PR #267, opened 2018-03-20
Improve the HTTPTransformerSuite.scala tests

elibarzilay:
Make it use any free port instead of a fixed one.


mhamilton723:
Updated this PR and re-submitted

PR #274, opened 2018-03-29
Bugfixes

mhamilton723:


src/io/http/src/main/scala/HTTPClients.scala
```diff
@@ -31,7 +30,9 @@ private[ml] trait HTTPClient extends BaseClient[HTTPRequestData, HTTPResponseDat
   }
 
   private[ml] def fromResponse(response: Response): ContextOut  = {
-    ContextOut(new HTTPResponseData(response.response), response.context)
+    val out = ContextOut(new HTTPResponseData(response.response), response.context)
+    response.response.close()
```
drdarshan: This is an unusual convention so please check..
If a function is passed in (say) a stream, it is not usually responsible for closing it - it's typically the caller's responsibility. In fact, there are cases where closing an input stream inside a function might be the wrong thing to do (what if I want to do something with the stream later on)?

It might work in this particular case but this pattern is confusing and might be error-prone down the line.

src/io/http/src/main/scala/HTTPClients.scala
```diff
@@ -64,8 +65,13 @@ object AdvancedHTTPHandling {
     if (suceeded) {
       response
     } else {
-      Thread.sleep(retriesLeft.head)
-      sendWithRetries(client, request, retriesLeft.tail)
+      if (retriesLeft.isEmpty){
+        response
+      }else{
+        response.close()
+        Thread.sleep(retriesLeft.head)
```
drdarshan:
```
if (succeeded || retriesLeft.isEmpty) {
  response
} else {
}
```

src/io/http/src/main/scala/HTTPClients.scala
mhamilton723: Will give this some more thinking, thanks!

src/cntk-model/src/main/scala/CNTKModel.scala
```diff
@@ -93,24 +93,28 @@ private object CNTKModelUtils extends java.io.Serializable {
                  inputNode: Int,
                  outputNodeName: Option[String],
                  outputNodeIndex: Option[Int])(inputRows: Iterator[Row]): Iterator[Row] = {
-    val device = DeviceDescriptor.useDefaultDevice
-    val m = fromSerializable(broadcastedModel.value).clone(ParameterCloningMethod.Share)
-    val outputNode: Option[CNTKFunction] = (outputNodeName, outputNodeIndex) match {
-      case (Some(name), None) =>
+    if (inputRows.hasNext){
```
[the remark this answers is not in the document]
mhamilton723: Done

src/io/http/src/main/scala/HTTPClients.scala
mhamilton723: Refactored so this is better. This simplified alot so thanks for that!

PR #275, opened 2018-03-30
Updated 302 notebook

imatiach-msft:
Fixed 302 notebook for local docker case


mhamilton723:
Can you also include the fix on the medical entity nb? Thanks amigo!

PR #277, opened 2018-04-02
Standardize Batching across the library

mhamilton723:


src/cntk-model/src/main/scala/CNTKModel.scala
```diff
@@ -8,24 +8,25 @@ import com.microsoft.CNTK.{DataType => CNTKDataType, SerializableFunction => CNT
...
 import scala.collection.mutable
-import scala.collection.mutable.ListBuffer
 
 private object CNTKModelUtils extends java.io.Serializable {
 
   def applyCNTKFunction(model: CNTKFunction,
                         inputFVV: FloatVectorVector,
                         inputVar: Variable,
                         device: DeviceDescriptor): FloatVectorVector = {
+    println(s"evaluating ${System.nanoTime()/1000}")
```
mhamilton723: remove

PR #281, opened 2018-04-12
Added lightGBM docs

imatiach-msft:
Added lightGBM docs


mhamilton723:
Put these commits on my doc PR. Thanks for doing this so quickly!

PR #282, opened 2018-04-12
Adding parameters to lightGBM classifier and regressor.

katerinagl:
These parameters were added:
maxBin, baggingFraction, baggingFreq, baggingSeed, featureFraction, maxDepth, minSumHessianInLeaf


imatiach-msft:
It looks like there are just some minor style issues - the line length needs to be below 120 chars

imatiach-msft:
I've fixed some minor style issues, running a new build

imatiach-msft:
I've updated the default values for the params to be same as the defaults in LightGBM here:
https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.rst

katerinagl:
seems I don't have access to build results... so i can't fix it, sorry

katerinagl:
ok, thank you very much :)

PR #284, opened 2018-04-13
Fix Benchmarks Trait 

mhamilton723:


src/core/test/base/src/main/scala/TestBase.scala
```diff
@@ -204,3 +205,61 @@ abstract class TestBase extends FunSuite with BeforeAndAfterEachTestData with Be
...
+    }
+  }
+
+  val sortInDataframeEquality = false
+
+  val baseDfEq = new Equality[DataFrame]{
+    def areEqual(a: DataFrame, bAny: Any): Boolean = bAny match {
+      case ds:Dataset[_] =>
+        val b = ds.toDF()
+        if(a.columns !== b.columns){
```
imatiach-msft: the spacing here is a little weird, eg there should be a space between if and (, ) and {.
Also same issues below - perhaps style cop wasn't run yet

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
```diff
@@ -3,98 +3,125 @@
...
-           |    $hist
-           |  $benchmarkMetricsFile:$i:
-           |    $acc
-           |.""".stripMargin)
-    }
+  def compareBenchmark(bmNew: Benchmark, bmOld: Benchmark): Assertion = {
+    assert(bmNew.name == bmOld.name, "Benchmark names do not match")
+    assert(bmNew.higherIsBetter == bmOld.higherIsBetter, "higherIsBetter does not match")
+    assert(bmNew.precision === bmNew.precision, "precision does not match")
+    assert(bmNew.precision >=0, "precision needs to be positive")
```
imatiach-msft: I think precision could be negative if you are going to the 1's, 10's, hundreds etc place (by whole values)
In fact, I believe the regression benchmarks did this?
Also, the spacing is a little weird, as in the ">= 0" should be a space between op and value

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
```diff
@@ -3,98 +3,125 @@
...
+
+    val diff = bmNew.value - bmOld.value
+    assert(
+      if (bmNew.higherIsBetter){
+        diff + bmNew.precision > 0
+      }else{
```
imatiach-msft: spacing on the else needs to be fixed

src/lightgbm/src/test/resources/benchmarks_VerifyLightGBMRegressor.csv
```diff
@@ -0,0 +1,6 @@
+name,value,precision,higherIsBetter
+LightGBMRegressor_/home/marhamil/lib/datasets/Regression/Train/energyefficiency2012_data.train.csv,4.04001976483622,1.0,true
```
imatiach-msft: I see the full path (including marhamil username here), looks like this needs to be fixed

src/train-classifier/src/test/resources/benchmarks_VerifyTrainClassifier.csv
```diff
@@ -0,0 +1,113 @@
...
+MultilayerPerceptronClassifier_PimaIndian.csv_AUPR, 0.6845425867507886, 0.01, true
+RandomForestClassification_PimaIndian.csv_AUROC, 0.8288461538461543, 0.01, true
+RandomForestClassification_PimaIndian.csv_AUPR, 0.7225351272502225, 0.01, true
+DecisionTreeClassification_PimaIndian.csv_AUROC, 0.6195299145299146, 0.01, true
+DecisionTreeClassification_PimaIndian.csv_AUPR, 0.5574243067309479, 0.01, true
+LogisticRegression_PimaIndian.csv_AUROC, 0.5, 0.01, true
+LogisticRegression_PimaIndian.csv_AUPR, 0.6845425867507886, 0.01, true
+NaiveBayesClassifier_PimaIndian.csv_AUROC, 0.38431623931623937, 0.01, true
+NaiveBayesClassifier_PimaIndian.csv_AUPR, 0.30590923737343967, 0.01, true
+GradientBoostedTreesClassification_data_banknote_authentication.csv_AUROC, 0.9790730757711891, 0.01, true
```
imatiach-msft: I'd prefer to see learner, dataset and metric as separate strings instead of just one mega string

src/train-classifier/src/test/scala/VerifyTrainClassifier.scala
```diff
@@ -96,60 +94,16 @@ class VerifyTrainClassifier extends Benchmarks with EstimatorFuzzing[TrainClassi
...
 
-    val logisticRegressor = TrainClassifierTestUtilities.createLogisticRegressor(mockLabelColumn)
-    TrainClassifierTestUtilities.trainScoreDataset(mockLabelColumn, catDataset, logisticRegressor)
-
-    val randomForestClassifier = TrainClassifierTestUtilities.createRandomForestClassifier(mockLabelColumn)
-    TrainClassifierTestUtilities.trainScoreDataset(mockLabelColumn, catDataset, randomForestClassifier)
-  }
-
-  test("Verify saving and loading a trained classifier model") {
-    val dataset: DataFrame = createMockDataset
```
imatiach-msft: I would recommend keeping this, even though it is checked in the fuzzing, this also validates saving twice and some properties and it is good to have just in case fuzzing tests "fail to fail" since they are automated

imatiach-msft: (almost too automated)

src/tune-hyperparameters/src/test/scala/VerifyTuneHyperparameters.scala
```diff
@@ -143,8 +141,8 @@ class VerifyTuneHyperparameters extends Benchmarks {
     new TuneHyperparameters()
       .setModels(models.toArray)
       .setEvaluationMetric(MetricConstants.AccuracySparkMetric)
-      .setNumFolds(4)
-      .setNumRuns(models.length * 3)
+      .setNumFolds(2)
+      .setNumRuns(2)
```
imatiach-msft: 2 runs seems very low, I would keep it as models.length * 3

imatiach-msft: or maybe models.length * 2 at least

src/core/test/base/src/main/scala/TestBase.scala
mhamilton723: done

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
mhamilton723: done

src/core/test/benchmarks/src/main/scala/Benchmarks.scala
mhamilton723: Done

src/lightgbm/src/test/resources/benchmarks_VerifyLightGBMRegressor.csv
mhamilton723: thanks!

src/tune-hyperparameters/src/test/scala/VerifyTuneHyperparameters.scala
mhamilton723: this takes a super long time to run, we just need to test the basic functionality here

src/train-classifier/src/test/resources/benchmarks_VerifyTrainClassifier.csv
mhamilton723: unfortunately this really complicates things downstream, I thinks its better to have something simple and robust and leave more complex data comparisons as a good getting started project for a new member

src/train-classifier/src/test/scala/VerifyTrainClassifier.scala
mhamilton723: THis test does some very sketchy things with the working directory that makes it very brittle.  We should add these sorts of checks in the fuzzing infrastructure so everyone can benefit.

PR #288, opened 2018-04-18
Elaborate on databricks set up.

sharatsc:
Requires resolver options
https://mmlspark.azureedge.net/maven instead of the default maven central url


imatiach-msft:
@sharatsc I don't think this should actually be necessary, it just takes time for the packages to appear in maven central, eg see here:
https://mvnrepository.com/artifact/Azure/mmlspark
Right now I still only see up to 0.11 version, I think it just takes some time for them to appear there.

sharatsc:
Sounds good. I'll close this.

PR #289, opened 2018-04-18
Add missing import length in PySpark Serving Docco

chappers:
Adds the missing import which completes hello world example for pyspark serving


mhamilton723:
Thanks for finding this!

PR #309, opened 2018-06-01
Hostname to IP Address conversion

imatiach-msft:
Hostname to IP Address conversion

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMUtils.scala
+++ b/src/lightgbm/src/main/scala/LightGBMUtils.scala
@@ -4,10 +4,11 @@
 package com.microsoft.ml.spark
 
 import java.io._
-import java.net.{ServerSocket, Socket}
+import java.net.{InetAddress, ServerSocket, Socket}
 import java.util.concurrent.Executors
 
 import com.microsoft.ml.lightgbm._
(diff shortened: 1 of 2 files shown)


src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -109,6 +110,13 @@ object LightGBMUtils {
     idAsInt
   }
 
+  def getHostToIP(hostname: String): String = {
+    if (InetAddresses.isInetAddress(hostname))
+      hostname
+    else
+      InetAddress.getByName(hostname).getAddress.map(_ & 0xFF).mkString(".")
```
ywskycn: I think here you can directly use getHostAddress() function?

imatiach-msft: done, I was thinking of using that too originally but this seemed more reliable since it returns the ints directly, but since it worked for your cluster I will use it

mhamilton723:
@imatiach-msft This looks great, could you squash it and give it a good Commit message? Thank you!!

imatiach-msft:
@mhamilton723 I've updated the PR as suggested, thank you!

PR #323, opened 2018-06-19
Passing numMachines param to LightGBM

ywskycn:
Although we already pass list of nodes to LightGBM, it still needs the numMachines param.

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMRegressor.scala
+++ b/src/lightgbm/src/main/scala/LightGBMRegressor.scala
@@ -71,7 +71,7 @@ class LightGBMRegressor(override val uid: String)
     log.info(s"Nodes used for LightGBM: ${nodes.mkString(",")}")
     val trainParams = RegressorTrainParams(getParallelism, getNumIterations, getLearningRate, getNumLeaves,
       getApplication, getAlpha, getMaxBin, getBaggingFraction, getBaggingFreq, getBaggingSeed, getFeatureFraction,
-      getMaxDepth, getMinSumHessianInLeaf, nodes.length)
+      getMaxDepth, getMinSumHessianInLeaf, numExecutorCores)
     val networkParams = NetworkParams(nodes.toMap, getDefaultListenPort, inetAddress, port)
     val lightGBMBooster = df
       .mapPartitions(TrainUtils.trainLightGBM(networkParams, getLabelCol, getFeaturesCol,
--- a/src/lightgbm/src/main/scala/LightGBMRegressor.scala
+++ b/src/lightgbm/src/main/scala/LightGBMRegressor.scala
@@ -11,6 +11,7 @@ import org.apache.spark.sql._
 
 import scala.concurrent.Await
 import scala.concurrent.duration.{Duration, SECONDS}
+import scala.math.min
 import scala.reflect.runtime.universe.{TypeTag, typeTag}
 
 object LightGBMRegressor extends DefaultParamsReadable[LightGBMRegressor]
@@ -57,11 +58,11 @@ class LightGBMRegressor(override val uid: String)
     */
   override protected def train(dataset: Dataset[_]): LightGBMRegressionModel = {
     val numCoresPerExec = LightGBMUtils.getNumCoresPerExecutor(dataset)
-    val numExecutorCores = LightGBMUtils.getNumExecutorCores(dataset, numCoresPerExec)
-    // Reduce number of partitions to number of executor cores
-    val df = dataset.toDF().coalesce(numExecutorCores).cache()
+    val numWorkers = min(numCoresPerExec * numExec, dataset.rdd.getNumPartitions)
+    // Reduce number of partitions to number of executor cores if needed
+    val df = dataset.toDF().coalesce(numWorkers).cache()
     val (inetAddress, port, future) =
-      LightGBMUtils.createDriverNodesThread(numExecutorCores, df, log, getTimeout)
+      LightGBMUtils.createDriverNodesThread(numWorkers, df, log, getTimeout)
 
     val nodes = LightGBMUtils.getNodes(df, getDefaultListenPort, numCoresPerExec)
     /* Run a parallel job via map partitions to initialize the native library and network,
@@ -71,7 +72,7 @@ class LightGBMRegressor(override val uid: String)
     log.info(s"Nodes used for LightGBM: ${nodes.mkString(",")}")
     val trainParams = RegressorTrainParams(getParallelism, getNumIterations, getLearningRate, getNumLeaves,
       getApplication, getAlpha, getMaxBin, getBaggingFraction, getBaggingFreq, getBaggingSeed, getFeatureFraction,
-      getMaxDepth, getMinSumHessianInLeaf, numExecutorCores)
+      getMaxDepth, getMinSumHessianInLeaf, numWorkers)
     val networkParams = NetworkParams(nodes.toMap, getDefaultListenPort, inetAddress, port)
     val lightGBMBooster = df
       .mapPartitions(TrainUtils.trainLightGBM(networkParams, getLabelCol, getFeaturesCol,
--- a/src/lightgbm/src/main/scala/LightGBMRegressor.scala
+++ b/src/lightgbm/src/main/scala/LightGBMRegressor.scala
@@ -58,7 +58,8 @@ class LightGBMRegressor(override val uid: String)
     */
   override protected def train(dataset: Dataset[_]): LightGBMRegressionModel = {
     val numCoresPerExec = LightGBMUtils.getNumCoresPerExecutor(dataset)
-    val numWorkers = min(numCoresPerExec * numExec, dataset.rdd.getNumPartitions)
+    val numExecutorCores = LightGBMUtils.getNumExecutorCores(dataset, numCoresPerExec)
+    val numWorkers = min(numExecutorCores, dataset.rdd.getNumPartitions)
     // Reduce number of partitions to number of executor cores if needed
     val df = dataset.toDF().coalesce(numWorkers).cache()
     val (inetAddress, port, future) =
--- a/src/lightgbm/src/main/scala/LightGBMRegressor.scala
+++ b/src/lightgbm/src/main/scala/LightGBMRegressor.scala
@@ -71,7 +71,7 @@ class LightGBMRegressor(override val uid: String)
     log.info(s"Nodes used for LightGBM: ${nodes.mkString(",")}")
     val trainParams = RegressorTrainParams(getParallelism, getNumIterations, getLearningRate, getNumLeaves,
       getApplication, getAlpha, getMaxBin, getBaggingFraction, getBaggingFreq, getBaggingSeed, getFeatureFraction,
-      getMaxDepth, getMinSumHessianInLeaf)
+      getMaxDepth, getMinSumHessianInLeaf, nodes.length)
     val networkParams = NetworkParams(nodes.toMap, getDefaultListenPort, inetAddress, port)
     val lightGBMBooster = df
       .mapPartitions(TrainUtils.trainLightGBM(networkParams, getLabelCol, getFeaturesCol,
--- a/src/lightgbm/src/main/scala/TrainParams.scala
+++ b/src/lightgbm/src/main/scala/TrainParams.scala
@@ -17,12 +17,14 @@ abstract class TrainParams extends Serializable {
   def featureFraction: Double
   def maxDepth: Int
   def minSumHessianInLeaf: Double
+  def numMachines: Int
 
   override def toString(): String = {
     s"is_pre_partition=True boosting_type=gbdt tree_learner=$parallelism num_iterations=$numIterations " +
       s"learning_rate=$learningRate num_leaves=$numLeaves " +
       s"max_bin=$maxBin bagging_fraction=$baggingFraction bagging_freq=$baggingFreq bagging_seed=$baggingSeed " +
-      s"feature_fraction=$featureFraction max_depth=$maxDepth min_sum_hessian_in_leaf=$minSumHessianInLeaf"
+      s"feature_fraction=$featureFraction max_depth=$maxDepth min_sum_hessian_in_leaf=$minSumHessianInLeaf " +
+      s"numMachines=$numMachines"
   }
 }
 
@@ -31,7 +33,7 @@ abstract class TrainParams extends Serializable {
 case class ClassifierTrainParams(val parallelism: String, val numIterations: Int, val learningRate: Double,
                                  val numLeaves: Int, val maxBin: Int, val baggingFraction: Double, val baggingFreq: Int,
                                  val baggingSeed: Int, val featureFraction: Double, val maxDepth: Int,
-                                 val minSumHessianInLeaf: Double)
+                                 val minSumHessianInLeaf: Double, val numMachines: Int)
   extends TrainParams {
   override def toString(): String = {
     s"objective=binary metric=binary_logloss,auc ${super.toString()}"
@@ -44,7 +46,7 @@ case class RegressorTrainParams(val parallelism: String, val numIterations: Int,
                                 val numLeaves: Int, val application: String, val alpha: Double, val maxBin: Int,
                                 val baggingFraction: Double, val baggingFreq: Int,
                                 val baggingSeed: Int, val featureFraction: Double,
-                                val maxDepth: Int, val minSumHessianInLeaf: Double)
+                                val maxDepth: Int, val minSumHessianInLeaf: Double, val numMachines: Int)
   extends TrainParams {
   override def toString(): String = {
     s"objective=$application alpha=$alpha ${super.toString()}"
--- a/src/lightgbm/src/main/scala/TrainParams.scala
+++ b/src/lightgbm/src/main/scala/TrainParams.scala
@@ -24,7 +24,7 @@ abstract class TrainParams extends Serializable {
       s"learning_rate=$learningRate num_leaves=$numLeaves " +
       s"max_bin=$maxBin bagging_fraction=$baggingFraction bagging_freq=$baggingFreq bagging_seed=$baggingSeed " +
       s"feature_fraction=$featureFraction max_depth=$maxDepth min_sum_hessian_in_leaf=$minSumHessianInLeaf " +
-      s"numMachines=$numMachines"
+      s"num_machines=$numMachines"
   }
 }
 
--- a/src/lightgbm/src/main/scala/LightGBMClassifier.scala
+++ b/src/lightgbm/src/main/scala/LightGBMClassifier.scala
@@ -11,6 +11,7 @@ import org.apache.spark.sql._
 
 import scala.concurrent.Await
 import scala.concurrent.duration.{Duration, SECONDS}
+import scala.math.min
 import scala.reflect.runtime.universe.{TypeTag, typeTag}
 
 object LightGBMClassifier extends DefaultParamsReadable[LightGBMClassifier]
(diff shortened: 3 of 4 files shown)


mhamilton723:
This looks good but @imatiach-msft should really comment on this. Also, this leads me to believe that this should be cleaned up to make these huge parameter strings a bit more elegant

imatiach-msft:
This looks good to me, let me run a build and verify that this improves the accuracy for LightGBM on my Higgs notebook

imatiach-msft:
@ywskycn I ran this on the Higgs notebook with 2 iterations and the AUC was basically unchanged (slight decrease from 0.7415 to 0.7410) on an Azure Databricks cluster.  Do you have an example notebook where you see the performance improvement?

imatiach-msft:
@ywskycn I will try to run this on the notebook you sent me last week to see if the performance improves

ywskycn:
@imatiach-msft I tried the new patch in my local cluster with different size datasets, and the accuracy looks much better compared to before. And from the log, all executors are synced in each iteration, and resolves the issue mentioned in https://github.com/Azure/mmlspark/issues/318.

Basically, without this patch, each executor uses its own local data to train a local model, without global sync. Thus the result model would have a bad accuracy.

src/lightgbm/src/main/scala/LightGBMRegressor.scala
```diff
@@ -71,7 +71,7 @@ class LightGBMRegressor(override val uid: String)
     log.info(s"Nodes used for LightGBM: ${nodes.mkString(",")}")
     val trainParams = RegressorTrainParams(getParallelism, getNumIterations, getLearningRate, getNumLeaves,
       getApplication, getAlpha, getMaxBin, getBaggingFraction, getBaggingFreq, getBaggingSeed, getFeatureFraction,
-      getMaxDepth, getMinSumHessianInLeaf)
+      getMaxDepth, getMinSumHessianInLeaf, nodes.length)
```
troszok: I run this new version in our environment yesterday and I did see only slight improvement and I think we suffer from the same problem (the iterations are running out of sync), as our results in spark are not as good as the results we are getting on the same data in R.

@ywskycn @imatiach-msft i think here we should have number of partitions not the number of machines (in our case the number of machines is 3 and number of partitions 96) and i see in the logs:

```
18/06/19 20:15:37 INFO LightGBMRegressor: Nodes used for LightGBM: (1,172.33.8.230),(2,172.33.0.91),(0,172.33.5.247)
```

And I still the see the lines like these below that IMHO suggest that we are not syncing the model properly.

```

June 19th 2018, 22:22:11.745 | 18/06/19 20:22:10 INFO LightGBMRegressor: LightGBM running iteration: 5935 with result: 0 and is finished: 0
-- | --

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:10 INFO LightGBMRegressor: LightGBM running iteration: 5989 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:10 INFO LightGBMRegressor: LightGBM running iteration: 5889 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:10 INFO LightGBMRegressor: LightGBM running iteration: 5999 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:11 INFO LightGBMRegressor: LightGBM running iteration: 5978 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:10 INFO LightGBMRegressor: LightGBM running iteration: 5933 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:11 INFO LightGBMRegressor: LightGBM running iteration: 5934 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:11 INFO LightGBMRegressor: LightGBM running iteration: 5991 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:11 INFO LightGBMRegressor: LightGBM running iteration: 5939 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:10 INFO LightGBMRegressor: LightGBM running iteration: 5887 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:10 INFO LightGBMRegressor: LightGBM running iteration: 5975 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:11 INFO LightGBMRegressor: LightGBM running iteration: 5937 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:11 INFO LightGBMRegressor: LightGBM running iteration: 5976 with result: 0 and is finished: 0

  | June 19th 2018, 22:22:11.745 | 18/06/19 20:22:10 INFO LightGBMRegressor: LightGBM running iteration: 5999 with result: 0 and is finished: 0

```

ywskycn: I think here the nodes refers to the partitions, from https://github.com/Azure/mmlspark/blob/master/src/lightgbm/src/main/scala/LightGBMUtils.scala#L189.

If the input DataFrame's partitions is larger than #executors * #cores, it should be coalesced at the beginning. https://github.com/Azure/mmlspark/blob/master/src/lightgbm/src/main/scala/LightGBMRegressor.scala#L62.

Do I miss anything here?

src/lightgbm/src/main/scala/TrainParams.scala
```diff
@@ -17,12 +17,14 @@ abstract class TrainParams extends Serializable {
...
   def minSumHessianInLeaf: Double
+  def numMachines: Int
 
   override def toString(): String = {
     s"is_pre_partition=True boosting_type=gbdt tree_learner=$parallelism num_iterations=$numIterations " +
       s"learning_rate=$learningRate num_leaves=$numLeaves " +
       s"max_bin=$maxBin bagging_fraction=$baggingFraction bagging_freq=$baggingFreq bagging_seed=$baggingSeed " +
-      s"feature_fraction=$featureFraction max_depth=$maxDepth min_sum_hessian_in_leaf=$minSumHessianInLeaf"
+      s"feature_fraction=$featureFraction max_depth=$maxDepth min_sum_hessian_in_leaf=$minSumHessianInLeaf " +
+      s"numMachines=$numMachines"
```
troszok: @ywskycn i think the param should be `num_machines`. When using your version I was getting:

```
[LightGBM] [Warning] Unknown parameter: numMachines
```

after changing it to the `num_machines` the warning was gone and the training started to be synchronized across workers. 

The problem is that the training is very slow now it takes around 30 seconds per iteration. So I am now waiting for the results.

src/lightgbm/src/main/scala/LightGBMRegressor.scala
troszok: @ywskycn i think the nodes refers to the spark executors - as it gets it from this method:
https://github.com/Azure/mmlspark/blob/master/src/lightgbm/src/main/scala/LightGBMUtils.scala#L145

and in my case in line https://github.com/Azure/mmlspark/blob/master/src/lightgbm/src/main/scala/LightGBMRegressor.scala#L71
it prints:

```
18/06/19 20:15:37 INFO LightGBMRegressor: Nodes used for LightGBM: (1,172.33.8.230),(2,172.33.0.91),(0,172.33.5.247)
```

While later in the process I am having the following debug from LightGBM:

```
18/06/20 12:40:58 INFO LightGBMRegressor: driver writing back to all connections: 172.33.4.88:12410,172.33.4.88:12419,172.33.4.88:12411,172.33.4.88:12417,172.33.4.88:12412,172.33.4.88:12403,172.33.4.88:12409,172.33.4.88:12401,172.33.4.88:12415,172.33.4.88:12402,172.33.4.88:12407,172.33.4.88:12405,172.33.4.88:12418,172.33.4.88:12408,172.33.4.88:12416,172.33.4.88:12404,172.33.4.88:12406,172.33.4.88:12400,172.33.4.88:12413,172.33.4.88:12414
```

which is displaying all the LightGBM workers used in the training and it reflects the number of partitions. 

(Those two  lines are from different trainings so the IPs can differ).

src/lightgbm/src/main/scala/TrainParams.scala
imatiach-msft: yes, it can be num_machine or num_machines (see aliases), good catch:

https://github.com/Microsoft/LightGBM/blob/0584065c036414f6e4ae01e72314f2f9cbed7329/docs/Parameters.rst#network-parameters

num_machines, default = 1, type = int, aliases: num_machine, constraints: num_machines > 0

    the number of machines for parallel learning application
    this parameter is needed to be set in both socket and mpi versions

src/lightgbm/src/main/scala/LightGBMRegressor.scala
imatiach-msft: @troszok  @ywskycn  this is a very good catch, that's actually the number of executors, not nodes

imatiach-msft: @troszok @ywskycn I think you want to pass in numExecutorCores

imatiach-msft:
@ywskycn this looks promising, but it seems that we need to make a few minor changes, eg numMachines parameter is ignored currently and we are passing the number of executors instead of:
workers = number_of_executors * cores_per_executor

ywskycn:
Thanks @troszok @imatiach-msft  for the good catch. Updated the diff.

ywskycn:
Actually instead of numExecutorCores, here should be "dataset"'s partition number, right? Consider the dataset's #partitions may be smaller than numExecutorCores. This also affect `LightGBMUtils.createDriverNodesThread(numExecutorCores, df, log, getTimeout)`, where I think we also should use dataset's #partitions.

imatiach-msft:
@ywskycn maybe the variable name needs to be made clearer ---

*in general*:
numExecutorCores = num executors * num cores per executor

But you are right, in the special case where there are fewer partitions than num executors * num cores per executor, we use the number of partitions:
numExecutorCores = min(num executors * num cores per executor, numPartitions)

Perhaps there is a better variable name, but the logic should be correct (?)

imatiach-msft:
@ywskycn actually, looking at the code more, there does seem to be an issue, when we call getNodes:
val nodes = LightGBMUtils.getNodes(df, getDefaultListenPort, numCoresPerExec)
That is where we delve into the special case:
// Special case when num partitions < num executors
getNodesFromPartitions(data, defaultListenPort, nodes.toMap)
It looks like you are right that we are not doing this for getNumExecutorCores

ywskycn:
yes. I'm doing some tests locally to include "numExecutorCores = min(num executors * num cores per executor, numPartitions)". Will update the diff soon.

Another issue there is: `val getSubsetExecutors = data.rdd.getNumPartitions < nodes.length`. Here I think we should compare #partitions with "num executors * num cores", instead of just nodes, right?

imatiach-msft:
@ywskycn good catch, that is correct, we should be using num execs * num cores there - maybe the variable "nodes" is misleading now, after the recent code refactor it's not really "nodes" anymore but instead just executors.  We really need to have more test cases as it seems different scenarios/cases are getting broken frequently with new code changes...

imatiach-msft:
@ywskycn great, let me run another build, one without tests and another with tests run - and I will try it on my Higgs dataset example first (I also expect that in the second build there might be test failures as supposedly the metrics, eg AUC, should change)

imatiach-msft:
@ywskycn it looks like there is minor correction that needs to be made, there is no numExec variable:

2018-06-20T20:36:37.6104260Z ##[error]$HOME/agent/_work/4/s/src/lightgbm/src/main/scala/LightGBMClassifier.scala:38:44: not found: value numExec
2018-06-20T20:36:37.6125150Z ##[error]val numWorkers = min(numCoresPerExec * numExec, dataset.rdd.getNumPartitions)
2018-06-20T20:36:37.6140390Z ##[error]^
2018-06-20T20:36:37.7501020Z ##[error]$HOME/agent/_work/4/s/src/lightgbm/src/main/scala/LightGBMRegressor.scala:61:44: not found: value numExec
2018-06-20T20:36:37.7525210Z ##[error]val numWorkers = min(numCoresPerExec * numExec, dataset.rdd.getNumPartitions)
2018-06-20T20:36:37.7548120Z ##[error]^
2018-06-20T20:36:38.3245200Z ##[error]two errors found

imatiach-msft:
@ywskycn I think you still need to calculate:
val numExecutorCores = LightGBMUtils.getNumExecutorCores(dataset, numCoresPerExec) 
and maybe then use:
val numWorkers = min(numExecutorCores, dataset.rdd.getNumPartitions)
I think that might fix the build, let me know when you've updated the code

imatiach-msft:
@ywskycn I verified on Azure Databricks that the AUC improved slightly with the new changes on the Higgs example, it was 0.7448 (from 0.7415) for 2 iterations and 0.811 (from 0.809) for 100 iterations

imatiach-msft:
Also, I got your branch locally, ran the tests and verified that the results were consistently (slightly) better (except for one dataset) than before, so I think this is a good change, especially if you see that the workers are synchronized now

imatiach-msft:
@ywskycn  I think we can merge this in tomorrow - @troszok did you have a chance to try out the new changes?  Do you still see this issue:
"as our results in spark are not as good as the results we are getting on the same data in R"
If so there might be other issues that we may need to look into - might you have a simple notebook example with a public dataset that I could investigate?

imatiach-msft:
@mhamilton723 could we merge this PR?

mhamilton723:
Thanks for your hard work @ywskycn and @imatiach-msft! Do you think you could squash this into a single commit which is based on top of master called "Passing numMachines param to LightGBM"? then ill queue a final build and push it in!

imatiach-msft:
@mhamilton723 @ywskycn yes, please squash - this is a great change to have, thanks for fixing this @ywskycn.  As a follow up, I need to investigate how this parameter works in LightGBM package more and if there are any other parameters that should be set.

troszok:
@imatiach-msft I started the training but the training is really slow now (1 iteration takes between 15-30 seconds) so I am still waiting for it to complete. I will let you know if the accuracy is in par with the results we are getting in R but in general it is really strange that it is so slow. 

Did you compare the speed of the LightGBM on Spark with the LightGBM in R? 
I will try to run your notebook on our cluster and see how long it will take maybe it is something specific to our configuration.

imatiach-msft:
@troszok for me it took the same amount of time, are you sure you are using the right build, since there have been several iterations of the code with fixes:

--packages com.microsoft.ml.spark:mmlspark_2.11:0.12.dev21+4.g90b0ddc
--repositories https://mmlspark.azureedge.net/maven


I've also seen an issue where for the quantile regressor the learner gets stuck, which I still need to investigate more

imatiach-msft:
@troszok could you also send your log output -- if it is taking a while, almost always something went wrong; for the Higgs 4 GB data it took 80 seconds on 5 nodes

troszok:
@imatiach-msft thank you - I am using exactly 0.12.dev21+4 version of the mmlspark.

I see this in the logs (only the last several hundred lines): 
https://gist.github.com/troszok/971c4b24b1261eb5fe78c0dfcfc28583

imatiach-msft:
@troszok the first thing I noticed is you are setting numIterations=1000 which is quite large (default in LGBM is 100), was this intentional?

troszok:
@imatiach-msft yes - it is intentional  - when we train our models in R we are using 5000 iterations to get the production quality model - as you can see other params (eta, min_sum_hessian_in_leaf etc..) are set accordingly to prevent overfitting. 

Our training in R with 5000 iterations takes around 1 hour.

troszok:
@imatiach-msft so after some digging i found out that the option that causes the LightGBM to really slow down (from < 1 second per iteration) to > 15 seconds per iteration is: 

```
setMaxDepth(15).setNumLeaves(216)
```

Strange thing is that in R we do not have such a slowdown. Do you think there is anything specific in case of Spark? Networking overhead?

I also saw in the logs:
```
[LightGBM] [Warning] Set TCP_NODELAY failed.
```

Could that influence the speed of sync after each iteration?

imatiach-msft:
@troszok this is a very interesting find.  Would you be able to share a sample of your R code, specifically the parameters that you are initializing with?  If you have a sample dataset and R file that would be ideal.
  I'm wondering if either 1.) we are using different versions of lightgbm and there is some difference in the code or 2.) there might be some complex interaction happening with multiple parameters?

imatiach-msft:
@troszok another thing that I would really like to test is if running on a dataset with a single partition would lead to a speedup (since it would then be executed in a similar way to the R learner, which as I understand is single node based on your info), but this test could only be done if the dataset is small enough -- how large is the dataset?

imatiach-msft:
@troszok a max depth of 15 is quite large, I'm surprised the R code can run so quickly with such a large tree

imatiach-msft:
@troszok just guessing, but do you perchance have categorical features in your dataset?  Specifying which columns are categorical is something that I still need to figure out how to do in the spark case (which LGBM can natively handle).

troszok:
@imatiach-msft we do not have the categorical features (yet but we are willing to have them soon). 

The dataset I am currently using are ~10GB parquet files 7M rows with ~200 features.

Following your suggestion the test I did today was to coalesce my data to a single partition and the strange thing is that now I am using my CPU fully - regardless of the number of cores I assign to the spark job:

screen shot 2018-06-22 at 10 47 26

So it looks like when we let it LightGBM will by itself use all the resources. I checked the documentation to find this:

```
num_threads, default = 0, type = int, aliases: num_thread, nthread, nthreads number of threads for LightGBM
0 means default number of threads in OpenMP
```

I am not familiar with the details of OpenMP  but shouldn't we specify explicitly the num threads each executor should be using? Otherwise if we have multiple spark executors on single machine wouldn't they fight for the resources between each other?

The metrics I am getting when running training with partitioned data suggest that, as I see in the graph (but it can be also network overhead):

screen shot 2018-06-21 at 09 08 37

imatiach-msft:
@troszok this is a good point:
"I am not familiar with the details of OpenMP but shouldn't we specify explicitly the num threads each executor should be using? Otherwise if we have multiple spark executors on single machine wouldn't they fight for the resources between each other?"
I also thought about this - each executor has a separate JVM, and it seems that each JVM has some number of cores assigned to it (not the total for the machine).  However, lightGBM is executed in the native layer, so it might be that it is trying to use all cores which results in bad behavior.  I see that in the distributed example here:
https://github.com/Microsoft/LightGBM/blob/master/examples/parallel_learning/train.conf
They don't specify the num_threads (it is commented out).
I wonder if we do specify the num_threads (say set it to the number assigned per executor in the JVM), if LightGBM will allocate the cores correctly and we would get better performance, it's worth trying out.  It should be a straightforward code change, I'll do a test PR and create a build.

imatiach-msft:
@troszok does the new build here improve performance for you:
https://github.com/Azure/mmlspark/pull/326
--packages com.microsoft.ml.spark:mmlspark_2.11:0.12.dev24+1.gea49272
--repositories https://mmlspark.azureedge.net/maven

imatiach-msft:
I ran it on my cluster but the performance seems the same - probably it's because I am using 1 executor per machine in my Azure Databricks setup

imatiach-msft:
@troszok I'm not seeing this issue because my cluster setup has just 1 executor per node in azure databricks.  I'm no sure how to change the configuration, will have to look into it more.

ywskycn:
I also looked num_threads para before. I have a question here, consider a spark with _m_ executor and each executor has _n_ vcores. If we partition the original dataset into _m*n_ partitions, each partition will be only assigned one vcore, right? So in that case, num_thread can be just 1, as you may only still just have 1 core for 1 partition. (for a single server, #vcores should be configured larger than #physical_cores, I think).

imatiach-msft:
"If we partition the original dataset into m*n partitions, each partition will be only assigned one vcore, right?"
@ywskycn yes, that is definitely correct.  "So in that case, num_thread can be just 1", interesting, that does make more sense, let me create another PR and see if using 1 for num threads improves the performance, supposedly I should see a speedup compared to using the number of cores in the previous PR (although interestingly I didn't see any change compared to not specifying the parameter)

troszok:
@ywskycn @imatiach-msft i will be able to test it only tomorrow but thanks for the builds already! :)

imatiach-msft:
@troszok @ywskycn I tested out the new PR #327 but unfortunately I didn't notice any difference in performance on the Higgs dataset (for 2 iterations it was 78 secs before and 83.6 secs with the fix to use 1 thread, for 100 iterations it was 164 seconds before and 192 seconds after the change to use 1 thread).  Please let me know if you see any difference in performance with either of the two PRs #327 or #326  , I am currently not able to see any significant difference, but it could be related to my cluster configuration (Azure Databricks, 5 workers, Standard_DS4_v2 driver and workers which have 8 cores, default Azure Databricks settings).

PR #326, opened 2018-06-22
performance improvement: passing num_threads to lightgbm to be number of cores per executor

imatiach-msft:
performance improvement: passing num_threads to lightgbm to be number of cores per executor


imatiach-msft:
@troszok @mhamilton723 @ywskycn closing #326 as well

PR #327, opened 2018-06-23
setting num threads to 1 for lightGBM (to test if perf improves)

imatiach-msft:
setting num threads to 1 for lightGBM (to test if perf improves)

### DIFF
--- a/src/lightgbm/src/main/scala/TrainParams.scala
+++ b/src/lightgbm/src/main/scala/TrainParams.scala
@@ -24,7 +24,7 @@ abstract class TrainParams extends Serializable {
       s"learning_rate=$learningRate num_leaves=$numLeaves " +
       s"max_bin=$maxBin bagging_fraction=$baggingFraction bagging_freq=$baggingFreq bagging_seed=$baggingSeed " +
       s"feature_fraction=$featureFraction max_depth=$maxDepth min_sum_hessian_in_leaf=$minSumHessianInLeaf " +
-      s"num_machines=$numMachines"
+      s"num_machines=$numMachines num_threads=1"
   }
 }
 


troszok:
@imatiach-msft i just tested this version and:
- when I do coalesce(1) it works as expected (so it means really slow) - we use singe thread for training - the CPU usage on every machine is very low
- when I use the default number of partitions (which is the number of cores available in the cluster) I see again the problem with the out of sync iterations:

```
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3894 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3745 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3920 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3977 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3783 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3868 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3891 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3939 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3921 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3916 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3777 with result: 0 and is finished: 0
18/06/25 06:56:03 INFO LightGBMRegressor: LightGBM running iteration: 3866 with result: 0 and is finished: 0
```

Looks like we either can use num_machines (for parallel training with multiple executors, where lightgbm uses all the CPU resources) or num_threads (for single machine and than you can tune how many threads you want to use). 

Unless you have some other ideas - I think this is dead end I will move back to the previous num_machines only version and try to figure out the difference between the R and Spark.

imatiach-msft:
thanks for testing it out @troszok, I will close both PRs #327 and #326 now since you are seeing worse performance and I am not seeing any difference.
I did go through LightGBM source and I see it being read and used:
https://github.com/Microsoft/LightGBM/blob/dc6995742a5284a1e942978e2542fc49adda9ea1/src/c_api.cpp#L126
but it is used separately from the num_machines variable:
https://github.com/Microsoft/LightGBM/blob/dc6995742a5284a1e942978e2542fc49adda9ea1/src/network/network.cpp#L27
Which is passed in network init

FYI @ywskycn @mhamilton723

PR #336, opened 2018-07-03
Added params writable and readable to lightGBM models

terrytangyuan:
We had issues with `copy()` during our custom cross validation. https://github.com/Azure/mmlspark/commit/0deeca8304fae8734b5eb0a23ac68c3412f25446 didn't quite fix it for us. It seems like these additional changes are needed in order to fix it.

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMClassifier.scala
+++ b/src/lightgbm/src/main/scala/LightGBMClassifier.scala
@@ -25,7 +25,7 @@ object LightGBMClassifier extends DefaultParamsReadable[LightGBMClassifier]
 @InternalWrapper
 class LightGBMClassifier(override val uid: String)
   extends ProbabilisticClassifier[Vector, LightGBMClassifier, LightGBMClassificationModel]
-  with LightGBMParams {
+  with LightGBMParams with DefaultParamsWritable {
   def this() = this(Identifiable.randomUID("LightGBMClassifier"))
 
   /** Trains the LightGBM Classification model.
@@ -73,7 +73,7 @@ class LightGBMClassificationModel(
   featuresColName: String, predictionColName: String, probColName: String,
   rawPredictionColName: String, thresholdValues: Option[Array[Double]])
     extends ProbabilisticClassificationModel[Vector, LightGBMClassificationModel]
-    with ConstructorWritable[LightGBMClassificationModel] {
+    with ConstructorWritable[LightGBMClassificationModel] with DefaultParamsWritable {
 
   // Update the underlying Spark ML params
   // (for proper serialization to work we put them on constructor instead of using copy as in Spark ML)
@@ -120,3 +120,4 @@ class LightGBMClassificationModel(
 }
 
 object LightGBMClassificationModel extends ConstructorReadable[LightGBMClassificationModel]
+  with DefaultParamsReadable[LightGBMClassificationModel]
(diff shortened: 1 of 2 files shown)


src/lightgbm/src/main/scala/LightGBMClassifier.scala
```diff
@@ -73,7 +73,7 @@ class LightGBMClassificationModel(
   featuresColName: String, predictionColName: String, probColName: String,
   rawPredictionColName: String, thresholdValues: Option[Array[Double]])
     extends ProbabilisticClassificationModel[Vector, LightGBMClassificationModel]
-    with ConstructorWritable[LightGBMClassificationModel] {
+    with ConstructorWritable[LightGBMClassificationModel] with DefaultParamsWritable {
```
mhamilton723: This should either use one or the other. If it uses Constructor Writable it should be parameterized by the constructor alone, if it uses DefaultParamsWritable it should have an empty constructor and use SparkML's Parameter class to pass the state on to the model. Please let me know if you need me to explain in more depth!

src/lightgbm/src/main/scala/LightGBMClassifier.scala
```diff
@@ -120,3 +120,4 @@ class LightGBMClassificationModel(
 }
 
 object LightGBMClassificationModel extends ConstructorReadable[LightGBMClassificationModel]
+  with DefaultParamsReadable[LightGBMClassificationModel]
```
mhamilton723: Likewise with the comment above

src/lightgbm/src/main/scala/LightGBMClassifier.scala
```diff
@@ -25,7 +25,7 @@ object LightGBMClassifier extends DefaultParamsReadable[LightGBMClassifier]
 @InternalWrapper
 class LightGBMClassifier(override val uid: String)
   extends ProbabilisticClassifier[Vector, LightGBMClassifier, LightGBMClassificationModel]
-  with LightGBMParams {
+  with LightGBMParams with DefaultParamsWritable {
```
mhamilton723: This looks like a proper usage of the trait

src/lightgbm/src/main/scala/LightGBMRegressor.scala
```diff
@@ -91,7 +91,7 @@ class LightGBMRegressor(override val uid: String)
 class LightGBMRegressionModel(override val uid: String, model: LightGBMBooster, labelColName: String,
                                   featuresColName: String, predictionColName: String)
   extends RegressionModel[Vector, LightGBMRegressionModel]
-    with ConstructorWritable[LightGBMRegressionModel] {
+    with ConstructorWritable[LightGBMRegressionModel] with DefaultParamsWritable {
```
mhamilton723: Same as above

src/lightgbm/src/main/scala/LightGBMRegressor.scala
```diff
@@ -112,3 +112,4 @@ class LightGBMRegressionModel(override val uid: String, model: LightGBMBooster,
 }
 
 object LightGBMRegressionModel extends ConstructorReadable[LightGBMRegressionModel]
+  with DefaultParamsReadable[LightGBMClassificationModel]
```
mhamilton723: Same as above

terrytangyuan:
@mhamilton723 @imatiach-msft Thanks for your comments! Your comments make sense. I'll dig a bit more to see if I can get a repro. Closing this for now and I'll open a separate issue if I can get a repro.  I am working in Scala BTW.

mhamilton723:
Thanks for your investigations @terrytangyuan let me know if you get stuck and we can skype to help out

PR #349, opened 2018-07-17
Sample code for web service

jaymathe:
Using mmlspark serving


mhamilton723:
Updated this PR in #350

PR #352, opened 2018-07-19
changed port and added example

inchiosa:
1. Changed port from 8888 to 8898 to not conflict with Jupyter port.
2. Removed label column from test example
3. Added a second test example that produces a response of 1.0
4. Leaving output cells in notebook, so folks can see what the notebook is supposed to produce.


mhamilton723:
Thanks a ton for adding this @inchiosa. DO you think you could remove the notbook output for consistency's sake. The general idea of keeping notbook ouputs for legibility is a good one, but we will need to do a bit of refactoring work to make this work for all notebooks.  If you need I can do that before the merge as well!

Issue #353, opened 2018-08-04
Tweedie for LightGBM


nightflight-dk:
Hello please add support for Tweedie and regressor export (only classifier can be saved today), thanks

imatiach-msft:
@nightflight-dk tweedie is already available as one of the options:

https://github.com/Azure/mmlspark/blob/master/src/lightgbm/src/main/scala/LightGBMRegressor.scala#L47

Regressor export has already been added here:

https://github.com/Azure/mmlspark/pull/333/files

but only after the official v0.13 release, you can use a custom build instead however, eg:

--packages com.microsoft.ml.spark:mmlspark_2.11:0.13.dev18+3.gab13363
--repositories https://mmlspark.azureedge.net/maven


Hope this helps!

nightflight-dk:
Thank you for a quick response Ilya, for full Tweedie support please add the option to configure the variance power vs keeping the default 1.5. Early stopping rounds and verbosity during the training would be welcome as well. We could followup internally if you prefer, my alias dakowalc

imatiach-msft:
@nightflight-dk thanks, I've sent a PR to add the variance power parameter to LightGBM regressor.
I will look into adding early stopping and verbosity as well.

imatiach-msft:
@nightflight-dk  The PR built successfully (and the test I added also passed), you can use the custom build with the following params:

--packages com.microsoft.ml.spark:mmlspark_2.11:0.13.dev21+1.gca24275
--repositories https://mmlspark.azureedge.net/maven.

Let me know if you get a chance to test it out, thanks!

nightflight-dk:
Impressive resolve time Ilya, thank you. I can confirm the setters now in use on my cluster.

Unfortunately  I am running into allocation errors and timing-out / lost workers, java.net.ConnectException: Connection refused (Connection refused)

Even on a large cluster with small dataset (96 cores, <1M data points with 20 features) but I suppose it’s not related to your changes.

Any ETA for GPU support in the Spark version? Distributed multi-GPU training on partitioned RDD’s would be a fantastic addition.

Thanks, Damian

Cloud + AI
Market Intelligence
dakowalc@microsoft.com<mailto:dakowalc@microsoft.com>

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

imatiach-msft:
@nightflight-dk closing this issue as tweedie variance power parameter has been merged in

[...]

PR #354, opened 2018-08-06
added tweedie variance power to LightGBM

imatiach-msft:
added tweedie variance power parameter to LightGBM regressor.
Note tweedie_variance_power >= 1.0 and < 2.0
Also added test which uses cross validator.
resolves issue #353


drdarshan:
Looks good!

PR #355, opened 2018-08-13
Adding getBestModel and getBestModelInfo to TuneHyperparametersModel

imatiach-msft:
Adding getBestModel and getBestModelInfo to TuneHyperparametersModel (python API)

### DIFF
--- a/src/codegen/src/main/scala/WrapperGenerator.scala
+++ b/src/codegen/src/main/scala/WrapperGenerator.scala
@@ -23,6 +23,8 @@ abstract class WrapperGenerator {
 
   def wrapperName(myClass: Class[_]): String
 
+  def modelWrapperName(myClass: Class[_], modelName: String): String
+
   def generateEstimatorWrapper(entryPoint: Estimator[_],
                                entryPointName: String,
                                entryPointQualifiedName: String,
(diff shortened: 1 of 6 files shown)


src/codegen/src/main/scala/WrapperGenerator.scala
```diff
@@ -59,16 +61,18 @@ abstract class WrapperGenerator {
...
             val typeArgs = sc.getGenericSuperclass.asInstanceOf[ParameterizedType]
               .getActualTypeArguments
             val getModelFromGenericType = (modelType: Type) => {
               val modelClass = modelType.getTypeName.split("<").head
-              (modelClass.split("\\.").last, modelClass)
+              (modelWrapperName(myClass, modelClass.split("\\.").last), modelClass)
             }
             val (modelClass, modelQualifiedClass) = sc.getSuperclass.getSimpleName match {
               case "Estimator" => getModelFromGenericType(typeArgs.head)
+              case "ProbabilisticClassifier" => getModelFromGenericType(typeArgs(2))
```
mhamilton723: Why do we need to specifically code against this API here? seems like this muddies the abstraction and is not needed becase a Probabbalistic Classifier is a predictor

src/codegen/src/main/scala/WrapperGenerator.scala
```diff
@@ -23,6 +23,8 @@ abstract class WrapperGenerator {
 
   def wrapperName(myClass: Class[_]): String
 
+  def modelWrapperName(myClass: Class[_], modelName: String): String
+
```
mhamilton723: does this need to also take in the modelName?

src/lightgbm/src/main/python/LightGBMRegressor.py
```diff
@@ -9,7 +9,7 @@
     basestring = str
 
 from mmlspark._LightGBMRegressor import _LightGBMRegressor
-from mmlspark._LightGBMRegressor import M
+from mmlspark._LightGBMRegressor import _LightGBMRegressionModel
```
mhamilton723: This is nice

imatiach-msft: thanks

src/codegen/src/main/scala/WrapperGenerator.scala
imatiach-msft: the model wrapper needs to take in the model name because that is what it returns, it uses the class to determine whether to put an underscore at the beginning

src/codegen/src/main/scala/WrapperGenerator.scala
imatiach-msft: I needed probabilistic classifier to get the model type, otherwise in the autogenerated lightgbm wrappers the model was "M" instead of lightgbmclassificationmodel

PR #356, opened 2018-08-14
added model downloader R wrapper

imatiach-msft:
added model downloader R wrapper

### DIFF
--- a/src/downloader/src/main/R/model_downloader.R
+++ b/src/downloader/src/main/R/model_downloader.R
@@ -1,0 +1,157 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in project root for information.
+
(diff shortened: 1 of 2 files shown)


src/downloader/src/main/R/model_downloader.R
```diff
@@ -0,0 +1,49 @@
...
+#' Downloads the model by given name
+#'
+#' @param sdf_model_downloader The model downloader
+#' @param name The name of the model
+#' @param ... Optional arguments; currently unused.
+#'
+#' @family Model downloader
+#'
+#' @export
+sdf_download_by_name <- function(model_downloader, name, ...) {
```
mhamilton723: why the sdf prefix?

imatiach-msft: just convention

imatiach-msft: changed prefix to smd_...

PR #361, opened 2018-08-25
Add ImageLIME

mhamilton723:


src/codegen/src/test/scala/CogegenTest.scala
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import com.microsoft.ml.spark.codegen.CodeGen
+
+class CogegenTest extends TestBase {
```
imatiach-msft: minor comment: CodegenTest (same with file)

PR #362, opened 2018-08-26
Add Vagrantfile for Windows Developer Setup

mhamilton723:


drdarshan:
Holy smokes! This is awesome! Does it actually work as you'd expect? Happy to try it out and write some docs for you on Monday.

mhamilton723:
@drdarshan was able to get intellij up and running and can debug tests :)! I've also made an internal only use fully packaged box so that users dont need to figure out the magic stuff to get IJ to work. My advice is that the best way to use it is to run 

```vagrant ssh-config```

and use the output in moba so that you can forward IJ and other apps.

File: tools/vagrant/Vagrantfile [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+	#URL="https://download.jetbrains.com/idea/ideaI$ed-$VERSION.tar.gz"
+	echo $URL
+	# Truncate filename
+	FILE=$(basename ${URL})
+	# Set download directory
+	DEST=/home/vagrant/Downloads/$FILE
+	DIR=/home/vagrant/Downloads/idea
+	echo "Downloading idea-I$ed-$VERSION to $DEST..."
+
+	mkdir /home/vagrant/Downloads/
```
drdarshan: Suggest `mkdir -p`

tools/vagrant/Vagrantfile
```diff
@@ -0,0 +1,113 @@
...
+	# Fetch the most recent version
+	VERSION="173.4674.33"
+	# Prepend base URL for download
+	URL="https://www.jetbrains.com/intellij-repository/releases/com/jetbrains/intellij/idea/ideaI$ed/$VERSION/ideaI$ed-$VERSION.zip"
+	#URL="https://download.jetbrains.com/idea/ideaI$ed-$VERSION.tar.gz"
+	echo $URL
+	# Truncate filename
+	FILE=$(basename ${URL})
+	# Set download directory
+	DEST=/home/vagrant/Downloads/$FILE
```
drdarshan: Is this script run as the root user or as the `vagrant` user?
Use `$HOME` instead of `/home/vagrant` ?

tools/vagrant/Vagrantfile
```diff
@@ -0,0 +1,113 @@
...
+	# Grab executable folder
+	BIN="$DIR/bin"
+	# Add permissions to install directory
+	sudo chmod -R +rwx ${DIR}
+	# Create symlink entry
+	echo "alias idea=${BIN}/idea.sh" >> /home/vagrant/.bashrc
+	 echo "Done."	
+
+	# Install mmlspark
+	cd /home/vagrant
```
drdarshan: Avoid hard coded paths..

tools/vagrant/Vagrantfile
```diff
@@ -0,0 +1,113 @@
...
+  #
+  # View the documentation for the provider you are using for more
+  # information on available options.
+
+  # Enable provisioning with a shell script. Additional provisioners such as
+  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
+  # documentation for more information about their specific syntax and use.
+  config.vm.provision "shell", inline: <<-SHELL
+     apt-get update
+     apt-get update
```
drdarshan: Why doesn't runme install these dependencies?

tools/vagrant/Vagrantfile
mhamilton723: yes its run as root so $HOME is /home/root

tools/vagrant/Vagrantfile
mhamilton723: some are for Intellij, others runme will stop and tell you to add these libs.

MichelleC2386:
Want to learn more about codes and content

PR #384, opened 2018-10-03
Use scala sys API in utils

### LINKED ISSUE #247: Get rid of `getEnv` and all `System.getenv`
More generally switch to using `sys` instead of `System`.

Specifically: `src/core/env/src/main/scala/EnvironmentUtils.scala`

haal:
Resolves #247

### DIFF
--- a/src/core/env/src/main/scala/EnvironmentUtils.scala
+++ b/src/core/env/src/main/scala/EnvironmentUtils.scala
@@ -49,12 +49,4 @@ object EnvironmentUtils {
     }
   }
 
-  def getEnv(s:String): String = {
-    System.getenv(s) match {
-      case null =>
-        throw new IllegalArgumentException(s"Environment variable $s not found")
-      case v => v
-    }
-  }
-
 }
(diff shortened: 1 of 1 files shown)


imatiach-msft:
@mhamilton723 the image search suite tests that were added recently seem to be failing often, please take a look

mhamilton723:
Sorry Bing changed its outputs and the test was a bit too static. I made it a bit more lenient and am quing a build for the fix now! will be ready by morning london time

imatiach-msft:
@haal I've rebased this branch against latest master with fix from @mhamilton723 , will start another build

src/core/env/src/main/scala/EnvironmentUtils.scala
```diff
@@ -18,7 +18,7 @@ object EnvironmentUtils {
     val nvsmicmd = if (IsWindows) {
       // Unlikely nvidia is on the path
       val nvsmi = Paths.get(
-        System.getenv("ProgramFiles"),
+        sys.env.getOrElse("ProgramFiles", null),
```
mhamilton723: this should just be sys.env.get to keep with the same semantics. The old code threw an error if program files were not found this will fail silently.

mhamilton723:
Thank you @haal for this simplification. Just a minor nit and its ready to go

src/core/env/src/main/scala/EnvironmentUtils.scala
haal: wanted to have same behaviour on this line as System.getenv - returning value or null, but I can change it to throw exception if variable not found (java.util.NoSuchElementException)

PR #401, opened 2018-10-17
referred to glibc jar

imatiach-msft:
temporary fix for @land1725


imatiach-msft:
closing old temporary PR

PR #404, opened 2018-10-22
use conf to get num cores per executor

imatiach-msft:
We should use a combination of spark conf and number of cores per machine via java.lang.Runtime.getRuntime.availableProcessors to figure out the number of cores per executor.
Fixed based on discussion here: https://github.com/Azure/mmlspark/issues/379
Thank to @humbinal for figuring out how to fix the issue.
Also should resolve the issue brought by @ywskycn here: https://github.com/Azure/mmlspark/issues/346

Note: This may be eventually resolved in a better way with spark's barrier execution mode when it is implemented aka project hydrogen (without needing to compute num cores):
https://jira.apache.org/jira/browse/SPARK-24374

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMUtils.scala
+++ b/src/lightgbm/src/main/scala/LightGBMUtils.scala
@@ -132,8 +132,20 @@ object LightGBMUtils {
     */
   def getNumCoresPerExecutor(dataset: Dataset[_]): Int = {
     val spark = dataset.sparkSession
-    import spark.implicits._
-    spark.range(0, 1).map(_ => java.lang.Runtime.getRuntime.availableProcessors).collect.head
+    try {
+      val confCores = spark.sparkContext.getConf
+        .get("spark.executor.cores").toInt
+      val confTaskCpus = spark.sparkContext.getConf
+        .get("spark.task.cpus", "1").toInt
+      confCores / confTaskCpus
+    } catch {
+      case _: NoSuchElementException =>
+        // If spark.executor.cores is not defined, get the cores per JVM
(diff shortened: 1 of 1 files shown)


ywskycn:
@imatiach-msft I think we also need to include "spark.task.cpus" which is the number of cores to allocate for each task. "spark.executor.cores"/"spark.task.cpus" is the number of tasks in one executor.

imatiach-msft:
@ywskycn done!  I used spark.executer.cores/spark.task.cpus.  It looks like task cpus is usually default to 1 anyway (or not defined, as on my databricks cluster).  Hopefully this will handle all cluster scenarios (other than possibly the spark standalone cluster with different # of cores per executor, which I haven't encountered yet from users).

ywskycn:
LGTM

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -133,7 +133,13 @@ object LightGBMUtils {
   def getNumCoresPerExecutor(dataset: Dataset[_]): Int = {
     val spark = dataset.sparkSession
     import spark.implicits._
-    spark.range(0, 1).map(_ => java.lang.Runtime.getRuntime.availableProcessors).collect.head
+    val numMachineCores = spark.range(0, 1)
```
mhamilton723: Looks good! @imatiach-msft is it possible to remove the numMachineCores calculation and just use the confCores calc?

imatiach-msft: updated based on our discussion (put the conf code inside a try/catch)

PR #415, opened 2018-10-30
Ranking evaluator and adapater

dciborow:


src/recommendation/src/main/scala/RankingAdapter.scala
```diff
@@ -0,0 +1,382 @@
...
+    "min ratings for users > 0",
+    ParamValidators.inRange(0, Integer.MAX_VALUE))
+
+  val minRatingsPerItem: IntParam = new IntParam(this,
+    "minRatingsPerItem",
+    "min ratings for items > 0",
+    ParamValidators.inRange(0, Integer.MAX_VALUE))
+
+  /** @group setParam */
+  def setMinRatingsPerUser(value: Int): this.type = set(minRatingsPerUser, value)
```
imatiach-msft: small style: can you put the get/set together under the param, it is easier to read that way

src/recommendation/src/main/scala/RankingAdapter.scala
```diff
@@ -0,0 +1,382 @@
...
+
+}
+
+trait RankingFunctions extends RankingParams with HasRecommenderCols {
+
+  private def filterByItemCount(dataset: Dataset[_], itemCol: String, userCol: String): DataFrame =
+    dataset
+      .groupBy(userCol)
+      .agg(col(userCol), count(col(itemCol)).alias("nitems"))
+      .where(col("nitems") >= $(minRatingsPerUser))
```
imatiach-msft: would be nice to move these to some constants (this is scala and not python so there will be compile time errors instead of runtime so it makes sense to have constants)

src/recommendation/src/main/scala/RankingAdapter.scala
```diff
@@ -0,0 +1,382 @@
...
+      val windowSpec = Window.partitionBy(userColumn).orderBy(col($(itemCol)).desc)
+
+      validationDataset
+        .select(userColumn, itemColumn)
+        .withColumn("rank", r().over(windowSpec).alias("rank"))
+        .where(col("rank") <= k)
+        .groupBy(userColumn)
+        .agg(col(userColumn), collect_list(col(itemColumn)))
+        .withColumnRenamed("collect_list(" + itemColumn + ")", "label")
+        .select(userColumn, "label")
```
imatiach-msft: you should use the labelCol param instead of having hard-coded value here, so user can specify the label col

src/recommendation/src/main/scala/RankingAdapter.scala
```diff
@@ -0,0 +1,382 @@
...
+        .add("recommendations", ArrayType(
+          new StructType().add("itemCol", IntegerType).add("rating", FloatType))
+        )
+      case "allItems" => new StructType()
+        .add("itemCol", IntegerType)
+        .add("recommendations", ArrayType(
+          new StructType().add("userCol", IntegerType).add("rating", FloatType))
+        )
+      case "normal"   => new StructType()
+        .add("userCol", IntegerType)
```
imatiach-msft: hmm, there are a lot of hard-coded string here, I think some of these should be constants, others params?

src/recommendation/src/main/scala/RankingEvaluator.scala
```diff
@@ -0,0 +1,162 @@
...
+import org.apache.spark.ml.param._
+import org.apache.spark.ml.recommendation.{HasRecommenderCols, RecEvaluatorParams}
+import org.apache.spark.ml.util.{ComplexParamsReadable, ComplexParamsWritable, Identifiable}
+import org.apache.spark.mllib.evaluation.RankingMetrics
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{Dataset, Row}
+
+import scala.reflect.ClassTag
+
+class AdvancedRankingMetrics[T: ClassTag](predictionAndLabels: RDD[(Array[T], Array[T])],
```
imatiach-msft: is it necessary for this to be an RDD instead of a DataFrame?

src/recommendation/src/main/scala/RankingEvaluator.scala
```diff
@@ -0,0 +1,162 @@
...
+  /** @group setParam */
+  def setPredictionCol(value: String): this.type = set(predictionCol, value)
+
+  def getMetrics(dataset: Dataset[_]): AdvancedRankingMetrics[Any] = {
+    val predictionAndLabels = dataset
+      .select(getPredictionCol, getLabelCol)
+      .rdd.map { case Row(prediction: Seq[Any], label: Seq[Any]) => (prediction.toArray, label.toArray) }
+      .cache()
+    new AdvancedRankingMetrics[Any](predictionAndLabels, getK, getNItems)
+    //TODO come back to this Any type
```
imatiach-msft: why is any type needed here?

dciborow: When I try to switch everything to Ints (which is what they are instead of any) I get this...
image

src/recommendation/src/main/scala/RankingEvaluator.scala
dciborow: RankingMetrics requires RDD so I am not sure if this can be switched without creating a RankingMetrics DF version.

src/recommendation/src/main/scala/RankingAdapter.scala
dciborow: I don't understand. Should $(minRatingsPerUser) be a val? or should nitems be a constant?

src/recommendation/src/main/scala/RankingEvaluator.scala
mhamilton723: Yes its necessary :/

src/recommendation/src/main/scala/RankingAdapter.scala
```diff
@@ -0,0 +1,160 @@
...
+    val perUserActualItemsDF = dataset
+      .withColumn("rank", r().over(windowSpec).alias("rank"))
+      .where(col("rank") <= getK)
+      .groupBy(getUserCol)
+      .agg(col(getUserCol), collect_list(col(getItemCol)).as(getLabelCol))
+      .select(getUserCol, getLabelCol)
+
+    val recs = getMode match {
+      case "allUsers" => this.recommendForAllUsers(getK)
+      case "allItems" => this.recommendForAllItems(getK)
```
mhamilton723: you had mentioned that allItems did not work. Is this still the case?

src/recommendation/src/main/scala/RecommendationHelper.scala
```diff
@@ -0,0 +1,70 @@
...
+  val ratingCol = new Param[String](this, "ratingCol", "Column of ratings")
+
+  /** @group setParam */
+  def setRatingCol(value: String): this.type = set(ratingCol, value)
+
+  def getRatingCol: String = $(ratingCol)
+
+}
+
+trait kTrait extends Params{
```
mhamilton723:
```suggestion
trait HasK extends Params{
```

src/recommendation/src/main/scala/RecommendationHelper.scala
```diff
@@ -0,0 +1,70 @@
...
+    val arrayType = ArrayType(
+      new StructType()
+        .add(dstOutputColumn, IntegerType)
+        .add("rating", FloatType)
+    )
+    recs.select(col("id").as(srcOutputColumn), col("recommendations").cast(arrayType))
+  }
+}
+
+trait PublicALSParams extends ALSParams
```
mhamilton723: Is this to get around Spark private namespacing?

dciborow: yes.

src/recommendation/src/main/scala/RankingAdapter.scala
dciborow: they all work.

dciborow: It's gone. Never see it used anyway.

PR #432, opened 2018-11-09
Update to spark 2.4.0

mhamilton723:


imatiach-msft:
@mhamilton723 this is awesome!  I would really like to try to update lightgbm to use the new barrier execution mode as soon as this is in (and test out barrier execution, I don't understand all of the details on how it works yet).

PR #459, opened 2018-12-18
Azure search

caseyhong:


mhamilton723:
This build failed because of scalastyle

src/io/http/src/test/scala/services/SearchWriterSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  test("run asserts") {
+    runAsserts(Map(
+      "test-0" -> 4,
+      "test-1" -> 10,
+      "test-2" -> 100000))
+  }
+
+//  // only use when you're done with testing/need to clear all of the indices within the service
```
mhamilton723: we should probably clean up after this testsuite but do it in a more surgical manner
aka create an index then delete it if its been created

mhamilton723: You can do this with beforeAll and afterAll functions

PR #463, opened 2018-12-27
fix categorical handling logic to use slot names and indexes

imatiach-msft:
fixes issue https://github.com/Azure/mmlspark/issues/367
the previous fix was actually incorrect because it was for columns and not slots on the features column (very bad mistake!)
the user can now specify either the slot index within the vector features column or the slot name (but the slot name may not always be there, eg for sparse case it is removed in FastVectorAssembler to prevent OOM on very sparse data)

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMRegressor.scala
+++ b/src/lightgbm/src/main/scala/LightGBMRegressor.scala
@@ -74,15 +74,16 @@ class LightGBMRegressor(override val uid: String)
      * translate the data to the LightGBM in-memory representation and train the models
      */
     val encoder = Encoders.kryo[LightGBMBooster]
-    val categoricals =
-      if (isDefined(categoricalColumns)) getCategoricalColumns
-      else Array.empty[String]
-    val categoricalIndexes = LightGBMUtils.getCategoricalIndexes(df, categoricals)
+
+    val categoricalSlotIndexesArr = get(categoricalSlotIndexes).getOrElse(Array.empty[Int])
+    val categoricalSlotNamesArr = get(categoricalSlotNames).getOrElse(Array.empty[String])
+    val categoricalIndexes = LightGBMUtils.getCategoricalIndexes(df, getFeaturesCol,
+      categoricalSlotIndexesArr, categoricalSlotNamesArr)
     val trainParams = RegressorTrainParams(getParallelism, getNumIterations, getLearningRate, getNumLeaves,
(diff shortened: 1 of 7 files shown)


src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -58,11 +60,34 @@ object LightGBMUtils {
...
+    val categoricalSlotNamesSet = HashSet(categoricalColumnSlotNames: _*)
+    val featuresSchema = df.schema(featuresCol)
+    val metadata = AttributeGroup.fromStructField(featuresSchema)
+    val categoricalIndexes =
+      if (metadata.attributes.isEmpty) Array[Int]()
+      else {
+        metadata.attributes.get.zipWithIndex.flatMap {
+          case (attr, idx) =>
+            if (attr == null) {
+              Iterator()
```
mhamilton723: this can be part of the case match 

``` case (null, idx) => Iterator()```

File: src/lightgbm/src/main/scala/LightGBMRegressor.scala [hunk 2]
```diff
@@ -74,15 +74,19 @@ class LightGBMRegressor(override val uid: String)
      * translate the data to the LightGBM in-memory representation and train the models
      */
     val encoder = Encoders.kryo[LightGBMBooster]
-    val categoricals =
```
mhamilton723: can replace with 

```get(CategoricalColumns).getOrElse(Array.empty[Int])```

src/lightgbm/src/main/scala/LightGBMUtils.scala
imatiach-msft: done!

XinyunTang:
Is this supported in pyspark? I just got an error saying that "categorical_feature" isn't a parameter in pyspark w/ lightGBM

imatiach-msft:
@XinyunTang yes, categoricals are supported.  For lightgbm you can specify categoricalSlotIndexes or categoricalSlotNames.  You can run help(your_lgbm_learner) to see the parameters and doc.  You can also see the autogenerated doc here: https://mmlspark.azureedge.net/docs/pyspark/LightGBMClassifier.html#

imatiach-msft:
in future please create an issue, it's easy to miss comments on closed PRs

XinyunTang:
Thanks! the help command is really helpful. You guys should definitely put the same information on git or doc. The current documentation is super limited. The examples are not clear either. It's a shame that you did so much work but people got confused because of the lack of support documentation.

PR #469, opened 2019-01-04
Sar

dciborow:


dciborow:
@mhamilton723 @imatiach-msft ready for review!

src/core/serialize/src/main/scala/params/DataFrameParam2.scala
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package org.apache.spark.ml.param
+
+import org.apache.spark.sql.{DataFrame, SparkSession}
+
+/** Param for DataFrame.  Needed as spark has explicit params for many different
+  * types but not DataFrame.
+  */
+class DataFrameParam2(parent: Params, name: String, doc: String, isValid: DataFrame => Boolean)
```
imatiach-msft: why is this called DataFrameParam2 instead of just DataFrameParam?

src/core/serialize/src/main/scala/params/ArrayParamMapParam.scala
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package org.apache.spark.ml.param
+
+/** Param for Estimator.  Needed as spark has explicit params for many different
```
imatiach-msft: update the doc (it looks like this was copy pasted from estimator param)

src/recommendation/src/main/python/RankingTrainValidationSplit.py
```diff
@@ -0,0 +1,83 @@
...
+            extra = dict()
+        newCV = Params.copy(self, extra)
+        if self.isSet(self.estimator):
+            newCV.setEstimator(self.getEstimator().copy(extra))
+        # estimatorParamMaps remain the same
+        if self.isSet(self.evaluator):
+            newCV.setEvaluator(self.getEvaluator().copy(extra))
+        return newCV
+
+    def _create_model(self, java_model):
```
imatiach-msft: if you are not overriding the model, which it looks like you are not, then this is already handled in the base class right?

imatiach-msft: oh, it looks like you put the model in a separate file and you are overriding it.  I think we usually put them in the same file.

src/recommendation/src/main/scala/RankingTrainValidationSplit.scala
```diff
@@ -0,0 +1,343 @@
...
+}
+
+object RankingTrainValidationSplit extends ComplexParamsReadable[RankingTrainValidationSplit] {
+  def popRow(r: Row): Any = r.getDouble(1)
+}
+
+@InternalWrapper
+class RankingTrainValidationSplitModel(
+  override val uid: String)
+  extends Model[RankingTrainValidationSplitModel] with RankingTrainValidationSplitParams
```
imatiach-msft: I don't think this is using all of those params right?  Those params are only on the estimator.  I don't think you need to include them on the model.  Also, since they are only on the estimator, you can move the set* methods to the params file next to the getters.

src/recommendation/src/main/scala/RankingTrainValidationSplit.scala
```diff
@@ -0,0 +1,343 @@
...
+  override def copy(extra: ParamMap): RankingTrainValidationSplitModel = {
+    val copied = new RankingTrainValidationSplitModel(uid)
+    copyValues(copied, extra).setParent(parent)
+  }
+
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    transformSchema(dataset.schema, logging = true)
+
+    //sort to pass unit test
+    $(bestModel).transform(dataset).sort("prediction")
```
imatiach-msft: can you explain this a bit more, why do you need the sort here?

src/recommendation/src/main/scala/RecommendationHelper.scala
```diff
@@ -33,6 +40,95 @@ object SparkHelper {
 
 trait RecommendationParams extends ALSParams
 
+trait RecommendationModelParams extends Params with ALSModelParams with HasPredictionCol {
```
imatiach-msft: it looks like this holds not only params but an interface - I would change the name of the trait then to make it more explicit that this isn't only for params.

src/recommendation/src/main/scala/RecommendationHelper.scala
```diff
@@ -33,6 +40,95 @@ object SparkHelper {
...
+    val sc = userDataFrame.sparkSession
+
+    val zipVector = udf((list: DenseVector) => {
+      val zipped = list.toArray.zipWithIndex
+      val sorted = zipped.map(f => (f._2, f._1)).sortWith(_._2 > _._2)
+      sorted.take(k)
+    })
+
+    val recs = sc.createDataFrame(userToItemMatrix)
+      .toDF("id", "ratings").withColumn("recommendations", zipVector($"ratings")).select("id", "recommendations")
```
imatiach-msft: in general it's best to move these kinds of common strings, especially if they repeat several times, to a constants file - maybe you could add a common constants file for recommenders and put them there

src/recommendation/src/test/python/RankingSpec.py
```diff
@@ -90,6 +90,55 @@ def test_adapter_evaluator(self):
...
+            .setTrainRatio(0.8) \
+            .setUserCol(customerIndex.getOutputCol()) \
+            .setRatingCol('rating') \
+            .setItemCol(ratingsIndex.getOutputCol())
+
+        tvmodel = tvRecommendationSplit.fit(transformedDf)
+
+        usersRecs = tvmodel.bestModel._call_java("recommendForAllUsers", 3)
+
+        print(usersRecs.take(1))
```
imatiach-msft: try to remove prints

imatiach-msft: maybe use some asserts instead?

imatiach-msft:
@dciborow this looks really good, I added a couple comments here and there.  Will take another look especially at the SAR logic later.  Does the build pass?  It looks like you haven't run a build yet.

src/recommendation/src/main/scala/RankingTrainValidationSplit.scala
dciborow: in order to pass a unit test.... if I don't sort the matching dataframe unit test fails.

src/core/serialize/src/main/scala/params/DataFrameParam.scala
```diff
@@ -0,0 +1,29 @@
...
+/** Param for DataFrame.  Needed as spark has explicit params for many different
+  * types but not DataFrame.
+  */
+class DataFrameParam(parent: Params, name: String, doc: String, isValid: DataFrame => Boolean)
+  extends ComplexParam[DataFrame](parent, name, doc, isValid) {
+
+  def this(parent: Params, name: String, doc: String) =
+    this(parent, name, doc, ParamValidators.alwaysTrue)
+
+  override def jsonEncode(value: DataFrame): String = {
```
mhamilton723: How big is this dataframe going to be? If its anything more than a few lines we should modify this to wrap and unwrap the dataframe in python

src/recommendation/src/main/scala/RecommendationHelper.scala
```diff
@@ -33,6 +37,92 @@ object SparkHelper {
...
+  private val recommendations = Constants.recommendations
+
+  def getALSModel(uid: String,
+    rank: Int,
+    userFactors: DataFrame,
+    itemFactors: DataFrame): ALSModel = {
+    new ALSModel(uid, rank, userFactors, itemFactors)
+  }
+
+  def recommendForAllItems(k: Int): DataFrame = ???
```
mhamilton723: just leave the impls blank, no need for ???.

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,484 @@
...
+
+    df.select(col(itemColumn).cast(IntegerType), col(features))
+      .withColumn("jaccardList", calculateFeature(col(itemColumn), col(features)))
+      .select(
+        col(itemColumn).cast(IntegerType),
+        col("jaccardList")
+      )
+  }
+
+  def isLarge(): Boolean = true //todo: add real is large check here
```
mhamilton723: What is this for?

src/recommendation/src/test/scala/SARSpec.scala
```diff
@@ -0,0 +1,22 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import scala.language.existentials
+
+class SARSpec extends RankingTestBase with EstimatorFuzzing[SAR] {
+  override def testObjects(): List[TestObject[SAR]] = {
```
mhamilton723: Can you add some tests that show that it improves over ALS on a dataset. This way we could sanity check. Also if there's a benchmark in a paper we could replicate that would be reassuring

src/core/serialize/src/main/scala/params/DataFrameParam.scala
dciborow: the biggest it could be would be 10mil rows and 40k columns... so that's probably big...

src/recommendation/src/main/scala/RecommendationHelper.scala
```diff
@@ -68,3 +158,11 @@ trait hasK extends Params{
 
   setDefault(k -> 10)
 }
+
+object Constants {
```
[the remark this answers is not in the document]
dciborow: @imatiach-msft recommending extracting these hard coded strings into constants

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,426 @@
...
+
+  override def fit(dataset: Dataset[_]): SARModel = {
+
+    //Create user and item affinity matricies
+    val (userAffinityMatrix, itemAffinityMatrix) =
+      if ($(autoIndex)) {
+
+        val customerIndex = new StringIndexer()
+          .setInputCol(getUserCol)
+          .setOutputCol(getUserCol + "temp")
```
mhamilton723: use temp col

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,426 @@
...
+
+        val itemIndex = new StringIndexer()
```
mhamilton723: Double Check that you are using the non-deprecated one

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,426 @@
...
+  }
+
+  def this() = this(Identifiable.randomUID("SAR"))
+
+  def calculateItemAffinities(df: Dataset[_]): DataFrame = {
+    SAR.calculateItemAffinities(getUserCol, getItemCol, getSupportThreshold, df, get(itemFeatures),
+      getSimilarityFunction)
+  }
+
+  def calculateUserAffinities(dataset: Dataset[_]): DataFrame = {
```
mhamilton723: Nit UserItem

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,426 @@
...
+    val referenceTime: Date = new SimpleDateFormat($(startTimeFormat))
+      .parse(get(startTime).getOrElse(Calendar.getInstance().getTime.toString))
+
+    val timeDecay = udf((time: String) => {
+      val activityDate = new SimpleDateFormat($(activityTimeFormat)).parse(time)
```
mhamilton723: Nit:Use getters

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,426 @@
...
+      .agg(collect_list(col("itemUserAffinityPair")))
+      .withColumn("flatList", flattenItems(col("collect_list(itemUserAffinityPair)")))
+      .select(col(getUserCol).cast(IntegerType), col("flatList"))
+  }
+}
+
+object SAR extends DefaultParamsReadable[SAR] {
+  private val features = "features"
+
+  def calculateItemAffinities(userColumn: String, itemColumn: String, supportThreshold: Int, transformedDf: Dataset[_],
```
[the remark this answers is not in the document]
dciborow: used the getter instead of passing this

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,426 @@
...
+        .toIndexedRowMatrix()
+        .rows.map(index => (index.index.toInt, index.vector))
+
+      val pairs =
+        itemFeatureMatrix.cartesian(itemFeatureMatrix)
+          .map(row => {
+            val vec1: linalg.Vector = row._1._2
+            val vec2 = row._2._2
+            //consider if not equal 0
+            val productArray = (0 to vec1.argmax + 1).map(i => vec1.apply(i) * vec2.apply(i)).toArray
```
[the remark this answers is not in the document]
dciborow: cold weighting has a bug and i've cut it from this PR

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,299 @@
...
+
+      def lift(countI: Double, countJ: Long, cooco: Double) = (cooco / (countI * countJ)).toFloat
+
+      def jaccard(countI: Double, countJ: Long, cooco: Double) = (cooco / (countI + countJ - cooco)).toFloat
+
+      val countI = features.apply(itemID)
+      features.toArray.indices.map(i => {
+        val countJ: Long = broadcastItemCounts.value.getOrElse(i, 0)
+        val cooco = features.apply(i)
+        if (!(cooco < getSupportThreshold)) {
```
mhamilton723: Nit for later if neccessary, you might be able to tighten this up with  higher order function. namely something like
```
val similarityLambda = getSimilarityFunction match {
  case "jaccard" => jaccard _
  case "lift" => jaccard _
  case _ => {case (_,_, cooco) => coocoo.toFloat}
}
```

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,299 @@
...
+
+      val sparse = SparseMatrix.fromCOO(userCount, itemCount, userItemPairs)
+      val sparseBSM: BSM[Double] = new BSM[Double](sparse.values, sparse.numRows, sparse.numCols, sparse.colPtrs,
+        sparse.rowIndices)
+      spark.sparkContext.broadcast(sparseBSM)
+    }
+
+    val makeDenseFeatureVectors = udf((wrappedList: Seq[Int]) => {
+      val vec = Array.fill[Double](userCount)(0.0)
+      wrappedList.foreach(user => vec(user) = 1.0)
```
mhamilton723: will this mutation work with a non mutable array?

src/recommendation/src/test/scala/SARSpec.scala
```diff
@@ -0,0 +1,302 @@
...
+    val answer = session.read.option("header", "true").csv(userPredFile).collect()
+
+    assert(rows(0).getString(0) == "0003000098E85347", "Assert Customer ID's Match")
+
+    assert(rows(0).getString(1) == answer(0).getString(1))
+    assert(rows(0).getString(2) == answer(0).getString(2))
+    assert(rows(0).getString(3) == answer(0).getString(3))
+    assert(rows(0).getString(4) == answer(0).getString(4))
+    assert(rows(0).getString(5) == answer(0).getString(5))
+    assert("%.3f".format(rows(0).getFloat(6)) == "%.3f".format(answer(0).getString(11).toFloat))
```
mhamilton723: if you are trying to do tolerant comparisions, take a look at the DataFrameEquality trait. It should have  row equality and a tolerant double equality thats very customizable. You hook into it with === instead of ==

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,299 @@
...
+        .collect()
+        .map(userItemPair => (userItemPair.getLong(0).toInt, userItemPair.getLong(1).toInt, 1.0))
+
+      val sparse = SparseMatrix.fromCOO(userCount, itemCount, userItemPairs)
+      val sparseBSM: BSM[Double] = new BSM[Double](sparse.values, sparse.numRows, sparse.numCols, sparse.colPtrs,
+        sparse.rowIndices)
+      spark.sparkContext.broadcast(sparseBSM)
+    }
+
+    val makeDenseFeatureVectors = udf((wrappedList: Seq[Int]) => {
```
mhamilton723: nit: name this something semantically relevant

src/recommendation/src/main/python/SAR.py
```diff
@@ -0,0 +1,23 @@
...
+from pyspark.ml.param.shared import *
+from pyspark.ml.common import inherit_doc
+from mmlspark.Utils import *
+from mmlspark._SAR import _SAR as sar
+from mmlspark.SARModel import SARModel as sarm
+
+
+@inherit_doc
+class SAR(sar):
+    def _create_model(self, java_model):
```
mhamilton723: because you took out the json encoding and decoding of parameters, you might need to add some wrapper code here to convert to and prom python dataframes.

src/recommendation/src/main/scala/SAR.scala
```diff
@@ -0,0 +1,426 @@
...
+
+    val timeDecay = udf((time: String) => {
+      val activityDate = new SimpleDateFormat($(activityTimeFormat)).parse(time)
+      val timeDifference = (referenceTime.getTime - activityDate.getTime) / (1000 * 60)
+      math.pow(2, -1.0 * timeDifference / ($(timeDecayCoeff) * 24 * 60))
+    })
+    val ratingWeight = udf((rating: Double) => rating)
+    val blendWeights = udf((theta: Double, roe: Double) => theta * roe)
+
+    val itemCount = dataset.select(col(getItemCol).cast(IntegerType)).groupBy().max(getItemCol).collect()(0).getInt(0)
```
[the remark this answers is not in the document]
dciborow: removing group by lead to dataframe does not have function max

src/recommendation/src/test/scala/SARSpec.scala
```diff
@@ -0,0 +1,235 @@
...
+          !bol
+        }).map(p => (p._1, ratings.toList(p._2)))
+    })
+
+    val row = model.recommendForAllUsers(10 + usersProducts.length)
+      .select(col("customerID"), filterScore(col("recommendations.itemID"), col("recommendations.rating")) as
+        "recommendations")
+      .select(col("customerID"), col("recommendations._1") as "itemID", col("recommendations._2") as "rating")
+      .select(
+        recoverUser(col("customerID")) as "customerID",
```
mhamilton723: for a future PR, you probabbly dont need this anymore

PR #495, opened 2019-02-18
Add Recommend For User Subset to SAR

dciborow:
Adding the following functions to SAR Model to align with ALS. 
RecommendForUserSubset


dciborow:
@mhamilton723 ready for review

PR #499, opened 2019-02-21
return list of param name to estimator and param from hyperparam builder

imatiach-msft:
resolves issue https://github.com/Azure/mmlspark/issues/473

HyperParam builder will now return a dictionary from param to (estimator, hyperparam), which is similar to parammap structure (parammap is dictionary from param to value but we can't quite do the same since we have a value generator and we need the estimator later on to extract the java param).
FYI @yueguoguo


yueguoguo:
Thanks @imatiach-msft! I will take it a go with your updates. 👍

PR #504, opened 2019-02-25
add evaluation data for early stopping

imatiach-msft:
fixes issue https://github.com/Azure/mmlspark/issues/435


ekerazha:
@imatiach-msft @mhamilton723 Is there anything preventing this PR to be merged?

imatiach-msft:
@ekerazha this PR doesn't work yet, it requires some additional changes (note the test is failing)

src/lightgbm/src/main/scala/LightGBMRegressor.scala
```diff
@@ -85,10 +85,13 @@ class LightGBMRegressor(override val uid: String)
       getVerbosity, categoricalIndexes, getBoostFromAverage, getBoostingType)
     log.info(s"LightGBMRegressor parameters: ${trainParams.toString}")
     val networkParams = NetworkParams(getDefaultListenPort, inetAddress, port)
-
+    val validationData =
+      if (get(validationIndicatorCol).isDefined && get(validationIndicatorCol).get.nonEmpty)
```
mhamilton723: will this ever be an empty string? I think if you use an empty string we should remove that API and just go with setting or not setting and then turn this into a map on an option

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -339,11 +339,14 @@ object LightGBMUtils {
     var data: Option[(SWIGTYPE_p_void, SWIGTYPE_p_double)] = None
     try {
       data = Some(generateData(numRows, rowsAsDoubleArray))
+      val reference =
```
mhamilton723: use .orNull on the option

src/lightgbm/src/main/scala/LightGBMUtils.scala
```diff
@@ -375,13 +379,16 @@ object LightGBMUtils {
       val datasetParams = "max_bin=255 is_pre_partition=True"
       val dataInt32bitType = lightgbmlibConstants.C_API_DTYPE_INT32
       val data64bitType = lightgbmlibConstants.C_API_DTYPE_FLOAT64
+      val reference =
+        if (referenceDataset.isDefined) referenceDataset.get
```
mhamilton723: orNull and inline as well

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -36,95 +36,172 @@ private object TrainUtils extends Serializable {
...
-          val booster = LightGBMUtils.getBoosterPtrFromModelString(trainParams.modelString)
-          LightGBMUtils.validate(lightgbmlib.LGBM_BoosterMerge(boosterPtr.get, booster), "Booster Merge")
+        val result = lightgbmlib.LGBM_BoosterUpdateOneIter(boosterPtr.get, isFinishedPtr)
+        LightGBMUtils.validate(result, "Booster Update One Iter")
+        isFinished = lightgbmlib.intp_value(isFinishedPtr)
+        log.info("LightGBM running iteration: " + iters + " with result: " +
+          result + " and is finished: " + isFinished)
+      } catch {
+        case e: java.lang.Exception => {
+          isFinished = 1
```
mhamilton723: dont use numbers to represent booleans

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -36,95 +36,172 @@ private object TrainUtils extends Serializable {
...
+                validationData: Option[Array[Row]], log: Logger,
+                trainParams: TrainParams, inputRows: Iterator[Row]): Iterator[LightGBMBooster] = {
+    if (!inputRows.hasNext)
+      List[LightGBMBooster]().toIterator
+    val rows = inputRows.toArray
+    var trainDatasetPtr: Option[SWIGTYPE_p_void] = None
+    var validDatasetPtr: Option[SWIGTYPE_p_void] = None
+    try {
+      val (trainRows, validRows) =
+        if (validationData.isDefined) {
```
mhamilton723: val (trainRows, validRows) = (rows, validationData)

isnt this what you are doing?

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -36,95 +36,172 @@ private object TrainUtils extends Serializable {
...
-          if (bufferOutLength > bufferLength) {
-            lightgbmlib.LGBM_BoosterSaveModelToStringSWIG(
-              boosterPtr.get, 0, -1, bufferOutLengthPtr, bufferOutLengthPtr)
-          } else tempM
-        log.info("Buffer output length for model: " + bufferOutLength)
+      trainDatasetPtr = generateDataset(trainRows, labelColumn, featuresColumn, weightColumn, None)
+      if (validRows.isDefined && validRows.get.nonEmpty) {
+        validDatasetPtr = generateDataset(validRows.get, labelColumn, featuresColumn, weightColumn, trainDatasetPtr)
+      }
+      var boosterPtr: Option[SWIGTYPE_p_void] = None
```
mhamilton723: please avoid vars if possible

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -36,95 +36,172 @@ private object TrainUtils extends Serializable {
...
-                                 "Finalize Booster")
+          LightGBMUtils.validate(lightgbmlib.LGBM_BoosterFree(boosterPtr.get), "Finalize Booster")
         }
       }
     } finally {
       // Free dataset
-      if (datasetPtr.isDefined) {
-        LightGBMUtils.validate(lightgbmlib.LGBM_DatasetFree(datasetPtr.get),
-                               "Finalize Dataset")
+      if (trainDatasetPtr.isDefined) {
```
mhamilton723: use option.foreach instead of if

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -36,95 +36,172 @@ private object TrainUtils extends Serializable {
...
+        LightGBMUtils.validate(lightgbmlib.LGBM_DatasetFree(trainDatasetPtr.get), "Finalize Train Dataset")
+      }
+      if (validDatasetPtr.isDefined) {
```
mhamilton723: use option.foreach

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -36,95 +36,172 @@ private object TrainUtils extends Serializable {
...
+                            bestScores: Array[Array[Double]], bestIter: Array[Int])
 
-      var boosterPtr: Option[SWIGTYPE_p_void] = None
+  def createBestEval(boosterPtr: Option[SWIGTYPE_p_void]): Option[BestEvaluation] = {
+    // Need to keep track of best scores for each metric, see callback.py in lightgbm for reference
+    val evalCountsPtr = lightgbmlib.new_intp()
+    val resultCounts = lightgbmlib.LGBM_BoosterGetEvalCounts(boosterPtr.get, evalCountsPtr)
+    LightGBMUtils.validate(resultCounts, "Booster Get Eval Counts")
+    val evalCounts = lightgbmlib.intp_value(evalCountsPtr)
+    val bestScore = new Array[Double](evalCounts)
```
mhamilton723: use maps instead of mutation

mhamilton723: why are you retuning an empty array of arrays? why not make an array later when you need it

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -36,95 +36,172 @@ private object TrainUtils extends Serializable {
...
+  }
+
+  def createBooster(trainParams: TrainParams, trainDatasetPtr: Option[SWIGTYPE_p_void],
+                    validDatasetPtr: Option[SWIGTYPE_p_void]): Option[SWIGTYPE_p_void] = {
+    // Create the booster
+    val boosterOutPtr = lightgbmlib.voidpp_handle()
+    val parameters = trainParams.toString()
+    LightGBMUtils.validate(lightgbmlib.LGBM_BoosterCreate(trainDatasetPtr.get, parameters, boosterOutPtr), "Booster")
+    val boosterPtr = Some(lightgbmlib.voidpp_value(boosterOutPtr))
+    if (trainParams.modelString != null && !trainParams.modelString.isEmpty) {
```
mhamilton723: will the model string ever be empty? If so we should try to remove that logic as early as possible and work with options

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -36,95 +36,172 @@ private object TrainUtils extends Serializable {
...
+      val booster = LightGBMUtils.getBoosterPtrFromModelString(trainParams.modelString)
+      LightGBMUtils.validate(lightgbmlib.LGBM_BoosterMerge(boosterPtr.get, booster), "Booster Merge")
+    }
+    if (validDatasetPtr.isDefined) {
```
mhamilton723: Use option.foreach

src/lightgbm/src/main/scala/LightGBMClassifier.scala
```diff
@@ -78,10 +78,13 @@ class LightGBMClassifier(override val uid: String)
       getIsUnbalance, getVerbosity, categoricalIndexes, classes, metric, getBoostFromAverage, getBoostingType)
     log.info(s"LightGBMClassifier parameters: ${trainParams.toString}")
     val networkParams = NetworkParams(getDefaultListenPort, inetAddress, port)
-
+    val validationData =
+      if (get(validationIndicatorCol).isDefined && get(validationIndicatorCol).get.nonEmpty)
```
mhamilton723: encode the nonempty check in param validator

src/core/contracts/src/main/scala/Params.scala
```diff
@@ -135,3 +135,14 @@ trait HasEvaluationMetric extends Wrappable {
   /** @group setParam */
   def setEvaluationMetric(value: String): this.type = set(evaluationMetric, value)
 }
+
+trait HasValidationIndicatorCol extends Wrappable {
+  val validationIndicatorCol: Param[String] =
+    new Param[String](this, "validationIndicatorCol", "Indicates whether the row is for training or validation")
```
mhamilton723: add a nonempty check here instead of throughout the code

src/lightgbm/src/main/scala/LightGBMRegressor.scala
imatiach-msft: it's just the column name. Supposedly it should never be an empty string but if the user for some bizarre reason sets it to empty string we can just ignore it.  Actually, this is a bad check.  I should instead check if the column exists on the dataset.  I modified the check.

src/lightgbm/src/main/scala/LightGBMUtils.scala
imatiach-msft: ah, that's very cool!  great suggestion.  done.

src/lightgbm/src/main/scala/LightGBMUtils.scala
imatiach-msft: good suggestion, done!

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: done!

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: good catch, I think I refactored this several times and forgot to remove this

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: what would be the better way to write this bit?  I need to try/finally and cleanup in case something failed, but if createBooster itself throws I shouldn't free the booster.

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: done!  fixed in some other places as well

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: done!

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: "why are you retuning an empty array of arrays? why not make an array later when you need it"
done
"use maps instead of mutation"
for different indexes you can have different best score/iteration, and we reuse the same array throughout so it doesn't make sense to allocate it more than once

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: done, changed to option of string

src/lightgbm/src/main/scala/TrainUtils.scala
imatiach-msft: done

src/lightgbm/src/main/scala/LightGBMClassifier.scala
imatiach-msft: it can be empty, we just don't use the validation column.  but I added a non-null check.

imatiach-msft: actually, I think it is fine if it is null, I modified this part of the code a bit differently anyway

src/core/contracts/src/main/scala/Params.scala
imatiach-msft: it can be empty, it just means we don't have a validation dataset

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -30,102 +30,167 @@ private object TrainUtils extends Serializable {
...
-        if (trainParams.modelString != null && !trainParams.modelString.isEmpty) {
-          val booster = LightGBMUtils.getBoosterPtrFromModelString(trainParams.modelString)
-          LightGBMUtils.validate(lightgbmlib.LGBM_BoosterMerge(boosterPtr.get, booster), "Booster Merge")
+        val result = lightgbmlib.LGBM_BoosterUpdateOneIter(boosterPtr.get, isFinishedPtr)
+        LightGBMUtils.validate(result, "Booster Update One Iter")
+        isFinished = lightgbmlib.intp_value(isFinishedPtr) == 1
+        log.info("LightGBM running iteration: " + iters + " with result: " +
+          result + " and is finished: " + isFinished)
+      } catch {
+        case e: java.lang.Exception => {
```
mhamilton723: Is there any way to make this catch a bit more specific?

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -30,102 +30,167 @@ private object TrainUtils extends Serializable {
...
+    val resultCounts = lightgbmlib.LGBM_BoosterGetEvalCounts(boosterPtr.get, evalCountsPtr)
+    LightGBMUtils.validate(resultCounts, "Booster Get Eval Counts")
+    val evalCounts = lightgbmlib.intp_value(evalCountsPtr)
+    val bestScore = new Array[Double](evalCounts)
+    val bestScores = new Array[Array[Double]](evalCounts)
+    val bestIter = new Array[Int](evalCounts)
+    // For debugging, can get metric names:
+    val evalNamesPtr = lightgbmlib.LGBM_BoosterGetEvalNamesSWIG(boosterPtr.get, evalCounts)
+    val evalNames = (0 until evalCounts).map(lightgbmlib.stringArray_getitem(evalNamesPtr, _)).toArray
+    Some(BestEvaluation(evalCounts, evalNames, bestScore, bestScores, bestIter))
```
mhamilton723: We should chat in person. You really just need eval names here

imatiach-msft: sure, feel free to refactor this more :)

PR #506, opened 2019-02-27
JNI performance improvements, mainly for prediction (3-4x)

eisber:
This depends on https://github.com/Microsoft/LightGBM/pull/2032


src/lightgbm/src/main/scala/LightGBMBooster.scala
```diff
@@ -35,81 +35,70 @@ class LightGBMBooster(val model: String) extends Serializable {
 
   lazy val numClasses: Int = getNumClasses()
 
+  @transient
+  var scoredDataOutPtr: SWIGTYPE_p_double = null
```
imatiach-msft: nice optimization!

src/lightgbm/src/main/scala/LightGBMBooster.scala
```diff
@@ -35,81 +35,70 @@ class LightGBMBooster(val model: String) extends Serializable {
...
-      if (data.isDefined) lightgbmlib.delete_doubleArray(data.get._2)
-      // Delete the output scored row
-      if (scoredDataOutPtr.isDefined) lightgbmlib.delete_doubleArray(scoredDataOutPtr.get)
-    }
+    val data64bitType = lightgbmlibConstants.C_API_DTYPE_FLOAT64
+
+    val numCols = row.length
+    val isRowMajor = 1
+
+    val datasetParams = "max_bin=255"
```
mhamilton723: question: what does this do?

imatiach-msft: Not related to this PR, but I just noticed the dataset param here is different from predictForCSR, which has is_pre_partition=True.
I think this is a bug.  Since the dataset is not distributed during prediction time, I think is_pre_partition shouldn't be set for CSR case during prediction, but I'm not sure why it isn't throwing some error right now.  Maybe it just ignores that param in this case, but we probably still shouldn't be setting it.

src/scalastyle-config.xml
```diff
@@ -67,7 +67,7 @@ package (?:com\.microsoft\.ml\.spark|org\.apache\.spark|com\.microsoft\.CNTK|com
   <check level="warning" class="org.scalastyle.scalariform.ReturnChecker" enabled="true"></check>
   <check level="warning" class="org.scalastyle.scalariform.NullChecker" enabled="true"></check>
   <check level="error" class="org.scalastyle.scalariform.NoCloneChecker" enabled="true"></check>
-  <check level="error" class="org.scalastyle.scalariform.NoFinalizeChecker" enabled="true"></check>
+  <check level="warning" class="org.scalastyle.scalariform.NoFinalizeChecker" enabled="true"></check>
```
mhamilton723: question: Why did this come up?

src/lightgbm/src/main/scala/LightGBMBooster.scala
imatiach-msft: https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.rst#max_bin

"max number of bins that feature values will be bucketed in"

imatiach-msft:
rebased PR has been merged

PR #518, opened 2019-03-25
[LightGBM] Ensure TrainUtils.translate() doesn't crash on an empty partition

rluta:
This patch ensures translate() method correctly stops processing when encountering an empty partition.

This prevents a crash when using LightGBM in python with CrossValidation and n_folds > 1.


imatiach-msft:
Good catch!  This is fixing a missing if/else, is that correct?
```
...
    if (!inputRows.hasNext) {
      List[LightGBMBooster]().toIterator
    } else {
...
```
The rest of the changes are just the tab indentation?

imatiach-msft:
Ah, it looks like the rest of the changes are just whitespace.  Merged to master after successful build.

rluta:
Indeed, the only change was the else clause on `if` statement. 
Everything else is whitespace noise from re-indentation.

Thanks for merging !

PR #524, opened 2019-03-27
UnicodeNormalize: handle null values

eisber:


src/pipeline-stages/src/main/scala/UnicodeNormalize.scala
```diff
@@ -49,10 +49,12 @@ class UnicodeNormalize(val uid: String) extends Transformer
...
     require(inputIndex != -1, s"Input column $getInputCol does not exist")
 
-    val normalizeFunc = (value: String) => Normalizer.normalize(value, Normalizer.Form.valueOf(getForm))
+    val normalizeFunc = (value: String) =>
+      if (value == null) null
+      else Normalizer.normalize(value, Normalizer.Form.valueOf(getForm))
 
     val f = if (getLower)
-      (value: String) => normalizeFunc(value.toLowerCase)
+      (value: String) => if (value == null) null else normalizeFunc(value.toLowerCase)
```
mhamilton723:
```suggestion
      (value: String) => Option(value).map(normalizeFunc(_.toLowerCase)).orNull 
```

eisber: for some reason the type cannot be resolve on _ . I only got it to work with

(value: String) => Option(value).map(s => normalizeFunc(s.toLowerCase)).orNull

mhamilton723: okay that works too if you dont mind changing it will remove some linter warnings

eisber: done

PR #526, opened 2019-04-01
pass through slot names to native structure

imatiach-msft:
resolves issue https://github.com/Azure/mmlspark/issues/332


imatiach-msft:
Note - this PR depends on corresponding LightGBM PR: https://github.com/Microsoft/LightGBM/pull/2118

PR #532, opened 2019-04-05
Add column name to UnsupportedOperationException for clarity

jwgwalton:
Debugging the valueindexer is quite difficult as it tells you the unsupported type but not what column this is referring to.

PR #537, opened 2019-04-10
Add column name to UnsupportedOperationException for clarity

jwgwalton:
Hi @imatiach-msft and @mhamilton723 , Sorry I messed up merging the changes from master and then squashing it so just opened this.


jwgwalton:
Previous pull request was #532

imatiach-msft:
@jwgwalton seem to be getting this error now:

##[error]$HOME/agent/_work/1/s/src/value-indexer/src/main/scala/ValueIndexer.scala:65:50: not enough arguments for method apply: (index: Int)Char in class StringOps.
##[error]Unspecified value parameter index.
##[error]val column = dataset.schema(getInputCol).name()
##[error]^
##[error]one error found

imatiach-msft:
@jwgwalton it looks like name is a property and not a function, so you just need to remove the parentheses

jwgwalton:
Cheers @imatiach-msft i'll get there eventually

imatiach-msft:
@jwgwalton actually, why not just use getInputCol instead?  That's already the column name as a string.

PR #560, opened 2019-05-02
add init score column for continued training + test

imatiach-msft:
resolves issue https://github.com/Azure/mmlspark/issues/382

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMDataset.scala
+++ b/src/lightgbm/src/main/scala/LightGBMDataset.scala
@@ -51,6 +51,24 @@ class LightGBMDataset(val dataset: SWIGTYPE_p_void) extends AutoCloseable {
     }
   }
 
+  def addDoubleField(field: Array[Double], fieldName: String, numRows: Int): Unit = {
+    // Generate the column and add to dataset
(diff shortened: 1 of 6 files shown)


src/lightgbm/src/main/scala/LightGBMDataset.scala
```diff
@@ -51,6 +51,24 @@ class LightGBMDataset(val dataset: SWIGTYPE_p_void) extends AutoCloseable {
...
+        lightgbmlib.doubleArray_setitem(colArray.get, ri._2, ri._1))
+      val colAsVoidPtr = lightgbmlib.double_to_voidp_ptr(colArray.get)
+      val data64bitType = lightgbmlibConstants.C_API_DTYPE_FLOAT64
+      LightGBMUtils.validate(
+        lightgbmlib.LGBM_DatasetSetField(dataset, fieldName, colAsVoidPtr, numRows, data64bitType),
+        "DatasetSetField")
+    } finally {
+      // Free column
+      colArray.foreach(lightgbmlib.delete_doubleArray(_))
+    }
```
mhamilton723: Can this be combined with other functions that loook like this?

PR #577, opened 2019-06-01
Add MMLSPARK_PYSPARK_CORES to specify CPU core count for PySpark

ghost:
This is just a POC to get early feedback.

I run mmlspark locally on my notebook and figured out that only 2 of my 6 CPU cores were used when calculating Pi with PySpark, with code as below. I couldn't find an easy out-of-the-box mechanism to tweak this behavior. Therefore, I thought it'd be nice to make this configurable through env-vars so that users can tweak this during container creating. Thus, this pull request.

Please give me feedback whether you like this feature. If you do, I'll extend the documentation accordingly.

For reference, here's the code I run

``` (python)
import random
num_samples = 100000000
def inside(p):
  x, y = random.random(), random.random()
  return x*x + y*y < 1
count = sc.parallelize(range(0, num_samples)).filter(inside).count()
pi = 4.0 * count / num_samples
print(pi)
```

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -149,6 +149,20 @@ Let's break this command and go over the meaning of each part:
   run.
 
 
+## Environment variables
+
+Here's a list of all MMLSpark specific environment variables (`-e`):
+
+* `ACCEPT_EULA`: Used to bypass the EULA check if set to either `y` or `yes`. By
+  default, the EULA is not bypassed. Example usage: `-e ACCEPT_EULA=y`.
+* `MMLSPARK_JUPYTER_PORT`: The port that the Jupyter server listens to. Defaults
+  is `MMLSPARK_JUPYTER_PORT=8888`. E.g. `MMLSPARK_JUPYTER_PORT=80`.
+* `MMLSPARK_PYSPARK_THREADS`: Thread count property for the SparkContext. Should
+  be either a number such as 2, 3 or 8 or an asterisk (`*`) for using all
+  'logical cores'. Defaults is `MMLSPARK_PYSPARK_THREADS=2`. Example:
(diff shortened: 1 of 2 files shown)


mhamilton723:
Looks good @zulli73 if you add a line in the docs ill merge!

zulli73:
I've added documentation.

I added a whole new section covering all environment variables because I felt it didn't fit into any of the existing part of the documentation. Moreover, I thought about adding it to the example `docker run` command, but I didn't want to make that example more complicated than necessary. Hence, the new section.

drdarshan (approved):
Thank you @zulli73! This looks good to me.

drdarshan:
Hello @zulli73, if you don't mind, please resolve the conflict and I'll trigger the merge. Thank you for your contribution!

zulli73:
Hi @drdarshan. 

Short story: Has this pull request become obsolete? Searching for "local[", all results use "local[*]" which indicates that the latest version at master may already use all CPU cores.

Long story: I'd happily fix merge conflicts, but I have troubles to understand the change that caused this merge conflict d34f9d1: The file I modified got removed and it's not obvious to me why it became obsolete. It seems to me that since that change, no new Docker image has been pushed - therefore I can't easily check whether Spark utilizes all available CPU cores since that commit. Finally: I couldn't find the docs for building the Docker image myself/locally

ranadeepsingh:
Closing as superseded. The old MMLSpark launcher was replaced by [#565](https://github.com/microsoft/SynapseML/pull/565) with standard Spark local-core configuration, and the project was later renamed to SynapseML in [#1204](https://github.com/microsoft/SynapseML/pull/1204). Both files changed here have been deleted.

PR #578, opened 2019-06-04
defer lightgbm probability calculation to native core to fix multiclass bug in some scenarios

imatiach-msft:
fixes issue https://github.com/Azure/mmlspark/issues/568

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMClassifier.scala
+++ b/src/lightgbm/src/main/scala/LightGBMClassifier.scala
@@ -96,42 +96,42 @@ class LightGBMClassificationModel(
   override def transform(dataset: Dataset[_]): DataFrame = {
     transformSchema(dataset.schema, logging = true)
     if (isDefined(thresholds)) {
-      require($(thresholds).length == numClasses, this.getClass.getSimpleName +
+      require(getThresholds.length == numClasses, this.getClass.getSimpleName +
         ".transform() called with non-matching numClasses and thresholds.length." +
-        s" numClasses=$numClasses, but thresholds has length ${$(thresholds).length}")
+        s" numClasses=$numClasses, but thresholds has length ${getThresholds.length}")
     }
 
     // Output selected columns only.
     var outputData = dataset
     var numColsOutput = 0
-    if ($(rawPredictionCol).nonEmpty) {
+    if (getRawPredictionCol.nonEmpty) {
       val predictRawUDF = udf { (features: Any) =>
         predictRaw(features.asInstanceOf[Vector])
       }
       outputData = outputData.withColumn(getRawPredictionCol, predictRawUDF(col(getFeaturesCol)))
       numColsOutput += 1
     }
-    if ($(probabilityCol).nonEmpty) {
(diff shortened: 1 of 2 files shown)


src/lightgbm/src/main/scala/LightGBMClassifier.scala
```diff
@@ -84,16 +85,66 @@ class LightGBMClassificationModel(
...
+    * Implementation based on ProbabilisticClassifier with slight modifications to
+    * avoid calling raw2probabilityInPlace to defer the probability calculation
+    * to lightgbm native code.
+    *
+    * @param dataset input dataset
+    * @return transformed dataset
+    */
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    transformSchema(dataset.schema, logging = true)
+    if (isDefined(thresholds)) {
```
mhamilton723: Is there any way to make this function not mutable??

src/lightgbm/src/main/scala/LightGBMClassifier.scala
```diff
@@ -84,16 +85,66 @@ class LightGBMClassificationModel(
...
+        predictRaw(features.asInstanceOf[Vector])
+      }
+      outputData = outputData.withColumn(getRawPredictionCol, predictRawUDF(col(getFeaturesCol)))
+      numColsOutput += 1
+    }
+    if ($(probabilityCol).nonEmpty) {
+      val probabilityUDF = udf { (features: Any) =>
+        predictProbability(features.asInstanceOf[Vector])
+      }
+      val probUDF = probabilityUDF(col($(featuresCol)))
```
mhamilton723: Use getFeaturesCol instead of $ accessors

src/lightgbm/src/main/scala/LightGBMClassifier.scala
```diff
@@ -84,16 +85,66 @@ class LightGBMClassificationModel(
...
+
+    if (numColsOutput == 0) {
+      this.logWarning(s"$uid: LightGBMClassificationModel.transform() was called as NOOP" +
+        " since no output columns were set.")
+    }
+    outputData.toDF
+  }
+
+  override protected def raw2probabilityInPlace(rawPrediction: Vector): Vector = {
+    throw new NotImplementedError("Unexpected error in LightGBMClassificationModel:" +
```
mhamilton723: Why does this need to be here?

src/lightgbm/src/main/scala/LightGBMClassifier.scala
imatiach-msft: this is just taken from Spark ML with some modifications to prevent calling raw to probability

src/lightgbm/src/main/scala/LightGBMClassifier.scala
imatiach-msft: it's required by the base ProbabilisticClassifier class from Spark ML

src/lightgbm/src/main/scala/LightGBMClassifier.scala
imatiach-msft: done!

PR #594, opened 2019-07-01
adding barrier execution mode to lightgbm learners

imatiach-msft:
By default LightGBM uses regular spark paradigm for launching tasks and communicates with the driver to coordinate task execution.
The driver thread aggregates all task host:port information and then communicates the full list back to the workers in order for NetworkInit to be called.
There have been some issues on certain cluster configurations because the driver needs to know how many tasks there are, and this computation is surprisingly non-trivial in spark.
With the next v0.18 release there is a new UseBarrierExecutionMode flag, which when activated uses the barrier() stage to block all tasks.
The barrier execution mode simplifies the logic to aggregate host:port information across all tasks, so the driver will no longer need to precompute the number of tasks in advance.
To use it in scala, you can call setUseBarrierExecutionMode(true), for example:

```
val lgbm = new LightGBMClassifier()
    .setLabelCol(labelColumn)
    .setObjective(binaryObjective)
    .setUseBarrierExecutionMode(true)
...
<train classifier

### DIFF
--- a/src/lightgbm/src/main/scala/LightGBMBase.scala
+++ b/src/lightgbm/src/main/scala/LightGBMBase.scala
@@ -60,7 +60,7 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
     // Reduce number of partitions to number of executor cores
     val df = dataset.select(trainingCols.map(name => col(name)):_*).toDF().coalesce(numWorkers)
     val (inetAddress, port, future) =
-      LightGBMUtils.createDriverNodesThread(numWorkers, df, log, getTimeout)
+      LightGBMUtils.createDriverNodesThread(numWorkers, df, log, getTimeout, getUseBarrierExecutionMode)
 
     /* Run a parallel job via map partitions to initialize the native library and network,
      * translate the data to the LightGBM in-memory representation and train the models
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/lightgbm/src/main/scala/LightGBMBase.scala [hunk 1]
```diff
@@ -73,17 +73,25 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
...
     val preprocessedDF = preprocessData(df)
     val schema = preprocessedDF.schema
-    val lightGBMBooster = preprocessedDF
-      .mapPartitions(TrainUtils.trainLightGBM(networkParams, getLabelCol, getFeaturesCol, get(weightCol),
-        get(initScoreCol), getOptGroupCol, validationData, log, trainParams, numCoresPerExec, schema))(encoder)
-      .reduce((booster1, _) => booster1)
+    val lightGBMBooster =
+      if (getUseBarrierExecutionMode) {
+        preprocessedDF.rdd.barrier().mapPartitions(TrainUtils.trainLightGBM(networkParams, getLabelCol,
+          getFeaturesCol, get(weightCol), get(initScoreCol), getOptGroupCol, validationData, log, trainParams,
```
mhamilton723: maybe pull this out into a val because its used in both places 

```
val mapPartitionsFunc = TrainUtils.trainLightGBM(networkParams, getLabelCol, getFeaturesCol, get(weightCol),    get(initScoreCol), getOptGroupCol, validationData, log, trainParams, numCoresPerExec, schema)
```

src/lightgbm/src/main/scala/TrainUtils.scala
```diff
@@ -363,4 +386,41 @@ private object TrainUtils extends Serializable {
     }
   }
 
+  def trainLightGBMBarrier(networkParams: NetworkParams, labelColumn: String, featuresColumn: String,
```
mhamilton723: Is this supposed to be used up by my other comment? I think you didn't use it up there

imatiach-msft: woops, this was written during an earlier refactor and I forgot to remove it, removed!

src/lightgbm/src/main/scala/LightGBMBase.scala
imatiach-msft: good idea, done!

PR #596, opened 2019-07-02
VW on Spark...

eisber:
ready for review


src/project/build.scala
```diff
@@ -38,7 +38,8 @@ object Extras {
     "com.jcraft"          % "jsch"         % "0.1.54",
     "com.jcraft"          % "jsch"         % "0.1.54",
     "org.apache.httpcomponents" % "httpclient"   % "4.5.6",
-    "com.microsoft.ml.lightgbm" %  "lightgbmlib" % "2.2.350"
+    "com.microsoft.ml.lightgbm" %  "lightgbmlib" % "2.2.350", // from "file:///home/vagrant/LightGBM/lightgbmlib.jar",
```
imatiach-msft: you can remove the comments (// from ...)

src/vw/src/main/scala/HasNumBits.scala
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.param.{BooleanParam, IntParam}
+
+trait HasNumBits extends Wrappable {
```
dciborow: How about a doc string explaining this string?

src/vw/build.sbt
```diff
@@ -0,0 +1,2 @@
+//> DependsOn: core
+//> DependsOn: featurize
```
imatiach-msft: what do you need the featurize dependency for?

src/vw/src/main/scala/VectorUDTUtil.scala
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package org.apache.spark.ml
+
+import org.apache.spark.ml.linalg.VectorUDT
```
imatiach-msft: I think you are supposed to use:

import org.apache.spark.ml.linalg.SQLDataTypes.VectorType

for non spark applications

src/vw/src/main/scala/VowpalWabbitClassifier.scala
```diff
@@ -0,0 +1,99 @@
...
+    val probability2predictionUdf = udf(probability2prediction _)
+    df2.withColumn($(predictionCol), probability2predictionUdf(col($(probabilityCol))))
+  }
+
+  override def copy(extra: ParamMap): VowpalWabbitClassificationModel =
+    new VowpalWabbitClassificationModel(uid, model, getLabelCol, getFeaturesCol, getAdditionalFeatures,
+      getPredictionCol, getProbabilityCol, getRawPredictionCol)
+
+  protected override def predictRaw(features: Row): Vector = {
+    throw new NotImplementedError("Not implement")
```
imatiach-msft: Not implemented

src/vw/src/main/scala/VowpalWabbitClassifier.scala
```diff
@@ -0,0 +1,99 @@
...
+  set(featuresCol, featuresColName)
+  set(additionalFeatures, additionalFeaturesName)
+  set(predictionCol, predictionColName)
+  set(probabilityCol, probabilityColName)
+  set(rawPredictionCol, rawPredictionColName)
+
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    val df = transformImplInternal(dataset)
+
+    // which mode one once to use depends a bit on how this should be deployed
```
imatiach-msft: which mode one *wants* to use

src/vw/src/main/scala/VowpalWabbitClassifier.scala
```diff
@@ -0,0 +1,99 @@
...
+@InternalWrapper
+class VowpalWabbitClassificationModel(
+    override val uid: String, val model: Array[Byte], labelColName: String,
+    featuresColName: String, additionalFeaturesName: Array[String],
+    predictionColName: String, probabilityColName: String, rawPredictionColName: String)
+  extends ProbabilisticClassificationModel[Row, VowpalWabbitClassificationModel]
+    with VowpalWabbitBaseModel with HasProbabilityCol // TODO: HasThresholds
+    with ConstructorWritable[VowpalWabbitClassificationModel]
+{
+  def numClasses: Int = 2
```
imatiach-msft: is multiclass supported?

src/vw/build.sbt
imatiach-msft: nevermind I see it is referenced in the VW featurizer

src/vw/src/main/scala/VectorUDTUtil.scala
eisber: thanks!

src/vw/src/main/scala/VowpalWabbitClassifier.scala
eisber: thanks!

src/vw/src/main/scala/VowpalWabbitClassifier.scala
eisber: VW does, the wrapper not yet.

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,410 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+case class NamespaceInfo (hash: Int, featureGroup: Char, colIdx: Int)
+
+object VWUtil {
+  /**
+    * Implementation of C# using(...) { } pattern.
+    */
+  def autoClose[A <: AutoCloseable,B](closeable: A)(fun: (A) => B): B = {
+    try {
```
mhamilton723: we have a similar function in StreamUtilities called `using` perhaps re-use that if you need

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,410 @@
...
+    * @param inputRows input data.
+    * @return iterator over the potentially cached data.
+    */
+  private def createDataIterator(inputRows: Iterator[Row], numPassesInSpark: Int) = {
+    if (numPassesInSpark == 1)
+    // avoid caching inputRows
+      () => inputRows
+    else {
+      // cache all row references to be able to quickly get the iterator again
+      val inputRowsArr = inputRows.toArray
```
mhamilton723: maybe make this explicit through a parameter and leverage dataframe caching instead

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,410 @@
...
+      () => inputRowsArr.iterator
+    }
+  }
+
+  private def buildCommandLineArguments(vwArgs: String, contextArgs: () => String = () => "") = {
```
mhamilton723: can you add a docstring to this? Not sure what the context args is lazy. Also take a look at lazy args in scala

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  /**
+    * Internal training loop.
+    * @param df the input data frame.
+    * @param vwArgs vw command line arguments.
+    * @param contextArgs This lambda returns command line arguments that are executed in the final execution context.
+    *                    It is used to get the partition id.
+    */
+  private def trainInternal(df: DataFrame, vwArgs: String, contextArgs: () => String = () => "") = {
```
mhamilton723: lazy args might be better here

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,410 @@
...
+    }
+  }
+
+  private def buildCommandLineArguments(vwArgs: String, contextArgs: () => String = () => "") = {
+    val args = new StringBuilder
+    args.append(vwArgs)
+      .append(" ").append(contextArgs())
+
+    // have to pass to get multi-pass to work
+    if (args.indexOf("--no_stdin") == -1)
```
mhamilton723: Consider keeping the args + arg logic in the types until the last minute then move them all to the training string

src/vw/src/main/scala/VowpalWabbitClassifier.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  protected override def raw2probabilityInPlace(rawPrediction: Vector): Vector= {
+    throw new NotImplementedError("Not implemented")
+  }
+
+  override val ttag: TypeTag[VowpalWabbitClassificationModel] =
+    typeTag[VowpalWabbitClassificationModel]
+
+  override def objectsToSave: List[Any] =
+    List(uid, model, getLabelCol, getFeaturesCol, getAdditionalFeatures, getPredictionCol,
```
mhamilton723: Consider making these parameters of the model in SparkML and using regular saving logic

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
```diff
@@ -0,0 +1,140 @@
...
+          throw new RuntimeException(s"Unsupported array element type: ${dataType}")
+
+        new StringArrayFeaturizer(idx, name, namespaceHash, getMask)
+      }
+      case m: MapType => {
+        if (m.keyType != DataTypes.StringType)
+          throw new RuntimeException(s"Unsupported map key type: ${dataType}")
+
+        m.valueType match {
+          case StringType => new MapStringFeaturizer(idx, name, namespaceHash, getMask)
```
mhamilton723: Seems like this could be cleaned up some using recursion but im not sure what the logic should be so i will defer to your expertise

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
```diff
@@ -0,0 +1,140 @@
...
+          case FloatType => new MapFeaturizer[Float](idx, name, namespaceHash, getMask, v => v.toDouble)
+          case IntegerType => new MapFeaturizer[Int](idx, name, namespaceHash, getMask, v => v.toDouble)
+          case LongType => new MapFeaturizer[Long](idx, name, namespaceHash, getMask, v => v.toDouble)
+          case ShortType => new MapFeaturizer[Short](idx, name, namespaceHash, getMask, v => v.toDouble)
+          case ByteType => new MapFeaturizer[Byte](idx, name, namespaceHash, getMask, v => v.toDouble)
+          case _ => throw new RuntimeException(s"Unsupported map value type: ${dataType}")
+        }
+      }
+      case m: Any =>
+        if (m.typeName == "vector") // unfortunately the type is private
```
mhamilton723: We have a namespace injection for this

src/vw/src/main/scala/VowpalWabbitRegressor.scala
```diff
@@ -0,0 +1,66 @@
...
+
+    new VowpalWabbitRegressorModel(uid, binaryModel, getLabelCol, getFeaturesCol, getAdditionalFeatures,
+      getPredictionCol)
+  }
+
+  override def copy(extra: ParamMap): VowpalWabbitRegressor = defaultCopy(extra)
+}
+
+@InternalWrapper
+class VowpalWabbitRegressorModel(override val uid: String, val model: Array[Byte], labelColName: String,
```
mhamilton723: Same comment as classifier

src/vw/src/main/scala/featurizer/MapFeaturizer.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+    * Featurize a single row.
+    * @param row input row.
+    * @param indices output indices.
+    * @param values output values.
+    * @note this interface isn't very Scala-esce, but it avoids lots of allocation.
+    *       Also due to SparseVector limitations we don't support 64bit indices (e.g. indices are signed 32bit ints)
+    */
+  override def featurize(row: Row, indices: ArrayBuilder[Int], values: ArrayBuilder[Double]): Unit = {
```
mhamilton723: does VW expose any of their featurizers through JNI?

src/vw/src/test/scala/VerifyVowpalWabbitRegressorFuzzing.scala
```diff
@@ -0,0 +1,60 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+import org.apache.spark.ml.util.MLReadable
+import org.apache.spark.sql.{Column, DataFrame}
+
+class VerifyVowpalWabbitRegressorFuzzing extends EstimatorFuzzing[VowpalWabbitRegressor] {
```
mhamilton723: This should be in your regular test class as opposed to another test class

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: thanks! didn't know. done.

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,410 @@
...
+  *
+  * @note parameters that regularly are swept through are exposed as proper parameters.
+  */
+@InternalWrapper
+trait VowpalWabbitBase extends Wrappable
+  with DefaultParamsWritable
+  with HasWeightCol
+  with Logging
+{
+  // can we switch to https://docs.scala-lang.org/overviews/macros/paradise.html ?
```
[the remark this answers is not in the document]
eisber: I'd like to use meta-programming to expose all 140 parameters...

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,410 @@
...
+  def getCacheRows: Boolean = $(cacheRows)
+  def setCacheRows(value: Boolean): this.type = set(cacheRows, value)
+
+  val enableCacheFile = new BooleanParam(this, "enableCacheFile", "Enable local cache file")
+  setDefault(enableCacheFile -> false)
+  def getEnableCacheFile: Boolean = $(enableCacheFile)
+  def setEnableCacheFile(value: Boolean): this.type = set(enableCacheFile, value)
+
+  // abstract methods that implementors need to provide (mixed in through Classifier,...)
+  def getLabelCol: String
```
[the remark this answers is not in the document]
eisber: hmm. it's been a while. I think it's not about them being different, they are provided through different inheritance structure ProbabilisticClassifier vs BaseRegressor and I need to use those methods in the base trait. better way of doing it or is the comment misleading/confusing?

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: I'll do that. contextArgs resolves the partitionId within the "map-partition context" see :291 () => s"--node ${TaskContext.get.partitionId}". Is there a better way of doing this? I wanted to keep all the distributed config foo together.

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
eisber: you have a pointer to an example for me?

src/vw/src/main/scala/featurizer/MapFeaturizer.scala
eisber: so by default VW operates on input string. Starting with C# I started mapping other language types into VW features. This featurization methods are programming language specific, but I'm trying to keep them the same across languages (C#, JSON,...)

src/vw/src/test/scala/VerifyVowpalWabbitRegressorFuzzing.scala
eisber: I separated it as it made testing it individually really hard. Maybe I'm just missing something in IntelliJ?

src/vw/src/main/scala/HasNumBits.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  def getNumBits: Int = $(numbits)
+  def setNumBits(value: Int): this.type = {
+    if (value < 1 || value > 30)
+      throw new IllegalArgumentException("Number of bits must be between 1 and 30 bits")
+    set(numbits, value)
+  }
+
+  /**
+    * @return the bitmask used to constraint the hash feature indices.
```
imatiach-msft: minor spelling: constrain

src/vw/src/main/scala/VectorUtils.scala
```diff
@@ -0,0 +1,62 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark
+
+object VectorUtils {
+  /**
+    * Sort of indices and assocated values. Either sums or ignores values for collisions.
```
imatiach-msft: minor spelling: associated

imatiach-msft: suggestion: Sorts indices and associated values

imatiach-msft: `Sorts indices and associated values`

src/vw/src/main/scala/VectorUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+    * @param indices indices to be sorted.
+    * @param values associated values to each index.
+    * @param sumCollisions if true values of index collisions are summed, otherwise ignored.
+    * @return
+    */
+  def sortAndDistinct(indices: Array[Int], values: Array[Double], sumCollisions: Boolean = true):
+  (Array[Int], Array[Double]) = {
+    if (indices.length == 0)
```
imatiach-msft: minor suggestion: indices.length <= 1

src/vw/src/main/scala/VowpalWabbitBase.scala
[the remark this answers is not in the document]
imatiach-msft: that would be cool, great idea

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+
+  val learningRate = new DoubleParam(this, "learningRate", "Learning rate")
+  def getLearningRate: Double = $(learningRate)
+  def setLearningRate(value: Double): this.type = set(learningRate, value)
+
+  val powerT = new DoubleParam(this, "powerT", "t power value")
+  def getPowerT: Double = $(powerT)
+  def setPowerT(value: Double): this.type = set(powerT, value)
+
+  val l1 = new DoubleParam(this, "l1", "l_1 lambda")
```
imatiach-msft: you don't need any defaults set for these?

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+  def getFeaturesCol: String
+
+  implicit class ParamStringBuilder(sb: StringBuilder) {
+    def appendParam[T](param: Param[T], option: String): StringBuilder = {
+      if (!get(param).isEmpty) {
+
+        param match {
+          case _: StringArrayParam => {
+            for (q <- get(param).get)
+              sb.append(' ').append(option).append(' ').append(q)
```
imatiach-msft: also you can use <array>.mkString(<separator>) here:
```
val prefix = ' ' + option + ' '
prefix + get(param).get.mkString(prefix)
```

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
mhamilton723: https://github.com/Azure/mmlspark/blob/master/src/main/scala/com/microsoft/ml/spark/stages/EnsembleByKey.scala

See VectorType

src/vw/src/main/scala/VowpalWabbitBase.scala
mhamilton723: https://alvinalexander.com/source-code/scala/simple-scala-call-name-example

This is what i was referring to that might help out!

src/vw/src/test/scala/VerifyVowpalWabbitRegressorFuzzing.scala
mhamilton723: Okay! This is good feedback

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+  def getHashSeed: Int = $(hashSeed)
+  def setHashSeed(value: Int): this.type = set(hashSeed, value)
+
+  val numBits = new IntParam(this, "numBits", "Number of bits used")
+  setDefault(numBits -> 18)
+
+  def getNumBits: Int = $(numBits)
+  def setNumBits(value: Int): this.type = set(numBits, value)
+
+  private def createLabelSetter(df: DataFrame) = {
```
imatiach-msft: consider putting all params in a separate params class and putting the base methods in a separate class, I think that would limit a lot of confusion.  Also, a lot of our style checks have limits on number of functions/params per class, so I've had to break even the lightgbm params into separate classes - not necessary here but just another reason to keep the params separate from all of the base methods.

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+  }
+
+  private def buildCommandLineArguments(vwArgs: String, contextArgs: () => String = () => "") = {
+    val args = new StringBuilder
+    args.append(vwArgs)
+      .append(" ").append(contextArgs())
+
+    // have to pass to get multi-pass to work
+    if (args.indexOf("--no_stdin") == -1)
+      args.append(" --no_stdin")
```
imatiach-msft: minor: move repeating strings to a constant (line 168 above has repeated constant string)

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+
+    val encoder = Encoders.kryo[Option[Array[Byte]]]
+
+    // schedule multiple mapPartitions in
+    val outerNumPasses = if (getEnableCacheFile) 1 else if (getNumPasses > 1 && !getCacheRows) getNumPasses else 1
+    var localInitialModel = if (isDefined(initialModel)) Some(getInitialModel) else None
+
+    for (p <- 0 until outerNumPasses) {
+      // dispatch to exectuors and collect the model of the first partition (everybody has the same at the end anyway)
+      localInitialModel = df.mapPartitions(inputRows => trainIteration(inputRows, localInitialModel, p))(encoder)
```
imatiach-msft: it doesn't look like you are selecting the relevant columns - I've made a similar mistake in lightgbm and recently fixed in.  Many users create a ton of temporary columns before using VectorAssembler to create the feature column and pass that with the label (and others like weight) columns to the learner.  Since you are using rows like me in lightgbm, when going over the rows in spark all of the column values are calculated.  This is bad because:
1.)  Most of the temporary columns the user created are not needed and can waste a lot of compute
2.) In my case lightgbm would run out of memory
The solution, which I added to lightgbm, was to select on the columns.  I think I should actually do that at an even earlier stage in lightgbm because the earlier irrelevant columns are dropped from training the better (performance and memory-wise).  Note this can't be done for scoring.

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+    * @param contextArgs This lambda returns command line arguments that are executed in the final execution context.
+    *                    It is used to get the partition id.
+    */
+  private def trainInternal(df: DataFrame, vwArgs: String, contextArgs: () => String = () => "") = {
+    val applyLabel = createLabelSetter(df)
+
+    val featureColIndices = generateNamespaceInfos(df.schema)
+
+    // only perform the inner-loop if we cache the inputRows
+    val numPassesInSpark = if (getEnableCacheFile) 1 else if(getCacheRows) getNumPasses else 1
```
imatiach-msft: if (!getEnableCacheFile && getCacheRows) getNumPasses else 1

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+            for (_ <- 0 until numPassesInSpark) {
+              // pass data to VW native part
+              for (row <- iteratorGenerator()) {
+                // transfer label
+                applyLabel(row, ex)
+
+                // transfer features
+                for (ns <- featureColIndices)
+                  row.get(ns.colIdx) match {
+                    case dense: DenseVector => ex.addToNamespaceDense(ns.featureGroup,
```
imatiach-msft: can VW handle mixed sparse and dense vectors?  spark has that often for different rows.

imatiach-msft: I had to convert to one type in lightgbm

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+
+          // only export the model on the first partition
+          Seq(if (TaskContext.get.partitionId == 0) Some(vw.getModel) else None).iterator
+        }.get // this will throw if there was an exception
+    }
+
+    val encoder = Encoders.kryo[Option[Array[Byte]]]
+
+    // schedule multiple mapPartitions in
+    val outerNumPasses = if (getEnableCacheFile) 1 else if (getNumPasses > 1 && !getCacheRows) getNumPasses else 1
```
imatiach-msft: if (!getEnableCacheFile && getNumPasses > 1 && !getCacheRows) getNumPasses  else 1

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+  }
+
+  /**
+    * Main training loop
+    * @param dataset input data.
+    * @return binary VW model.
+    */
+  protected def trainInternal(dataset: Dataset[_]): Array[Byte] = {
+    // follow LightGBM pattern
+    val numCoresPerExec = ClusterUtil.getNumCoresPerExecutor(dataset)
```
imatiach-msft: might be good to eventually extract common logic to a separate util

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+    val numExecutorCores = ClusterUtil.getNumExecutorCores(dataset, numCoresPerExec)
+    val numWorkers = min(numExecutorCores, dataset.rdd.getNumPartitions)
+
+    // Reduce number of partitions to number of executor cores
+    val df = if (dataset.rdd.getNumPartitions > numWorkers)
+        dataset.toDF.coalesce(numWorkers)
+      else
+        dataset.toDF
+
+    // add exposed parameters to the final command line args
```
imatiach-msft: are all of these params mandatory in VW?

src/vw/src/main/scala/VowpalWabbitBase.scala
imatiach-msft: can be done in a separate later PR though

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+  @transient
+  lazy val vw = new VowpalWabbitNative("--testonly --quiet", model)
+
+  @transient
+  lazy val example = vw.createExample()
+
+  @transient
+  lazy val vwArgs = vw.getArguments
+
+  val additionalFeatures = new StringArrayParam(this, "additionalFeatures", "Additional feature columns")
```
imatiach-msft: how does this param work?  If this is letting some additional cols pass through, I don't think it fits in with spark ML.  with predict, all original columns should be passed through with the additional ones added.

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,393 @@
...
+
+    for (ns <- featureColIndices)
+      namespaces.get(ns.colIdx) match {
+        case dense: DenseVector => example.addToNamespaceDense(ns.featureGroup,
+          ns.hash, dense.values)
+        case sparse: SparseVector => example.addToNamespaceSparse(ns.featureGroup,
+          sparse.indices, sparse.values)
+      }
+
+    // TODO: surface confidence
```
imatiach-msft: what is confidence in this case?

src/vw/src/main/scala/VowpalWabbitClassifier.scala
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Preparation for multi-class learning, though it no fun as numClasses is spread around multiple reductions
+@InternalWrapper
+class VowpalWabbitClassificationModel(
+    override val uid: String, val model: Array[Byte], labelColName: String,
+    featuresColName: String, additionalFeaturesName: Array[String],
+    predictionColName: String, probabilityColName: String, rawPredictionColName: String)
+  extends ProbabilisticClassificationModel[Row, VowpalWabbitClassificationModel]
+    with VowpalWabbitBaseModel with HasProbabilityCol // TODO: HasThresholds
```
imatiach-msft: you shouldn't need HasProbabilityCol as you get it already from ProbabilisticClassificationModel?

imatiach-msft: same thing for HasThresholds

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
```diff
@@ -0,0 +1,140 @@
...
+  setDefault(inputCols -> Array())
+
+  val seed = new IntParam(this, "seed", "Hash seed")
+  setDefault(seed -> 0)
+
+  def getSeed: Int = $(seed)
+  def setSeed(value: Int): this.type = set(seed, value)
+
+  val stringSplitInputCols = new StringArrayParam(this, "stringSplitInputCols",
+    "Input cols that should be split add word boundaries")
```
imatiach-msft: what do you mean by "add word boundaries"?

imatiach-msft: is this minor spelling mistake: add -> at?

eisber:
Your catching all those nice details… so I don’t want to replicate the default values from the C++ codebase as they might change.
Thus I’m only conditionally passing the values if provide.

Added a comment.

Markus

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
eisber: yeah... my spelling skill seems to deteriorate

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: Your catching all those nice details… so I don’t want to replicate the default values from the C++ codebase as they might change.
Thus I’m only conditionally passing the values if provide.

Added a comment.

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: confidence of the prediction.

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: when one wants to create quadratic features within VW you'd specify additionalFeatures. each feature column is treated as one namespace. Using -q 'uc' for columns 'user' and 'content' you'd get all quadratics for features in user/content (the first letter is called feature group and VW users are used to it... before somebody starts complaining ;))

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: no, I'm only adding them if supplied...

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: nicer ;)

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: VW internally treats everything sparse. Mapping from dense to sparse is straight forward in this layer.

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: nicer...

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
```diff
@@ -0,0 +1,140 @@
...
+  def getStringSplitInputCols: Array[String] = $(stringSplitInputCols)
+  def setStringSplitInputCols(value: Array[String]): this.type = set(stringSplitInputCols, value)
+
+  private def getAllInputCols = getInputCols ++ getStringSplitInputCols
+
+  private def getFeaturizer(name: String, dataType: DataType, idx: Int, namespaceHash: Int): Featurizer = {
+    val stringSplitInputCols = getStringSplitInputCols
+
+    dataType match {
+      case DoubleType => new NumericFeaturizer(idx, name, namespaceHash, getMask, r => r.getDouble(idx))
```
imatiach-msft: in the future, might be good to consider adding DateTime

imatiach-msft: ah, looks like you have it in the TODO below already

src/vw/src/main/scala/VowpalWabbitMurmurWithPrefix.scala
```diff
@@ -0,0 +1,77 @@
...
+import org.vowpalwabbit.spark.VowpalWabbitMurmur
+import java.nio.charset.StandardCharsets
+
+/**
+  * VW style murmur hash with pre-hashing of an initially specified prefix.
+  * @param prefix the prefix for each hashed value.
+  * @param maxSize maximum size of the string to be hashed.
+  */
+class VowpalWabbitMurmurWithPrefix(val prefix: String, val maxSize: Int = 2 * 1024) extends Serializable {
+  // worse case is 4 bytes per character
```
imatiach-msft: minor spelling: worst

src/vw/src/test/scala/VerifyVowpalWabbitClassifier.scala
```diff
@@ -0,0 +1,147 @@
...
+      .setNumFolds(3)
+      .setParallelism(2) // thanks to dynamic port assignment in spanning_tree.cc :)
+
+    val model = cv.fit(dataset)
+    model.transform(dataset)
+
+    val auc = model.avgMetrics(0)
+    println(s"areaUnderROC: $auc")
+
+    // TODO: what the best approach here? pass all individual parameters or the training args or both?
```
imatiach-msft: sorry not sure what this is asking, could you please clarify?  pass the training args where?

imatiach-msft: what is difference between individual params and training args?

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
eisber: Any suggestions on how to featurize the date? For predictive scenarios one might wants to featurize into weekends vs weekday, am/pm, ...

imatiach-msft:
Can you also update with latest code?  The codebase has completely changed, and some of the code you wrote will need to be updated with lightgbm changes I made recently (eg barrier execution mode).  I also can't run another build on the PR until it's updated to latest - actually I think you can now queue builds by just adding a comment here, once you've updated to latest master refactor from @mhamilton723 .

src/vw/src/main/scala/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,410 @@
...
+  def getEnableCacheFile: Boolean = $(enableCacheFile)
+  def setEnableCacheFile(value: Boolean): this.type = set(enableCacheFile, value)
+
+  // abstract methods that implementors need to provide (mixed in through Classifier,...)
+  def getLabelCol: String
+  def getFeaturesCol: String
+
+  implicit class ParamStringBuilder(sb: StringBuilder) {
+    def appendParam[T](param: Param[T], option: String): StringBuilder = {
+      if (!get(param).isEmpty) {
```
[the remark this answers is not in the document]
eisber: I tried mkString as Ilya suggested, but it made things longer...

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: fancy...

src/vw/src/main/scala/VowpalWabbitFeaturizer.scala
eisber: I implemented map of map of map ... thing for the JSON parser in VW - seems the natural thing to do from a dev perspective. 
In hindsight I think it was a mistake as it confuses people more than it helps.

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: at least the way I did it it got even more complicated (e.g. handling empty params)

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: added a TODO and will address later

src/vw/src/main/scala/VowpalWabbitRegressor.scala
eisber: refactored. Thanks Mark!!!

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: fixed.

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: thanks, Mark addressed it ;)

src/vw/src/main/scala/VowpalWabbitClassifier.scala
eisber: done

src/vw/src/test/scala/VerifyVowpalWabbitClassifier.scala
eisber: I think this was related to some intermediate code. not relevant anymore, thus removed the code.

src/vw/src/main/scala/VowpalWabbitBase.scala
eisber: since all the params are in the base class we decided to keep it there until the style checks complain.

src/autogen.sbt
```diff
@@ -0,0 +1,415 @@
+// Automatically generated, DO NOT EDIT
+
+val topDir = file(".")
```
imatiach-msft: I don't believe this file is needed, please remove

src/core/spark/src/main/scala/ClusterUtil.scala
```diff
@@ -0,0 +1,86 @@
...
+
+object ClusterUtil {
+
+  /** Get number of cores from dummy dataset for 1 executor.
+    * Note: all executors have same number of cores,
+    * and this is more reliable than getting value from conf.
+    * @param dataset The dataset containing the current spark session.
+    * @return The number of cores per executor.
+    */
+  def getNumCoresPerExecutor(dataset: Dataset[_]): Int = {
```
imatiach-msft: this method had changed, please update to latest code in master:

  /** Get number of cores from dummy dataset for 1 executor.
    * Note: all executors have same number of cores,
    * and this is more reliable than getting value from conf.
    * @param dataset The dataset containing the current spark session.
    * @return The number of cores per executor.
    */
  def getNumCoresPerExecutor(dataset: Dataset[_], log: Logger): Int = {
    val spark = dataset.sparkSession
    val confTaskCpus =
      try {
        spark.sparkContext.getConf.get("spark.task.cpus", "1").toInt
      } catch {
        case _: NoSuchElementException => 1
      }
    try {
      val confCores = spark.sparkContext.getConf
        .get("spark.executor.cores").toInt
      val coresPerExec = confCores / confTaskCpus
      log.info(s"LightGBM calculated num cores per executor as $coresPerExec from $confCores " +
        s"cores and $confTaskCpus task CPUs")
      coresPerExec
    } catch {
      case _: NoSuchElementException =>
        // If spark.executor.cores is not defined, get the cores per JVM
        import spark.implicits._
        val numMachineCores = spark.range(0, 1)
          .map(_ => java.lang.Runtime.getRuntime.availableProcessors).collect.head
        val coresPerExec = numMachineCores / confTaskCpus
        log.info(s"LightGBM calculated num cores per executor as $coresPerExec from " +
          s"$numMachineCores machine cores from JVM and $confTaskCpus task CPUs")
        coresPerExec
    }
  }

src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
```diff
@@ -0,0 +1,96 @@
...
+import org.slf4j.Logger
+
+object ClusterUtil {
+  /** Get number of cores from dummy dataset for 1 executor.
+    * Note: all executors have same number of cores,
+    * and this is more reliable than getting value from conf.
+    * @param dataset The dataset containing the current spark session.
+    * @return The number of cores per executor.
+    */
+  def getNumCoresPerExecutor(dataset: Dataset[_], log: Logger): Int = {
```
imatiach-msft: it looks like you are redefining this method twice, I think one of the files is a bad copy

src/vw/build.sbt
```diff
@@ -0,0 +1,2 @@
+//> DependsOn: core
```
imatiach-msft: I think after the refactor this file can be removed?

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,421 @@
...
+  def getFeaturesCol: String
+
+  implicit class ParamStringBuilder(sb: StringBuilder) {
+    def appendParam[T](param: Param[T], option: String): StringBuilder = {
+      if (!get(param).isEmpty) {
+
+        param match {
+          case _: StringArrayParam => {
+            for (q <- get(param).get)
+              sb.append(' ').append(option).append(' ').append(q)
```
imatiach-msft: simplify as:
```
sb.append(s" $option $q")
```

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,421 @@
...
+          }
+          case _ => sb.append(' ').append(option).append(' ').append(get(param).get)
```
imatiach-msft: similarly:
```
sb.append(s" $option ${get(param).get}")
```

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,421 @@
...
+      .toArray
+
+  /**
+    * If multiple passes are requested and cache file is not enabled, cache the data in-memory.
+    * @param inputRows input data.
+    * @return iterator over the potentially cached data.
+    */
+  private def createDataIterator(inputRows: Iterator[Row], numPassesInSpark: Int) = {
+    if (numPassesInSpark == 1)
+    // avoid caching inputRows
```
imatiach-msft: minor: spacing seems off for comment

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,421 @@
...
+      () => inputRows
+    else {
+      // cache all row references to be able to quickly get the iterator again
+      // TODO: switch dataframe.cache()
```
imatiach-msft: can you elaborate on the TODO?

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,421 @@
...
+    */
+  private def trainInternal(df: DataFrame, vwArgs: String, contextArgs: => String = "") = {
+    val applyLabel = createLabelSetter(df)
+
+    val featureColIndices = generateNamespaceInfos(df.schema)
+
+    // only perform the inner-loop if we cache the inputRows
+    val numPassesInSpark = if (!getEnableCacheFile && getCacheRows) getNumPasses else 1
+
+    def trainIteration(inputRows: Iterator[Row],
```
imatiach-msft: have you tested with empty partitions?

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
```diff
@@ -0,0 +1,421 @@
...
+
+              vw.endPass
+            }
+
+            if (getEnableCacheFile)
+              vw.performRemainingPasses
+          }
+
+          // only export the model on the first partition
+          Seq(if (TaskContext.get.partitionId == 0) Some(vw.getModel) else None).iterator
```
imatiach-msft: will this work with empty partitions - and also if local initial model is empty:
if (localInitialModel.isEmpty) new VowpalWabbitNative(args)

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
imatiach-msft: in lightgbm I had to fix a bug recently by ignoring empty partitions and returning from them early on network init - it can happen often apparently with bad repartitioning from user

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitFeaturizer.scala
```diff
@@ -0,0 +1,141 @@
...
+
+    val mode = udf((r: Row) => {
+      val indices = ArrayBuilder.make[Int]
+      val values = ArrayBuilder.make[Double]
+
+      // educated guess on size
+      indices.sizeHint(featurizers.length)
+      values.sizeHint(featurizers.length)
+
+        // apply all featurizers
```
imatiach-msft: (nit) comment spacing off

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitFeaturizer.scala
```diff
@@ -0,0 +1,141 @@
...
+    })
+
+    dataset.toDF.withColumn(getOutputCol, mode.apply(struct(fieldSubset.map(f => col(f.name)): _*)))
+  }
+
+  override def copy(extra: ParamMap): VowpalWabbitFeaturizer = defaultCopy(extra)
+
+  override def transformSchema(schema: StructType): StructType = {
+    val fieldNames = schema.fields.map(_.name)
+    for (f <- getAllInputCols)
```
imatiach-msft: getAllInputCols.foreach( { f => if (!fieldNames.contains(f)) throw new IllegalArgumentException("missing input column " + f) } )

src/main/scala/com/microsoft/ml/spark/vw/featurizer/StringSplitFeaturizer.scala
```diff
@@ -0,0 +1,54 @@
...
+  extends Featurizer(fieldIdx) {
+
+  /**
+    * (?U) makes \w unicode aware
+    * https://stackoverflow.com/questions/4304928/unicode-equivalents-for-w-and-b-in-java-regular-expressions
+    * we could follow
+    * https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html
+    * but that strips single character words...
+    */
+  val nonWhiteSpaces = "(?U)\\w+".r
```
imatiach-msft: in the future might be nice to allow the user to specify this

imatiach-msft: eg make it a user param

src/test/resources/benchmarks/benchmarks_VerifyVowpalWabbitRegressor.csv
```diff
@@ -0,0 +1,81 @@
+name,value,precision,higherIsBetter
+VowpalWabbitRegressor_energyefficiency2012_data.train.csv__cacheFiletrue_cacheRowstrue,8.137509233528203,1.0,false
+VowpalWabbitRegressor_energyefficiency2012_data.train.csv__cacheFiletrue_cacheRowsfalse,8.137509233528203,1.0,false
+VowpalWabbitRegressor_energyefficiency2012_data.train.csv__cacheFilefalse_cacheRowstrue,7.778246082309875,1.0,false
+VowpalWabbitRegressor_energyefficiency2012_data.train.csv__cacheFilefalse_cacheRowsfalse,7.779053506047849,1.0,false
+VowpalWabbitRegressor_energyefficiency2012_data.train.csv_--bfgs_cacheFiletrue_cacheRowstrue,24.48041415526995,1.0,false
```
imatiach-msft: strange, why do some of these have --

src/autogen.sbt
eisber: done. thanks

src/core/spark/src/main/scala/ClusterUtil.scala
eisber: you actually found more duplicate code...

src/vw/build.sbt
eisber: done

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
eisber: added this comment:

      // TODO: based on Mark's comment use Spark dataframe caching instead of materializing
      // row reference into memory

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
eisber: can you point me to the unit test and code?

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitFeaturizer.scala
eisber: maybe it's just me, but I don't feel this improves readability. if you have a strong opinion on this as it's not to scala style enough

src/main/scala/com/microsoft/ml/spark/vw/featurizer/StringSplitFeaturizer.scala
eisber: added TODO

PR #610, opened 2019-07-08
Add r package upload

mhamilton723:


build.sbt
```diff
@@ -126,6 +138,15 @@ publishDocs := {
   uploadToBlob(unifiedDocDir.toString, version.value, "docs", s.log)
 }
 
+val publishR = TaskKey[Unit]("publishR", "publish R package to blob")
+publishR := {
+  val s = streams.value
+  (run in IntegrationTest2).toTask("").value
+  val rPackage = join("target", "scala-2.11", "generated", "package", "R")
+    .listFiles().head
+  singleUploadToBlob(rPackage.toString,rPackage.getName, "rrr", s.log)
```
imatiach-msft: I think the rrr path doesn't work with DataBricks.  They have something else as I recall (I think it was web).

PR #651, opened 2019-08-19
[WIP] remove coalesce on barrier execution

imatiach-msft:


scalastyle-test-config.xml
```diff
@@ -45,7 +45,7 @@ package (?:com\.microsoft\.ml\.spark|org\.apache\.spark|com\.microsoft\.CNTK|com
   </check>
   <check level="warning" class="org.scalastyle.scalariform.MethodNamesChecker" enabled="true"><parameters>
     <parameter name="regex">^[a-z][A-Za-z0-9]*(_=)?$</parameter></parameters></check>
-  <check level="warning" class="org.scalastyle.scalariform.ClassTypeParameterChecker" enabled="true"><parameters>
+  <check level="warning" class="org.scalastyle.scalariform.ClassTypeParameterChecker" enabled="false"><parameters>
```
drdarshan: Any reason to be disabling this globally? Can we do it more locally scoped to the file that is generating this warning?

imatiach-msft: it seems like a bad rule to me in general - why wouldn't you want more descriptive types?  Why should the types be just a single letter?  It's not actually related to my PR, most classes generate this warning in the codebase.  I just made the change because I saw this warning so often that it was hiding other more serious warnings in my intellij and was very distracting in general.  Sure, I could try and disable it on a per-file basis, but to me it seems like most files that use types go against this rule anyway.  Spark ML in general also seems to go against this rule.

imatiach-msft:
@drdarshan replied to comment

PR #669, opened 2019-08-26
fix: Better error message when the group column is not a Int/Long

chris-smith-zocdoc:
Fixes https://github.com/Azure/mmlspark/issues/664


imatiach-msft:
@mhamilton723 can you take a look at the e2e cluster error?  it looks like a configuration issue

PR #672, opened 2019-08-27
feat: add metric parameter to lightgbm learners

imatiach-msft:
resolves https://github.com/Azure/mmlspark/issues/631 and https://github.com/Azure/mmlspark/issues/628

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
@@ -50,15 +50,13 @@ class LightGBMClassifier(override val uid: String)
      * so we infer the actual numClasses from the dataset here
      */
     val actualNumClasses = getNumClasses(dataset)
-    val metric =
-      if (getObjective == LightGBMConstants.BinaryObjective) "binary_logloss,auc"
-      else LightGBMConstants.MulticlassObjective
     val modelStr = if (getModelString == null || getModelString.isEmpty) None else get(modelString)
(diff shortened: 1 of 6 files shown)


candalfigomoro:
@imatiach-msft 
Does it support multiple evaluation metrics (just like LightGBM)?

imatiach-msft:
@candalfigomoro yes it should, with comma separated metrics.  For early stopping we also use all metrics specified (note early stopping is not provided in native api so R/python and our spark implementation all have their own implementation):
https://github.com/Azure/mmlspark/blob/master/src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala#L211

nishant-ai:
@imatiach-msft 
Does it support just ndcg or ndcg@N as well ?

imatiach-msft:
@nishant-ai I think it should support all lightgbm metrics as specified here:
https://github.com/microsoft/LightGBM/blob/master/docs/Parameters.rst#metric
but for NDCG@N I believe we need to expose the parameter eval_at based on a quick look through the lightgbm docs:
https://github.com/microsoft/LightGBM/blob/master/docs/Parameters.rst#eval_at
specifically for early stopping, the scala version of lightgbm does use the correct comparison function here:
https://github.com/Azure/mmlspark/blob/master/src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala#L215

nishant-ai:
Yes, I meant the eval_at parameter only. I had a look at the code and the eval_at parameter used for @N metrics didn't seem to be exposed, just wanted to confirm that.
Can you please make the changes needed so that NDCG@N can be used as well ? Thanks.

nishant-ai:
Thanks @imatiach-msft !

PR #685, opened 2019-09-11
fix: categorical handling on lightgbm learners

imatiach-msft:
fix categorical handling on lightgbm learners

It seems that the categorical_feature parameter needs to be set on the dataset instead of the booster

resolves https://github.com/Azure/mmlspark/issues/665 and https://github.com/Azure/mmlspark/issues/660 

also opened a documentation defect here:
https://github.com/microsoft/LightGBM/issues/2398

### DIFF
--- a/src/test/scala/com/microsoft/ml/spark/lightgbm/split1/VerifyLightGBMClassifier.scala
+++ b/src/test/scala/com/microsoft/ml/spark/lightgbm/split1/VerifyLightGBMClassifier.scala
@@ -122,7 +122,7 @@ class VerifyLightGBMClassifier extends Benchmarks with EstimatorFuzzing[LightGBM
   lazy val pimaDF: DataFrame = loadBinary("PimaIndian.csv", "Diabetes mellitus").cache()
   lazy val taskDF: DataFrame = loadBinary("task.train.csv", "TaskFailed10").cache()
   lazy val breastTissueDF: DataFrame = loadMulticlass("BreastTissue.csv", "Class").cache()
-  lazy val bankTrainDF: DataFrame = {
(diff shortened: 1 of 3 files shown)


src/test/scala/com/microsoft/ml/spark/lightgbm/split1/VerifyLightGBMClassifier.scala
```diff
@@ -133,7 +133,14 @@ class VerifyLightGBMClassifier extends Benchmarks with EstimatorFuzzing[LightGBM
       .fit(df)
       .transform(df).drop(categoricalColumns: _*)
       .withColumnRenamed("c_y", labelCol)
-    LightGBMUtils.getFeaturizer(df2, labelCol, featuresCol).transform(df2)
+    df2
+  }.cache()
+  lazy val indexedBankTrainDF: DataFrame = {
+    LightGBMUtils.getFeaturizer(unfeaturizedBankTrainDF, labelCol, featuresCol,
```
eisber: any reason you created a static method (LightGBMUtils.getFeaturizer) instead of exposing a class? Is this still composable as pipeline?

imatiach-msft: that is just a helper method that sets up the parameters and then calls the Featurize estimator:
https://github.com/Azure/mmlspark/blob/05a2bef54fa88a2293020215cf4cae34f2d212c5/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMUtils.scala#L51
so I could just use the class but this seems like better code re-use

imatiach-msft:
seems to be a random failure with publishing the docker image... rerunning build

PR #690, opened 2019-09-13
fix: cast errors for label, weight and init score columns

### LINKED ISSUE #682: LightGBMRanker on spark group/query parameter explain
Is there any documentation on how to use LightGBMRanker on spark ? specifically, how to set the query data as non-spark version of  LightGBMRanker  used a very special rule:that is [12,10,....]
means first 12 items belongt to first group,and the next 10 belong to second ..., so ,how dose spark  version set the value? I can see there is a groupCol parameter ,so what is this column's type?(can it be string id which represents a group,or must be bigint ?),and what is the rule of this column?(is it as just the non-spark version, in the above example,the first group's values are all 12,and the next group's values are 10 ,etc. or it can just be the group id ?).lastly,how many class can labelCol's value be? Is there also a max number of class limit just like usual lgb?
Hope to get response as soon as possible ,because I am in a matter of greate urgency ! thanks very much.

imatiach-msft:
created a new PR based on review for https://github.com/Azure/mmlspark/pull/688

fixes #682 based on user feedback

User may receive error that can be difficult to track down if they pass non-double column for label, weight or init score (int or long), so we just cast it to double for them if it is not already a double type column for convenience

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
@@ -5,13 +5,15 @@ package com.microsoft.ml.spark.lightgbm
 
 import com.microsoft.ml.spark.core.utils.ClusterUtil
 import org.apache.spark.ml.{Estimator, Model}
-import org.apache.spark.sql.{DataFrame, Dataset, Encoders}
+import org.apache.spark.sql.{Column, DataFrame, Dataset, Encoders}
 
 import scala.concurrent.Await
 import scala.concurrent.duration.{Duration, SECONDS}
 import scala.math.min
 import org.apache.spark.ml.param.shared.{HasFeaturesCol => HasFeaturesColSpark, HasLabelCol => HasLabelColSpark}
 import org.apache.spark.sql.functions._
+import org.apache.spark.sql.types.{BooleanType, DataType, DoubleType, IntegerType, NumericType, StringType}
+import org.apache.spark.ml.linalg.SQLDataTypes.VectorType
 
 import scala.collection.mutable.ListBuffer
 import scala.language.existentials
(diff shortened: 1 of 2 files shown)


src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
```diff
@@ -77,19 +107,19 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
     }
   }
 
-  protected def getTrainingCols(): ListBuffer[String] = {
-    val trainingCols = ListBuffer(getLabelCol, getFeaturesCol)
+  protected def getTrainingCols(): ListBuffer[(String, DataType)] = {
+    val trainingCols = ListBuffer[(String, DataType)]((getLabelCol, DoubleType), (getFeaturesCol, VectorType))
```
mhamilton723: No mutability, Use a list of options and flatten

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
```diff
@@ -39,10 +41,38 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
...
+        else dataset(selectedColName)
+      }: _*
+    ).toDF()
+  }
+
+  protected def prepareDataframe(dataset: Dataset[_], trainingCols: ListBuffer[(String, DataType)],
                                  numWorkers: Int): DataFrame = {
+    // Select only relevant columns
+    val selectedCols = trainingCols.map { case (name, _) => col(name) }
+    val df = castColumns(dataset, trainingCols, selectedCols)
```
mhamilton723: if theres no diff between training and selected cols, move the selected cols computation inside the function and simplify please

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
```diff
@@ -39,10 +41,38 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
     }
   }
 
-  protected def prepareDataframe(dataset: Dataset[_], trainingCols: ListBuffer[String],
+  protected def castColumns(dataset: Dataset[_], trainingCols: ListBuffer[(String, DataType)],
+                            selectedCols: ListBuffer[Column]): DataFrame = {
+    val schema = dataset.schema
+    // Figure out which columns need to be cast
+    val castCols = trainingCols.flatMap {
```
mhamilton723: no need for two loops here as they are the same collections

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
imatiach-msft: done!

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
imatiach-msft: ah good catch, done!

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
imatiach-msft: great suggestion, done!

PR #693, opened 2019-09-17
fix: lightgbm binary classifier getting stuck due to single label in some partitions

imatiach-msft:
fix lightgbm binary classifier getting stuck due to single label in some partitions

By default, lightgbm will now fail with exception message instead of timing out with connection refused error when some workers only have a single class and exit early during training.


imatiach-msft:
closing as this has been fixed with a different PR

yinshurman:
Dose the version 0.18.1 contains the fix?

imatiach-msft:
nope, 0.18.1 does not, the 1.0 version does however

PR #696, opened 2019-09-23
feat: Add VowpalWabbit ngram support

eisber:
Vowpal Wabbit supports on-the-fly ngram generation, thus the featurized dataset does not need to contain the ngrams and will not blow up in memory. Minor complication in all of this is that the features are already hashed in the SparseVector and unfortunately **always get sorted** by the Spark SparseVector implementation.

To overcome this the sparse indices get prefixed with a counter at the highest level bits. The number of bits for the counter can be configured, but together with the hashing bits need to fit within 30 bits (yeah, signed integers). The prefix counter is stripped by VW JNI layer as the bit mask is always re-applied.


build.sbt
```diff
@@ -306,6 +306,11 @@ val settings = Seq(
     name, version, scalaVersion, sbtVersion,
     baseDirectory, datasetDir),
   parallelExecution in Test := false,
+  test in assembly := {},
```
mhamilton723: If you rebase and squash this should be included in previous changes

src/test/scala/com/microsoft/ml/spark/vw/VerifyVowpalWabbitClassifier.scala
```diff
@@ -209,6 +209,32 @@ class VerifyVowpalWabbitClassifier extends Benchmarks with EstimatorFuzzing[Vowp
...
+
+    val datasetFeaturized = featurizer.transform(dataset)
+
+    val vw1 = new VowpalWabbitClassifier()
+      .setArgs("--ngram f2 -a")
+    val classifier1 = vw1.fit(datasetFeaturized)
+
+    // 3 (words) + 2 (ngrams) + 1 (constant) = 6
+    // 4 (words) + 3 (ngrams) + 1 (constant) = 8
+    assert (classifier1.getPerformanceStatistics.select("totalNumberOfFeatures").head.get(0) == 14)
```
mhamilton723: no space after assert

build.sbt
eisber: rebased

PR #706, opened 2019-10-09
fix: validation data and ranker preprocessing

imatiach-msft:
fixes issue https://github.com/Azure/mmlspark/issues/703

specifically, filters the training data if validation index column is present, and adds preprocessing to validation data for ranking

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
@@ -133,11 +133,13 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
     val trainParams = getTrainParams(numWorkers, categoricalIndexes, dataset)
     log.info(s"LightGBM parameters: ${trainParams.toString()}")
     val networkParams = NetworkParams(getDefaultListenPort, inetAddress, port, getUseBarrierExecutionMode)
-    val validationData =
+    val (trainingData, validationData) =
       if (get(validationIndicatorCol).isDefined && dataset.columns.contains(getValidationIndicatorCol))
-        Some(sc.broadcast(df.filter(x => x.getBoolean(x.fieldIndex(getValidationIndicatorCol))).collect()))
-      else None
-    val preprocessedDF = preprocessData(df)
+        (df.filter(x => !x.getBoolean(x.fieldIndex(getValidationIndicatorCol))),
+          Some(sc.broadcast(preprocessData(df.filter(x =>
+            x.getBoolean(x.fieldIndex(getValidationIndicatorCol)))).collect())))
+      else (df, None)
+    val preprocessedDF = preprocessData(trainingData)
     val schema = preprocessedDF.schema
     val mapPartitionsFunc = TrainUtils.trainLightGBM(networkParams, getLabelCol, getFeaturesCol,
       get(weightCol), get(initScoreCol), getOptGroupCol, validationData, log, trainParams, numCoresPerExec, schema)(_)


src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
```diff
@@ -133,11 +133,13 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
...
     log.info(s"LightGBM parameters: ${trainParams.toString()}")
     val networkParams = NetworkParams(getDefaultListenPort, inetAddress, port, getUseBarrierExecutionMode)
-    val validationData =
+    val (trainingData, validationData) =
       if (get(validationIndicatorCol).isDefined && dataset.columns.contains(getValidationIndicatorCol))
-        Some(sc.broadcast(df.filter(x => x.getBoolean(x.fieldIndex(getValidationIndicatorCol))).collect()))
-      else None
-    val preprocessedDF = preprocessData(df)
+        (df.filter(x => !x.getBoolean(x.fieldIndex(getValidationIndicatorCol))),
+          Some(sc.broadcast(preprocessData(df.filter(x =>
```
chris-smith-zocdoc: I'm not sure this is 100% correct regarding the `preprocessData` call because we use [sortWithinPartitions](https://github.com/Azure/mmlspark/blob/master/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMRanker.scala#L75)

Since we `collect()` after calling `sortWithinPartitions` I think we just coalesce the partitions together when we really need to sort after coalescing them. 

```
Locally sorted -> Collect
Partition A: 1, 3, 5
Partition B: 2, 4, 6

On Driver: 1, 3, 5, 2, 4, 6
```

When what we need is
```
On Driver: 1, 2, 3, 4, 5, 6
```

imatiach-msft: @chris-smith-zocdoc thank you for the great comment.
collect should preserve order:

https://stackoverflow.com/questions/40005739/pyspark-parallelize-and-collect-order-preserving

I think the example above is a bit misleading, it should be more something like below, with both a user and some data for training on each row:

Partition A:
user3<some data>
user3<some data>
user3<some data>
user1<some data>
user2 <some data>
user1<some data>
user2<some data>

Partition B:
user4<some data>
user5<some data>
user4<some data>
user5<some data>
user5 <some data>
user6<some data>
user5<some data>

note the users on partitions *should* be different for both training and validation data but we don't have a way to enforce that right now, I would like to add a util for this eventually.
the users will then be sorted so the same user is in the same block:

Partition A:
user1<some data>
user1<some data>
user2 <some data>
user2<some data>
user3<some data>
user3<some data>
user3<some data>

Partition B:
user4<some data>
user4<some data>
user5<some data>
user5<some data>
user5 <some data>
user5<some data>
user6<some data>

note the user1,2,3,4,5,6 could be replaced with ints like 6,5,1,2,3,4 and it would be fine, the value itself doesn't matter.  The only important thing is that a user appears in a block before the next user, and users don't appear on different partitions, so I think it should be fine.  You are right though that currently we rely on the same user not appearing on multiple partitions which this could break in, but that would be a problem for the training data as well.  It's not a major blocker though, it wouldn't result in a failure, just the metrics (NDCG) would be worse after training.

chris-smith-zocdoc: I thought we need the data to be in increasing order for the group_id? Is that not the case? 

Currently there is no guarantee that the partitions contain continuous ranges of group_ids, it's just a random selection that we sort per partition. 

```
df = df.repartition(2, 'search_id').sortWithinPartitions('search_id')
```

You can see this in my example by displaying the dataframe. I have monotonically increasing group_ids but the first partition that is displayed is not monotonically increasing (missing 0,1,3, 6-9)


image


Once you go halfway through the dataframe and hit the second partition we see the transition from 98 -> 0 

image

imatiach-msft: @chris-smith-zocdoc 
"we need the data to be in increasing order for the group_id"
nope, in fact in lightgbm we don't even pass the id's themselves, we just pass the count of a specific group - eg 60 items for user1, 5 items for user2, 17 items for user3, etc -- so that list is much shorter than the number of rows in the dataset.  We take the group id column and sort it so that we can generate that list when creating the native lightgbm dataset representation in-memory.
"I have monotonically increasing group_ids but the first partition that is displayed is not monotonically increasing"
as I understand the group ids are just to group entities together, they don't need to be monotonically increasing.  I think of them as the equivalent of a guid.  However, a group should not be split across partitions, because that could lower the performance (eg NDCG) of the model.

imatiach-msft: the images above look fine to me, the most important thing is that none of the ids exist across two or more partitions

chris-smith-zocdoc: Thanks for the clarification. 

Is there any reason to force group_id to be numeric then? We have guids for our actual group_ids and it would be convenient if we could just use them directly instead of adding an additional integer version. I can open a separate issue for this if you'd like to have the conversation there instead.

imatiach-msft: a separate issue would be great.  lightgbm requires integers, but we could do a transformation in the mmlspark wrapper to turn any unique values to integers using (stringindexer in spark ML / or valueindexer in mmlspark).

PR #755, opened 2019-12-06
feat: Add KNN and ConditionalKNN Estimators

mhamilton723:


src/main/scala/com/microsoft/ml/spark/nn/BallTree.scala
```diff
@@ -0,0 +1,229 @@
...
+
+import breeze.linalg.functions.euclideanDistance
+import breeze.linalg.{DenseVector, norm}
+import breeze.linalg._
+
+private case class Query(point: DenseVector[Double],
+                         normOfQueryPoint: Double,
+                         bestMatches: BoundedPriorityQueue[BestMatch]) extends Serializable {
+
+  def this(point: DenseVector[Double], bestK: Int) = this(point, norm(point), Query.createQueue(bestK))
```
imatiach-msft: do you plan to support sparse vector as well?

imatiach-msft:
how does this compare to https://github.com/saurfang/spark-knn and https://github.com/JMailloH/kNN_IS ?

PR #758, opened 2019-12-12
fix: refactor python wrappers to use common class

imatiach-msft:
add set methods to lightgbm models in pyspark APIs

resolves issue https://github.com/Azure/mmlspark/issues/721

Update: the original issue was fixed with new codegen, this PR now updates the python wrappers to use a common class

### DIFF
--- a/src/main/python/mmlspark/lightgbm/mixin.py
+++ b/src/main/python/mmlspark/lightgbm/mixin.py
@@ -1,0 +1,66 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in project root for information.
+
+from pyspark.ml.linalg import SparseVector, DenseVector
+from pyspark.ml.common import inherit_doc
+from mmlspark.core.serialize.java_params_patch import *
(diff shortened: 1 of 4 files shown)


imatiach-msft:
tagging @chris-smith-zocdoc

src/main/python/mmlspark/lightgbm/mixin.py
```diff
@@ -0,0 +1,27 @@
...
+        """
+        return list(self._java_obj.getFeatureImportances(importance_type))
+
+    def setPredictionCol(self, predictionColName):
+        """
+        Set the prediction column name
+        """
+        self._java_obj.setPredictionCol(predictionColName)
+
+    def setFeaturesCol(self, featuresColName):
```
chris-smith-zocdoc: Can you add `setLabelCol` also?

imatiach-msft: note: the new codegen already added these methods, but I refactored the PR to cleanup the python code structure

mhamilton723 (approved):

PR #760, opened 2019-12-12
fix: added min_data_in_leaf parameter

imatiach-msft:
fix: added min_data_in_leaf parameter
resolves issue https://github.com/Azure/mmlspark/issues/746


louis925:
Hi, when is this commit going to be available as a release? Thanks!

PR #778, opened 2020-01-21
fix: LightGBMRanker should repartition by grouping column

JoanFM:
fix: Repartition by groupingColumn to guarantee proper LightGBMRanker results 
Consider repartitioning by grouping column since for now it is only guaranteed to have groups together in every partition but for correctness every group should be found together in the same partition.

A comment in https://github.com/Azure/mmlspark/issues/682 seems to agree on this statement


src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMRanker.scala
```diff
@@ -78,6 +79,20 @@ class LightGBMRanker(override val uid: String)
   }
 
   override def copy(extra: ParamMap): LightGBMRanker = defaultCopy(extra)
+
+  override def prepareDataframe(dataset: Dataset[_], trainingCols: Array[(String, Seq[DataType])],
+                                numWorkers: Int): DataFrame = {
+    val repartitionedDataset = getOptGroupCol match {
+      case None => dataset
+      case Some(groupingCol) => {
+        val df = dataset.repartition(new Column(groupingCol)).cache()
```
imatiach-msft: the changes are technically correct but repartition is very expensive in spark.  Also we are already doing a coalesce here which we should avoid if we are already repartitioning:

https://github.com/Azure/mmlspark/blob/master/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala#L101

the repartition also has to result in as many partitions as workers for lightgbm to work but that is taken care of by calling coalesce after this.  Maybe for now, we can have a setting on the ranker in order to allow the user to turn off the extra repartitioning done here?  That way, in case the user already has the data partitioned correctly prior to calling lightgbm ranker, execution time won't suffer.  We can leave this as the default behavior though.  In the future it would be nice to figure out how to combine this with the other coalesce/repartition (the latter done in barrier execution mode) in order to avoid doing the expensive operation twice.

JoanFM: Maybe we could repartition to the numWorkers? it may be more expensive than just the coalesce done afterwards but I think it would simplify the interface not having an extra flag. Plus, the coalesce will not affect afterwards, since data will be already partitioned to the exact number of partitions desired.

If this solution is not convincing for you I will add the extra setting.

JoanFM: Actually I have been thinking, shouldn't Spark DAG planner avoid extra repartitioning if data is already partitioned with the same grouping column policy?

JoanFM:
Can I get some help with the tests? It is expected for this PR to cause diffs but with the class structure in the tests I am having real trouble seeing what I need to check and update, can someone give me a hand on this?

Thank you very much

imatiach-msft:
@JoanFM it looks like some lightgbm ranker tests are failing, not sure why yet.  Will take a look when I get a chance.

JoanFM:
Yes it is expected I think, I observed that model result is depending on partitioning of the data, but I cannot track down where the expected output is taken from or anything. The complete fuzzying class structure is a little too complex if u are not familiar with it I guess

JoanFM:
Yes it is expected I think, I observed that model result is depending on partitioning of the data, but I cannot track down where the expected output is taken from or anything. The complete fuzzying class structure is a little too complex if u are not familiar with it I guess

JoanFM:
> @JoanFM it looks like some lightgbm ranker tests are failing, not sure why yet. Will take a look when I get a chance.

Could you point me in some direction to help me fix it on my own if you do not have the time?

imatiach-msft:
@JoanFM yes, will take a look sorry.  Not sure where to point you to otherwise, as I'm not sure why it is failing myself.  I was just going to get your branch and run the tests locally to see if I could reproduce the test failures.

JoanFM:
@imatiach-msft I could reproduce the issue locally, I think it is natural to have differences because now every lightGBM instance see a different set of samples. 
But I do not know how to change or how to see the expected output vs the obtained, But I guess you will see it much easier

JoanFM:
Hello @imatiach-msft ,

I have seen the source of discrepancy. 

At some point we test that when Ranker is saved, which I understand that is just the configuration, and the same loaded result in the same prediction.

But since repartition is non-deterministic I think, we cannot guarantee that we will get the same LightGBM model with the same trees and gains. Therefore, there are some discrepancies.

I think this test may be meaningless in this case. Let me know if it is okey this last change

imatiach-msft:
Maybe we can add a setting to turn off the repartitioning (and by default we would do the new repartitioning), but in the test case we can set the repartitioning to be off?

JoanFM:
Updated with configurable parameter

JoanFM:
Any idea what is failing? is there some test that is failing due to this change? Is there a way to see the cause of the failure?

imatiach-msft:
@JoanFM yes it looks like one of the notebook tests timed out:
Array(Failure(java.util.concurrent.TimeoutException: Notebook /MMLSparkBuild/build_1.0.0-rc1-61-6de883c7-SNAPSHOT/Classification - Adult Census with Vowpal Wabbit.ipynb timed out after 2400000 ms, job in state RUNNING,  For more information check the run page:  https://eastus2.azuredatabricks.net/?o=4652563148747856#job/5249/run/1  )) was not empty

I see this error when I click "details" on the build.  I'm guessing you don't have access because the devops account is still not public.  I mentioned to @mhamilton723 that the devops account needs to be made public so external developers can view it, but he hasn't had the cycles to work on this.

imatiach-msft:
it looks like datasets failed to download which caused all of the lightgbm tests to fail.  Weird.  Hoping it's a random issue, I think it might be related to the change @eisber recently did to update the datasets zip file.

imatiach-msft:
hmm it looks like that was some random networking failure with downloading the files @eisber , the test seems to be passing now.  Maybe we need to add some retry logic there in the future if download/unzip fails.

liyzcj:
@JoanFM  Why force materialization is needed here ?

JoanFM:
Hey @liyzcj ,

I am a little disconnected from this project, but I guess it may had to do with the lazyness of the operation?

Not 100% sure at this moment, may not even be needed.

liyzcj:
@JoanFM Thanks for reply, I may test for this later.

PR #780, opened 2020-01-23
feat: Expose parameter bin_construct_sample_cnt in spark for LightGBM

JoanFM:
feat: Expose parameter bin_construct_sample_cnt in mmlspark.

This parameter would be nice to be exposed to be able to consider a larger amount of samples per partition when training LightGBM models


imatiach-msft:
@JoanFM sorry I did a squash and merge instead of direct merge.  It said I couldn't merge directly due to conflicts (you would have needed to rebase against latest master), but I could do a squash merge directly.  Otherwise it would delay the merge because we would need to coordinate the rebase and re-run all of the tests again.

PR #784, opened 2020-01-27
Gramhagen/livy config

gramhagen:
Adding section in readme to install mmlspark for Apache Livy


gramhagen:
hmm, not sure how the accumulo stuff got in here... it should only be the readme change

gramhagen:
moving to #785

PR #792, opened 2020-01-31
fix: Guarantee one boosterPtr is allocated and freed per LightGBMBooster instance

JoanFM:
I am not sure this is a fix, and not sure about the potential performance implications.

I think these changes could be helpful in solving memory issues that seem related to concurrency in spark such as https://github.com/microsoft/LightGBM/issues/2392.

Please feel free to comment, is just a proposal but I am not 100% sure that it fixes these issues.

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
@@ -8,7 +8,11 @@ import com.microsoft.ml.spark.lightgbm.LightGBMUtils.getBoosterPtrFromModelStrin
 import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector}
 import org.apache.spark.sql.{SaveMode, SparkSession}
 
-protected class BoosterHandler(model: String) extends Serializable {
+/** Wraps the boosterPtr and guarantees that Native library is initialized
+ * everytime it is needed
+ * @param model The string serialized representation of the learner
+ */
+protected class BoosterHandler(model: String) {
 
   LightGBMUtils.initializeNativeLibrary()
 
@@ -39,9 +43,10 @@ protected class BoosterHandler(model: String) extends Serializable {
     dataLongLengthPtr
   }
 
-  lazy val numClasses = getNumClasses()
-
+  lazy val numClasses = getNumClasses
+  lazy val numFeatures = getNumFeatures
   lazy val numTotalModel = getNumTotalModel
+  lazy val numTotalModelPerIteration = getNumModelPerIteration
 
   lazy val rawScoreConstant = lightgbmlibConstants.C_API_PREDICT_RAW_SCORE
   lazy val normalScoreConstant = lightgbmlibConstants.C_API_PREDICT_RAW_SCORE
@@ -50,7 +55,7 @@ protected class BoosterHandler(model: String) extends Serializable {
   lazy val dataInt32bitType = lightgbmlibConstants.C_API_DTYPE_INT32
   lazy val data64bitType = lightgbmlibConstants.C_API_DTYPE_FLOAT64
 
-  def getNumClasses(): Int = {
+  private def getNumClasses: Int = {
     val numClassesOut = lightgbmlib.new_intp()
     LightGBMUtils.validate(
       lightgbmlib.LGBM_BoosterGetNumClasses(boosterPtr, numClassesOut),
@@ -58,7 +63,7 @@ protected class BoosterHandler(model: String) extends Serializable {
     lightgbmlib.intp_value(numClassesOut)
   }
 
-  def getNumModelPerIteration: Int = {
+  private def getNumModelPerIteration: Int = {
     val numModelPerIterationOut = lightgbmlib.new_intp()
     LightGBMUtils.validate(
       lightgbmlib.LGBM_BoosterNumModelPerIteration(boosterPtr, numModelPerIterationOut),
@@ -66,12 +71,20 @@ protected class BoosterHandler(model: String) extends Serializable {
     lightgbmlib.intp_value(numModelPerIterationOut)
   }
 
-  def getNumTotalModel: Int = {
+  private def getNumTotalModel: Int = {
     val numModelOut = lightgbmlib.new_intp()
     LightGBMUtils.validate(
       lightgbmlib.LGBM_BoosterNumberOfTotalModel(boosterPtr, numModelOut),
       "Booster total models")
     lightgbmlib.intp_value(numModelOut)
+  }
+
+  private def getNumFeatures: Int = {
+    val numFeaturesOut = lightgbmlib.new_intp()
+    LightGBMUtils.validate(
+      lightgbmlib.LGBM_BoosterGetNumFeature(boosterPtr, numFeaturesOut),
+      "Booster NumFeature")
+   lightgbmlib.intp_value(numFeaturesOut)
   }
 
   override def finalize(): Unit = {
@@ -94,7 +107,6 @@ protected class BoosterHandler(model: String) extends Serializable {
   */
 @SerialVersionUID(777L)
 class LightGBMBooster(val model: String) extends Serializable {
-
   /** Transient variable containing local machine's pointer to native booster
     */
   @transient
@@ -119,11 +131,13 @@ class LightGBMBooster(val model: String) extends Serializable {
     }
   }
 
-  lazy val numClasses: Int = boosterHandler.getNumClasses()
+  lazy val numClasses: Int = boosterHandler.numClasses
 
-  lazy val numTotalModel: Int = boosterHandler.getNumTotalModel
+  lazy val numFeatures: Int = boosterHandler.numFeatures
 
-  lazy val numModelPerIteration: Int = boosterHandler.getNumModelPerIteration
+  lazy val numTotalModel: Int = boosterHandler.numTotalModel
+
+  lazy val numModelPerIteration: Int = boosterHandler.numTotalModelPerIteration
 
   lazy val numIterations: Int = numTotalModel / numModelPerIteration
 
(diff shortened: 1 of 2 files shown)


imatiach-msft:
interesting, all of the regular tests are passing but the pyspark notebooks are all failing with errors like:

org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 956.0 failed 4 times, most recent failure: Lost task 0.3 in stage 956.0 (TID 12257, 10.139.64.7, executor 1): java.lang.NoClassDefFoundError: Could not initialize class com.microsoft.ml.lightgbm.lightgbmlibConstants
---------------------------------------------------------------------------
Py4JJavaError                             Traceback (most recent call last)
<command-2277356414511268> in <module>()
     12 lg_predictions = lg_model.transform(lr_test_data)
     13 
---> 14 display(lg_predictions.limit(10).toPandas())

/databricks/spark/python/pyspark/sql/dataframe.py in toPandas(self)
   2189 
   2190         # Below is toPandas without Arrow optimization.
-> 2191         pdf = pd.DataFrame.from_records(self.collect(), columns=self.columns)
   2192 
   2193         dtype = {}

/databricks/spark/python/pyspark/sql/dataframe.py in collect(self)
    546         # Default path used in OSS Spark / for non-DF-ACL clusters:
    547         with SCCallSiteSync(self._sc) as css:
--> 548             sock_info = self._jdf.collectToPython()
    549         return list(_load_from_socket(sock_info, BatchedSerializer(PickleSerializer())))

imatiach-msft:
seems to be some pickle issue

imatiach-msft:
oooooh I see the issue.  We are doing the init only on getting booster:
```
  lazy val boosterPtr: SWIGTYPE_p_void = {
    LightGBMUtils.initializeNativeLibrary()
    getBoosterPtrFromModelString(model)
  }
```
however it seems the init also needs to be done before we call anything that uses lightgbmlibConstants

JoanFM:
> oooooh I see the issue. We are doing the init only on getting booster:
> 
> ```
>   lazy val boosterPtr: SWIGTYPE_p_void = {
>     LightGBMUtils.initializeNativeLibrary()
>     getBoosterPtrFromModelString(model)
>   }
> ```
> 
> however it seems the init also needs to be done before we call anything that uses lightgbmlibConstants

Where was it done before then?

imatiach-msft:
we were just doing it at top of methods before using lightgbmlibConstants:
```
def score(features: Vector, raw: Boolean, classification: Boolean): Array[Double] = {
    // Reload booster on each node
    if (boosterPtr == null) {
      LightGBMUtils.initializeNativeLibrary()
      boosterPtr = getBoosterPtrFromModelString(model)
    }
    val kind =
      if (raw) lightgbmlibConstants.C_API_PREDICT_RAW_SCORE
      else lightgbmlibConstants.C_API_PREDICT_NORMAL
```

Note where we use **lightgbmlibConstants.C_API_PREDICT_RAW_SCORE, lightgbmlibConstants.C_API_PREDICT_NORMAL**

imatiach-msft:
now we are potentially accessing lightgbmlibConstants after the booster was loaded

JoanFM:
I see, I can try to find a way around this. But first I would like to be sure that you guys think this PR is a good contribution and if it could be the root of some memory issues me and some others have been experiencing.

JoanFM:
Did some changes to try to guarantee that native library is loaded any time is needed

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
```diff
@@ -8,38 +8,45 @@ import com.microsoft.ml.spark.lightgbm.LightGBMUtils.getBoosterPtrFromModelStrin
 import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector}
 import org.apache.spark.sql.{SaveMode, SparkSession}
 
+class BoosterHandler(model: String) extends Serializable {
+
+  LightGBMUtils.initializeNativeLibrary()
+
+  lazy val boosterPtr: SWIGTYPE_p_void = {
```
chris-smith-zocdoc: I think you still want these fields to have `@transient` so they are excluded from spark task serialization

JoanFM: Could be, but isn't it a little redundant if I declare the instance of this class to be transient?

JoanFM:
See benchmarking failure in tests for LightGBMRegressor and Classifier, It seems that the model trained differs, but I do not see how it can affect since LigthGBMBooster is created with a given trained model.

JoanFM:
In the last commit I removed this finalize method because according to scalastyle and Java recommendation they should be avoided, but how is it that in current master there is a finalize for LightGBMBooster?.

`  override def finalize(): Unit = {
    if (scoredDataLengthLongPtr != null)
      lightgbmlib.delete_int64_tp(scoredDataLengthLongPtr)
    if (scoredDataOutPtr != null)
      lightgbmlib.delete_doubleArray(scoredDataOutPtr)
    if (leafIndexDataLengthLongPtr != null)
      lightgbmlib.delete_int64_tp(leafIndexDataLengthLongPtr)
    if (leafIndexDataOutPtr != null)
      lightgbmlib.delete_doubleArray(leafIndexDataOutPtr)
    if (boosterPtr != null) {
      LightGBMUtils.validate(lightgbmlib.LGBM_BoosterFree(boosterPtr), "Finalize Booster")
    }
  }`

Some issues come to my mind, please correct me if some of these reflections are wrong or inaccurate, I am not very experienced with the JVM:

- In relation to BoosterHandler, I think it should be us responsible from deleting it since LightGBM explicitly releases the unique_ptr when returning it in the c_api in `LGBM_BoosterCreate`. Also I do not think the JVM can have information about the undelrying type since it just sees a pointer to void.

- When it comes to the dataPtr I also think we may need to make sure they are deleted. But I am not sure, maybe JVM has the capability to see all the instances. Also if mmlspark code is responsible fot this memory freeing, why are some other allocations not being taken care of? For instance in: `saveBoosterToString`.

- I think that if we apply this finalize protecting for a double free should work I think, but how could I pass the scalastyle issue?

- Is there any other way to make sure JVM will clean it all?

Thank you very much for your help

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
chris-smith-zocdoc: Initially I was confused because the class was still serializable, but I see you've fixed that now. 👍

chris-smith-zocdoc:
> In the last commit I removed this finalize method because according to scalastyle and Java recommendation they should be avoided

As a general practice yes, but they're perfect for working with native memory. I'd agree with this answer from [stackoverflow](https://stackoverflow.com/a/44367054) 

"You can use the finalizer to cleanup as a last resort, but, i would not use solely to manage the actual lifetime for the reasons you actually pointed out in your own question. As such, let the finalizer be the final attempt, but not the first."

Oracle also has a doc on the [subject](https://www.oracle.com/technical-resources/articles/javase/finalization.html)

> but how is it that in current master there is a finalize for LightGBMBooster?.
I'm not sure about how scalastyle checks are setup in this repo, but you should be able to disable the check by adding a comment to disable the check on the finalize method http://www.scalastyle.org/configuration.html


> In relation to BoosterHandler, I think it should be us responsible from deleting it since LightGBM explicitly releases the unique_ptr when returning it in the c_api in LGBM_BoosterCreate. 

Yes, it is our job to call the methods to free any memory we have allocated that LightGBM has given us pointers for.

> Also I do not think the JVM can have information about the undelrying type since it just sees a pointer to void.

The JVM is completely unaware of the unmanaged memory that LightGBM allocates. The JVM only tracks memory that it allocates itself (java objects)


> why are some other allocations not being taken care of?

I don't know this for sure, but it might have to do with the design of the spark apis. If we need to return objects to spark for it to serialize between tasks, we wouldn't be able to free it beforehand. This is also why finalizers are useful. 


> Are JNI swigCPtr deleted by JVM? Isn't there a swigCMemOwn parameter to be set that can parametrize who is responsible of deletion of these swigCPtrs?

I'm not very familiar with swig, Ilya might have a better answer for you

> I think that if we apply this finalize protecting for a double free should work I think, but how could I pass the scalastyle issue?

I think general best practice is to implement a `dispose()` and `finalize()` method, where finalize just calls dispose. This allows both the end user, and the GC to cleanup native memory.  After the memory is freed, the pointers should be set to null so we cannot attempt to free the memory again.

For scalastyle, check the link for examples on how to disable

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
JoanFM: Yes, thanks, I realised it was dangerous, now if somebody wants to use it is forced to use transient, and that is why I thought it was a little redundant. I will remove now the inner transients.

JoanFM:
Thanks you very much @chris-smith-zocdoc and @imatiach-msft 
 
I added back the finalize method calling a private freeNativeMemory. I did it private for now, because at this point I do not know if there is need for external calls to this method.

At the same time, I did not cover in this scope, I consider a little bit incoherent to ensure it is not null at delete time, but not when we try to use these pointers, although maybe I am unaware of some checks done under the hood.

I still get the same errors during test that find difficult to understand.

chris-smith-zocdoc:
Which error are you seeing? Can you post the stacktrace?

JoanFM:
I mean, test errors, but I do not see why

JoanFM:
> Which error are you seeing? Can you post the stacktrace?

I was seeing errors in the tests, but now I found the origin, wrongly ported a lightCBMLibConstant

imatiach-msft:
ping @mhamilton723 the cognitive tests still seem to be failing, even on other PRs

JoanFM:
Hello @imatiach-msft, any news about the investigation of these issues? Thanks

PR #799, opened 2020-02-10
fix: Possible multithreading issue when two scores may come in parallel they may not safely fill pointer values

JoanFM:
fix: Possible multithreading issue when two scores may come in parallel they may not safely fill pointer values

Two threads can fill up scoredDataOutPtr in parallel and mix up their values

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
@@ -8,7 +8,11 @@ import com.microsoft.ml.spark.lightgbm.LightGBMUtils.getBoosterPtrFromModelStrin
 import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector}
 import org.apache.spark.sql.{SaveMode, SparkSession}
 
-protected class BoosterHandler(model: String) extends Serializable {
+/** Wraps the boosterPtr and guarantees that Native library is initialized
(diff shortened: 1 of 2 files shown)


imatiach-msft:
@JoanFM it looks like this needs to be merged with latest?  I think I also see some code that has been merged already in the other PR.

JoanFM:
> @JoanFM it looks like this needs to be merged with latest? I think I also see some code that has been merged already in the other PR.

Yes, I am merging it

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
```diff
@@ -9,97 +9,143 @@ import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector}
...
+  lazy val data64bitType: Int = lightgbmlibConstants.C_API_DTYPE_FLOAT64
 
   private def getNumClasses: Int = {
     val numClassesOut = lightgbmlib.new_intp()
     LightGBMUtils.validate(
       lightgbmlib.LGBM_BoosterGetNumClasses(boosterPtr, numClassesOut),
       "Booster NumClasses")
-    lightgbmlib.intp_value(numClassesOut)
+    val out = lightgbmlib.intp_value(numClassesOut)
+    lightgbmlib.delete_intp(numClassesOut)
```
imatiach-msft: nice catch to delete the pointer here!

PR #821, opened 2020-03-10
added support for readable model, invert hash and re-using a previous…

### LINKED ISSUE #816: Vowpal wabbit in readable format
While saving the vowpal wabbit model using .save or .write, it saves the objects in multiple folders like this.
1.metadata
a.success flag
b.data file
2.complexParams
a.performanceStatistics
b.model

Could you please confirm if 2b.model is the vowpal wabbit model? Is there anyway to convert in to readable model (like the functionality in the original vowpal wabbit package) ? While reading that using vowpal wabbit in local it throws error saying bad format.

eisber:
…ly trained VW Spark model

close #816 

Please note I'm not sure how helpful invert hash is as the original input isn't passed along. The estimator only sees the Sparse/Dense Spark ML vector.


src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitBase.scala
```diff
@@ -163,6 +166,20 @@ trait VowpalWabbitBase extends Wrappable
   def getNumBits: Int = $(numBits)
   def setNumBits(value: Int): this.type = set(numBits, value)
 
+  val exportReadableModel = new BooleanParam(this, "exportReadableModel",
+    "If true exports the readable model")
+
+  def getExportReadableModel: Boolean = $(exportReadableModel)
```
mhamilton723: Is there a way to make this a method that performs the action as opposed to a flag that decides whether to perform the action?

eisber: before training it must be known that the readable model needs to be accessed afterwards (triggers additional computation)

PR #847, opened 2020-03-31
feat: add delegate to monitor training

ocworld:
It is needed to monitor and hook in each training iteration.

It's because how well training my model and how much better metrics such as auc in each boost round.

contacts: keunhyun.oh@ahnlab.com

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMParams.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMParams.scala
@@ -325,4 +325,11 @@ trait LightGBMParams extends Wrappable with DefaultParamsWritable with HasWeight
 
   def getMinDataInLeaf: Int = $(minDataInLeaf)
   def setMinDataInLeaf(value: Int): this.type = set(minDataInLeaf, value)
+
+  var delegate: Option[LightGBMDelegate] = None
+  def getDelegate: Option[LightGBMDelegate] = delegate
+  def setDelegate(delegate: LightGBMDelegate): this.type = {
+    this.delegate = Option(delegate)
+    this
+  }
 }
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
@@ -203,7 +203,7 @@ private object TrainUtils extends Serializable {
     val partitionId = TaskContext.getPartitionId
     while (!isFinished && iters < trainParams.numIterations) {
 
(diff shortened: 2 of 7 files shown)


ocworld:
@AhnLab-OSS

src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
```diff
@@ -247,9 +262,22 @@ private object TrainUtils extends Serializable {
...
         }
+
         lightgbmlib.delete_doubleArray(evalResults)
+
+        Option(Map(results:_*))
+      } else {
+        None
       }
+
+      if(delegate.isDefined) {
```
imatiach-msft: nit: space after if

src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
```diff
@@ -199,7 +199,14 @@ private object TrainUtils extends Serializable {
     val bestScore = new Array[Double](evalCounts)
     val bestScores = new Array[Array[Double]](evalCounts)
     val bestIter = new Array[Int](evalCounts)
+    val delegate = trainParams.delegate
+    val partitionId = TaskContext.getPartitionId
     while (!isFinished && iters < trainParams.numIterations) {
+
+      if(delegate.isDefined) {
```
imatiach-msft: nit: space after if

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMParams.scala
```diff
@@ -325,4 +325,11 @@ trait LightGBMParams extends Wrappable with DefaultParamsWritable with HasWeight
 
   def getMinDataInLeaf: Int = $(minDataInLeaf)
   def setMinDataInLeaf(value: Int): this.type = set(minDataInLeaf, value)
+
+  var delegate: Option[LightGBMDelegate] = None
```
imatiach-msft: we should probably figure out how to make this inherit from param and expose in python layer, but maybe not for this PR

src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
ocworld: it is fixed

src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
ocworld: it is fixed :)

PR #852, opened 2020-04-10
feat: add delegate functions to monitor and debug batch training and generating datasets

ocworld:
1) add functions for before/after batch training
2) add batchIndex params in delegate functions
3) add functions for before/after generating datasets. It is important to monitor and debug memory usages. I've spend much time to analyze memory and time in generating datasets. 
4) add columnParams to reduce complexity


ocworld:
@AhnLab-OSS

PR #855, opened 2020-04-20
fix: spark.executor.cores' default value based on master when counting workers

ocworld:
This pr is fixing spark.executor.cores' default value based on master when counting workers.

The default value of spark.executor.cores is different as what master is.

It is 1 in YARN mode. In other case, It is all the available cores on the worker in standalone and Mesos coarse-grained modes.
(https://spark.apache.org/docs/latest/configuration.html)

Currently, I've tried to use mmlspark based on spark on k8s. When I did not set spark.executor.cores, error was occured. It is because not matched expected num of mmlspark workers with actual value.

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
+++ b/src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
@@ -19,19 +19,7 @@ object ClusterUtil {
     */
   def getNumCoresPerExecutor(dataset: Dataset[_], log: Logger): Int = {
     val spark = dataset.sparkSession
-    val confTaskCpus =
-      try {
-        val taskCpusConfig = spark.sparkContext.getConf.getOption("spark.task.cpus")
-        if (taskCpusConfig.isEmpty) {
-          log.info("ClusterUtils did not detect spark.task.cpus config set, using default 1 instead")
-        }
-        taskCpusConfig.getOrElse("1").toInt
-      } catch {
-        case _: NoSuchElementException => {
-          log.info("spark.task.cpus config not set, using default 1 instead")
-          1
-        }
-      }
+    val confTaskCpus = getTaskCpus(dataset, log)
     try {
       val confCores = spark.sparkContext.getConf
         .get("spark.executor.cores").toInt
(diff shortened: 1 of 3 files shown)


imatiach-msft:
will need to test this out a bit since there have been a lot of issues in the past around this specific code section

imatiach-msft:
the build was all green but it's not updating in github, weird, looks like some github bug

ocworld:
@imatiach-msft I wrote new unittest and pushed. However, It is hard to tested mmlspark's unittest in my local environment . Please check it :)

ocworld:
@imatiach-msft Can I see why the build failed?

imatiach-msft:
@ocworld the style checker is failing on:

[error] /home/vsts/work/1/s/src/test/scala/com/microsoft/ml/spark/core/utils/VerifyClusterUtil.scala: Header does not match expected text

It looks like you need to add this copyright message at the top of the file to pass the style checker:
```
// Copyright (C) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License. See LICENSE in project root for information.
```

src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
```diff
@@ -19,37 +19,91 @@ object ClusterUtil {
...
+        val coresPerExec = defaultNumCores / confTaskCpus
         log.info(s"ClusterUtils calculated num cores per executor as $coresPerExec from " +
-          s"$numMachineCores machine cores from JVM and $confTaskCpus task CPUs")
+          s"default num cores($defaultNumCores) from master and $confTaskCpus task CPUs")
         coresPerExec
     }
   }
 
+  /** Get number of default cores from sparkSession(required) or master(optional) for 1 executor.
+    * @param spark The current spark session. If not set master param, the master in the session is used.
```
imatiach-msft, line 41: nit: If not set master param, the master in the session is used. ->
If master parameter is not set, the master in the spark session is used.

imatiach-msft:
@ocworld I tried this PR out on a cluster (azure databricks) with Higgs dataset and it didn't change the training time for me (compared to previous runs), so it seems safe to me at least in that scenario.  I think they use spark standalone mode for databricks clusters.

src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
ocworld: It is fixed

ocworld:
> @ocworld the style checker is failing on:
> 
> [error] /home/vsts/work/1/s/src/test/scala/com/microsoft/ml/spark/core/utils/VerifyClusterUtil.scala: Header does not match expected text
> 
> It looks like you need to add this copyright message at the top of the file VerifyClusterUtil.scala to pass the style checker:
> 
> ```
> // Copyright (C) Microsoft Corporation. All rights reserved.
> // Licensed under the MIT License. See LICENSE in project root for information.
> ```

Copyright is added in the file

ocworld:
> @ocworld I tried this PR out on a cluster (azure databricks) with Higgs dataset and it didn't change the training time for me (compared to previous runs), so it seems safe to me at least in that scenario. I think they use spark standalone mode for databricks clusters.

Thanks for your test

imatiach-msft:
it looks like @mhamilton723 is required to review as a code owner (can't seem to merge), can you please review this PR @mhamilton723 ?

ocworld:
@ahnlab-oss

PR #863, opened 2020-05-19
feat: add featuresShapCol to LightGBMClassifierModel

ocworld:
In this pr, featuresShapCol is added to LightGBMClassifierModel.

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
@@ -231,7 +231,8 @@ object LightGBMClassificationModel extends ConstructorReadable[LightGBMClassific
                               featuresColName: String = "features", predictionColName: String = "prediction",
                               probColName: String = "probability",
                               rawPredictionColName: String = "rawPrediction",
-                              leafPredictionColName: String = "leafPrediction"): LightGBMClassificationModel = {
+                              leafPredictionColName: String = "leafPrediction",
+                              featuresShapColName: String = "featuresShap"): LightGBMClassificationModel = {
     val uid = Identifiable.randomUID("LightGBMClassifier")
     val session = SparkSession.builder().getOrCreate()
     val textRdd = session.read.text(filename)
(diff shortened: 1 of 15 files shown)


ocworld:
@AhnLab-OSS

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
```diff
@@ -155,6 +160,11 @@ class LightGBMClassificationModel(
       outputData = outputData.withColumn(getLeafPredictionCol,  predLeafUDF(col(getFeaturesCol)))
       numColsOutput += 1
     }
+    if (getFeaturesShapCol.nonEmpty) {
+      val featureShapUDF = udf(featuresShap _)
+      outputData = outputData.withColumn(getLeafPredictionCol,  featureShapUDF(col(getFeaturesCol)))
```
imatiach-msft, line 165: should this be: getLeafPredictionCol -> getFeaturesShapCol (?)

ekerazha:
So we have SHAP for LightGBM Ranker and Classifier? Is it now missing only for Regressor?

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
```diff
@@ -86,16 +86,27 @@ trait HasLeafPredictionCol extends Params {
   def setLeafPredictionCol(value: String): this.type = set(leafPredictionCol, value)
 }
 
+trait HasFeaturesShapCol extends Params {
+  val featuresShapCol = new Param[String](this, "featuresShapCol",
+    "features' shap value column name that means their contribution for prediction")
```
imatiach-msft, line 91: Maybe change the description to:
```
Output SHAP vector column name after prediction containing the feature contribution values
```

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
ocworld: @imatiach-msft You're right. It is fixed now.

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
ocworld: @imatiach-msft It is changed to your suggestion

ocworld:
@ekerazha I assume that except regression SHAP is supported.
Please note that it has different usages for getting SHAP between LightGBM Ranker and Classifier in this pr.

imatiach-msft:
hmm seems the test is failing in the native code:

```
[info] + Test Verify LightGBM Classifier features shap took 0.583s 
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f4ea7779ef2, pid=7254, tid=0x00007f4ea4263700
#
# JRE version: OpenJDK Runtime Environment () (8.0_252-b14) (build 1.8.0_252-b14)
# Java VM: OpenJDK 64-Bit Server VM (25.252-b14 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libc.so.6+0x7fef2]
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /home/vsts/work/1/s/hs_err_pid7254.log
[thread 139974389262080 also had an error]
#
# If you would like to submit a bug report, please visit:
#   http://www.azulsystems.com/support/
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
timeout: the monitored command dumped core
/home/vsts/work/_temp/azureclitaskscript1589903456723.sh: line 6:  7253 Aborted
```

imatiach-msft:
running another build again, will try to run this branch locally also

imatiach-msft:
@ocworld great news, I was able to run your code locally and I discovered the issue - SHAP returns the expected value in addition to the feature contributions, which led to memory corruption.  Although I was not able to get the SHAP test to fail, I did see another test fail due to memory corruption from the SHAP test, and once I fixed the issue I validated that the memory corruption error is no longer occurring.  I pushed the fix directly to your branch, hope you don't mind.

imatiach-msft:
missed updating another test in ranker which was checking the number of features, fixed

imatiach-msft:
found another minor issue where the new column name was missing in the objects to save which caused a serialization failure, fixed, hopefully build should be green now

imatiach-msft:
once PR is merged I'll send a corresponding PR for
1.) pyspark API
2.) lightgbm regressor
to make it consistent across codebase

imatiach-msft:
@ocworld I noticed this PR won't work for multiclass case, since multiclass shap values are of shape (# features + 1 expected_value) * # classes, but we currently aren't making the array the size of # classes.  I can fix this too when I send out the regression PR.

ocworld:
@imatiach-msft Thanks for your notice

PR #864, opened 2020-05-27
feat: unify APIs across LightGBM learner types and add SHAP feature importances to regressor

imatiach-msft:
unify APIs across LightGBM learner types and add SHAP feature importances to regressor

- Unify the APIs across regressor, ranker and classifier for predicting leaf index, getting global split/gini importances and getting local SHAP feature importances
- Add local SHAP feature importances to regressor
- Update pyspark API

Update:
- made models use ComplexParamsWritable based on PR comments

### DIFF
--- a/src/main/python/mmlspark/lightgbm/LightGBMClassifier.py
+++ b/src/main/python/mmlspark/lightgbm/LightGBMClassifier.py
@@ -32,31 +32,23 @@ class LightGBMClassificationModel(_LightGBMClassificationModel):
         self._java_obj.saveNativeModel(filename, overwrite)
 
     @staticmethod
-    def loadNativeModelFromFile(filename, labelColName="label",
-                                featuresColName="features", predictionColName="prediction",
-                                probColName="probability", rawPredictionColName="rawPrediction"):
+    def loadNativeModelFromFile(filename):
         """
         Load the model from a native LightGBM text file.
         """
         ctx = SparkContext._active_spark_context
         loader = ctx._jvm.com.microsoft.ml.spark.lightgbm.LightGBMClassificationModel
-        java_model = loader.loadNativeModelFromFile(filename, labelColName,
-                                                    featuresColName, predictionColName,
-                                                    probColName, rawPredictionColName)
+        java_model = loader.loadNativeModelFromFile(filename)
         return JavaParams._from_java(java_model)
 
     @staticmethod
-    def loadNativeModelFromString(model, labelColName="label",
-                                  featuresColName="features", predictionColName="prediction",
-                                  probColName="probability", rawPredictionColName="rawPrediction"):
+    def loadNativeModelFromString(model):
         """
         Load the model from a native LightGBM model string.
         """
         ctx = SparkContext._active_spark_context
         loader = ctx._jvm.com.microsoft.ml.spark.lightgbm.LightGBMClassificationModel
-        java_model = loader.loadNativeModelFromString(model, labelColName,
-                                                      featuresColName, predictionColName,
-                                                      probColName, rawPredictionColName)
+        java_model = loader.loadNativeModelFromString(model)
         return JavaParams._from_java(java_model)
 
     def getFeatureImportances(self, importance_type="split"):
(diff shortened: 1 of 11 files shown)


imatiach-msft:
tagging @mhamilton723 @ocworld @JoanFM for the common API changes.
Note since users seem to like both having the method to call on the model for an instance and the column name I kept all options for all learner types.

src/main/python/mmlspark/lightgbm/LightGBMClassifier.py
```diff
@@ -34,33 +34,45 @@ def saveNativeModel(self, filename, overwrite=True):
...
         """
         Get the feature importances as a list.  The importance_type can be "split" or "gain".
         """
         return list(self._java_obj.getFeatureImportances(importance_type))
+
+    def getFeatureShaps(self, vector):
+        """
+        Get the local shap feature importances.
+        """
+        return list(self._java_obj.getFeatureShaps(vector))
```
imatiach-msft, line 78: noticed this doesn't actually quite work as the scala API needs to take in an array instead of a vector, will fix it in a subsequent PR with another issue for multiclass SHAP values

src/main/python/mmlspark/lightgbm/LightGBMClassifier.py
```diff
@@ -34,33 +34,45 @@ def saveNativeModel(self, filename, overwrite=True):
     @staticmethod
     def loadNativeModelFromFile(filename, labelColName="label",
                                 featuresColName="features", predictionColName="prediction",
```
mhamilton723, line 36: can we remove the col clutter on these? is there some way to set the cols in a unified way afterward?

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMRanker.scala
```diff
@@ -61,7 +62,8 @@ class LightGBMRanker(override val uid: String)
   }
 
   def getModel(trainParams: TrainParams, lightGBMBooster: LightGBMBooster): LightGBMRankerModel = {
-    new LightGBMRankerModel(uid, lightGBMBooster, getLabelCol, getFeaturesCol, getPredictionCol)
+    new LightGBMRankerModel(uid, lightGBMBooster, getLabelCol, getFeaturesCol, getPredictionCol,
```
mhamilton723, line 65: Can you make it so that you set the params with setLabelCol as opposed to using a big constructor? I think the constructors are causing a lot of clutter

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMRanker.scala
```diff
@@ -147,20 +165,26 @@ class LightGBMRankerModel(override val uid: String, override val model: LightGBM
 object LightGBMRankerModel extends ConstructorReadable[LightGBMRankerModel] {
   def loadNativeModelFromFile(filename: String, labelColName: String = "label",
                               featuresColName: String = "features",
-                              predictionColName: String = "prediction"): LightGBMRankerModel = {
```
mhamilton723, line 150: likewise alot of clutter here

src/main/python/mmlspark/lightgbm/LightGBMClassifier.py
imatiach-msft: done

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMRanker.scala
imatiach-msft: done, changed models to use ComplexParamsWritable based on discussion

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMRanker.scala
imatiach-msft: done, changed models to use ComplexParamsWritable based on discussion

src/main/python/mmlspark/lightgbm/LightGBMClassifier.py
imatiach-msft: fixed it in this PR

src/main/python/mmlspark/lightgbm/LightGBMRanker.py
```diff
@@ -32,31 +32,39 @@ def saveNativeModel(self, filename, overwrite=True):
...
         """
         Get the feature importances as a list.  The importance_type can be "split" or "gain".
         """
         return list(self._java_obj.getFeatureImportances(importance_type))
+
+    def getDenseFeatureShaps(self, vector):
+        """
+        Get the local shap feature importances.
+        """
+        return list(self._java_obj.getFeatureShaps(vector))
```
mhamilton723, line 64: Shouldn’t py4j make java arrays usable as lists on python side? Did this fix an error?

imatiach-msft: this is a method on the model, it wasn't available before.  getFeatureShaps needs to be changed to getDenseFeatureShaps (since it accepts vector on scala side), will fix that with another issue from another PR (shap values for multiclass case) in a follow-up.

PR #872, opened 2020-06-03
perf: reduce lightgbm prediction time by 20%

imatiach-msft:
perf: reduce lightgbm prediction time by 20%
depends on native changes:
https://github.com/microsoft/LightGBM/pull/3159
In testing, time went from 82159416923ns to 65852779813ns on the pima dataset (test added in PR but num iterations was 10000, not 100 as in PR for performance testing).


congnt-dsv:
@imatiach-msft When can you continue with this PR?

ranadeepsingh:
Closing as obsolete. This optimization depended on upstream [LightGBM #3159](https://github.com/lightgbm-org/LightGBM/pull/3159), which was closed unmerged after its benchmark was found invalid. Current SynapseML prediction code uses newer reusable native buffers and incompatible APIs, so this branch cannot be safely revived.

PR #882, opened 2020-06-24
fix: rename numCores to numWorkers in ClusterUtil

ocworld:
To avoid misunderstanding, in ClusterUtil, change rename words.
In my understanding, "numCores" is "numWorkers". 

I think, meaninig of "cores" is various like executor cores and tasks.

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
+++ b/src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
@@ -11,29 +11,29 @@ import org.apache.spark.sql.{Dataset, SparkSession}
 import org.slf4j.Logger
 
 object ClusterUtil {
-  /** Get number of cores from dummy dataset for 1 executor.
+  /** Get number of workers from dummy dataset for 1 executor.
     * Note: all executors have same number of cores,
     * and this is more reliable than getting value from conf.
     * @param dataset The dataset containing the current spark session.
-    * @return The number of cores per executor.
+    * @return The number of workers per executor.
     */
-  def getNumCoresPerExecutor(dataset: Dataset[_], log: Logger): Int = {
+  def getNumWorkersPerExecutor(dataset: Dataset[_], log: Logger): Int = {
     val spark = dataset.sparkSession
     val confTaskCpus = getTaskCpus(dataset, log)
     try {
       val confCores = spark.sparkContext.getConf.get("spark.executor.cores").toInt
-      val coresPerExec = confCores / confTaskCpus
-      log.info(s"ClusterUtils calculated num cores per executor as $coresPerExec from $confCores " +
+      val workersPerExec = confCores / confTaskCpus
+      log.info(s"ClusterUtils calculated num workers per executor as $workersPerExec from $confCores " +
         s"cores and $confTaskCpus task CPUs")
-      coresPerExec
+      workersPerExec
     } catch {
       case _: NoSuchElementException =>
         // If spark.executor.cores is not defined, get the cores based on master
         val defaultNumCores = getDefaultNumExecutorCores(spark, log)
-        val coresPerExec = defaultNumCores / confTaskCpus
-        log.info(s"ClusterUtils calculated num cores per executor as $coresPerExec from " +
+        val workersPerExec = defaultNumCores / confTaskCpus
+        log.info(s"ClusterUtils calculated num workers per executor as $workersPerExec from " +
           s"default num cores($defaultNumCores) from master and $confTaskCpus task CPUs")
-        coresPerExec
+        workersPerExec
     }
   }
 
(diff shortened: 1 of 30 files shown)


imatiach-msft:
@ocworld My understanding of terminology is the following:

https://stackoverflow.com/questions/32621990/what-are-workers-executors-cores-in-spark-standalone-cluster

image

Each machine or VM is a "worker" or sometimes referred to as a "node".  Each worker can have multiple executors, but often spark is configured with one executor per worker - but that is not always the case.  Also, each executor can have more than one core assigned to it, although often there is just one core per executor.  The executors run tasks, and usually in spark there are more tasks than executors, and the tasks are independent of each other.  However, in LightGBM we call NetworkInit to establish a ring of communication across workers, so the tasks are interdependent - hence we need to run all of the tasks at once.  If there are more tasks than executors, we either need to coalesce beforehand or, in the case of barrier execution mode, repartition (barrier execution does not allow coalesce in the same run).
I believe executors can run multiple tasks simultaneously, and that was the reasoning behind using "numCoresPerExec", for how many tasks can be run on an executor simultaneously, and "getNumExecutorCores" for the total number of executors * cores.  Let me look into this a bit more and play around with some clusters, as I recall this terminology seemed correct to me when I originally wrote this code (which was a while ago).  This terminology might not be correct for all cluster types though.  However, I think calling these workers might add more confusion since workers/nodes are individual machines that can sometimes hold multiple executors.

imatiach-msft:
@ocworld on the other hand, looking at this code:
```
    val numCoresPerExec = ClusterUtil.getNumCoresPerExecutor(dataset, log)
    val numExecutorCores = ClusterUtil.getNumExecutorCores(dataset, numCoresPerExec, log)
    val numWorkers = min(numExecutorCores, dataset.rdd.getNumPartitions)
```

numWorkers shouldn't be used either, maybe "numTasks" would be a better variable name, to avoid confusion between with spark workers/nodes.

src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
```diff
@@ -11,29 +11,29 @@ import org.apache.spark.sql.{Dataset, SparkSession}
...
-  /** Get number of cores from dummy dataset for 1 executor.
+  /** Get number of workers from dummy dataset for 1 executor.
     * Note: all executors have same number of cores,
     * and this is more reliable than getting value from conf.
     * @param dataset The dataset containing the current spark session.
-    * @return The number of cores per executor.
+    * @return The number of workers per executor.
     */
-  def getNumCoresPerExecutor(dataset: Dataset[_], log: Logger): Int = {
+  def getNumWorkersPerExecutor(dataset: Dataset[_], log: Logger): Int = {
```
imatiach-msft, line 20: maybe getNumTasksPerExecutor might be a better name?

ocworld:
@imatiach-msft I understand your consideration. Thanks to explain why you use "cores".
I think that terminology is very important and defining something is very difficult.

I agree with that "worker" is also confused. 
I've checked your comment and #881 

What do you think about using tasks instead of cores in not only variable but also function name?
```
val numTasksPerExec = ClusterUtil.getNumTasksPerExecutor(dataset, log)
val numExecutorTasks = ClusterUtil.getNumExecutorTasks(dataset, numTasksPerExec, log)
val numTasks = min(numExecutorTasks, dataset.rdd.getNumPartitions)
```

src/main/scala/com/microsoft/ml/spark/core/utils/ClusterUtil.scala
ocworld: Oh, when I write reply, you already commented about it :)

imatiach-msft:
I agree tasks is a slight improvement - on the other hand, it is still a bit confusing since tasks are the work that is run, not the "capacity" to run them, and a cluster can have more tasks than the number of executors * cores (in which case we try to prevent this before running lightgbm by coalesce in normal mode or repartition in barrier execution mode).  But, I am fine with renaming it to tasks, I think that may clear up more confusion since cores isn't a great name either.

ocworld:
> @ocworld I agree tasks is a slight improvement - on the other hand, it is still a bit confusing since tasks are the work that is run, not the "capacity" to run them, and a cluster can have more tasks than the number of executors * cores (in which case we try to prevent this before running lightgbm by coalesce in normal mode or repartition in barrier execution mode). But, I am fine with renaming it to tasks, I think that may clear up more confusion since cores isn't a great name either.

I agree with you about a "slight" improvement. Terminology is not easy always. It is changed and i pushed new commits now.

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBase.scala
```diff
@@ -144,30 +144,50 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
     }
   }
 
+  /**
+    * Retrieves the categorical indexes in the features column.
+    * @param df The dataset to train on.
+    * @return the categorical indexes in the features column.
+    */
+  private def getCategoricalIndexes(df: DataFrame): Array[Int] = {
```
imatiach-msft, line 152: there seems to be some merge conflict here?  Maybe your master branch is not updated to latest?

imatiach-msft:
@mhamilton723 could you merge this PR?  Thanks!

ocworld:
@ahnlab

PR #927, opened 2020-09-28
test: add infrastructure to run notebook tests on synapse

rohit21agrawal:


dciborow:
is this still active?

PR #939, opened 2020-10-21
README.md : adding pip install

JeanLescut:
there were no `pip install mmlspark` anywhere on the document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,9 +132,8 @@ documentation for [Scala](http://mmlspark.azureedge.net/docs/scala/) and
 
 ### Python
 
-To try out MMLSpark on a Python (or Conda) installation you can get Spark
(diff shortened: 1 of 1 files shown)


JeanLescut:
Deleting this PR... Sorry about that.
even with a `pip install mmlspark`, I cannot make the package work on my side
I noticed that version on pip is really weird (0.11111111), not sure it was linked...
Also I've tried something like :

```bash
python_path='python'
$python_path -m pip install --upgrade pip
$python_path -m pip install --upgrade setuptools
git clone https://github.com/Azure/mmlspark.git
cd mmlspark
git checkout v1.0.0-rc3
echo '__version__ = "v1.0.0-rc3"' >> src/main/python/mmlspark/__init__.py
$python_path -m pip install src/main/python
```

without sucess. I'm probably missing something obvious but it's okay

PR #945, opened 2020-10-28
fix: lazily creating a tempdir in NativeLoader

ocworld:
When predicting with a lightgbm model, there are so many lightgbm's so files on runtime.
It made failure on allocating memory on tmpfs in my application.

So, I tried to use java.library.path that has already "so" files.
When trying this approach, I found that it is possible to create so many /tmp/mml-nativesXXXX directories. Though, copied "so" files were not used.

Therefore, this pr is created to create lazily tempdir in NativeLoader to using getOrCreateTempDir() instead of making it in class's initializer.


mhamilton723:
ocworld thanks for making this PR I’ve seen some similiar things happen because the jvm exits abruptly. Perhaps another way to help this problem is to postfix the temp die with the MMLSpark version instead of a random string so at most one of these is made per version. This might also affect @eisber too right?

eisber:
not sure I get the change. The class initializer is run, but the extract not?
Apart from that the change looks functionality equivalent.

imatiach-msft (approved):

ocworld:
Because of many changes on master for a year, It is hard to merge. So, I'll create new pr based on the current master after this pr is closed.

PR #955, opened 2020-11-17
fix: Fix processing sparse vector size for getFeatureShaps

alexandrseverinov:
Currently `getFeatureShaps` does not work neither on dense vectors nor sparse ones.
For dense vectors `getDenseFeatureShaps` is not implemented.
But for sparse vectors error comes earlier: `vector.size` is `int` and should be passed further as it is.
This pull request fixes the error with processing field `size` of sparse vectors


mhamilton723:
Thank you @alexandrseverinov

PR #982, opened 2021-01-31
fix: Recommender Patch for Spark 3 Update

dciborow:
Fixing some bugs that popped up with Spark 3.


dciborow:
ready to be squashed
--

mhamilton723 (approved):

PR #994, opened 2021-03-01
fix: cleanup python tests

dciborow:


src/main/python/setup.py
```diff
@@ -6,46 +6,41 @@
...
-
     classifiers=[
         "Development Status :: 3 - Alpha",
         "Intended Audience :: Developers",
         "Intended Audience :: Data Scientists",
         "Topic :: Software Development :: Datascience Tools",
         "License :: OSI Approved :: MIT License",
         "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 3"
+        "Programming Language :: Python :: 3",
```
dciborow, line 42: i ran "Black" on this file. It is a python auto-formatter that helps standardize some styling like use of ".

mhamilton723 (approved):

PR #1009, opened 2021-03-16
Adds Spark and Open MPI

brockneedscoffee:
When installing spark via miniconda the spark home is not set so you cannot run spark in AML.  You will get errors when you try to set the spark context or if you set the framework to pyspark in an AML pipeline step.  This PR installs spark, sets the home, and installs open mpi dependency.

### DIFF
--- a/tools/docker/demo/Dockerfile
+++ b/tools/docker/demo/Dockerfile
@@ -1,7 +1,9 @@
 FROM ubuntu:16.04
 
-ARG SPARK_VERSION=3.0.1
-RUN apt-get -qq update && apt-get -qq -y install curl bzip2 \
+ENV SPARK_VERSION=3.0.2
+ENV HADOOP_VERSION=2.7
+
+RUN apt-get -qq update && apt-get -qq -y install curl bzip2 wget \
     && curl -sSL https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -o /tmp/miniconda.sh \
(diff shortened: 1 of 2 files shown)


mhamilton723:
@brockneedscoffee - looks good but we dont test docker notebook as part of the build yet so could you manually test these out?

Also could you group all the run steps together into a single run step and put ENV stuff in top of docker. Putting all of the run steps together (and deleting ant .tgzs you download in these steps) makes the docker image considerably smaller thanks!

tools/docker/demo/Dockerfile
```diff
@@ -1,23 +1,50 @@
...
+# Install Spark and move it to the folder "/opt/spark"
+RUN tar -xzf spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz && \
+    mv spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION} /opt/spark && \
+    rm spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz 
+
+ENV SPARK_HOME /opt/spark
 ENV PATH /opt/conda/bin:$PATH
 ENV JAVA_HOME /usr/lib/jvm/java-1.8.0-openjdk-amd64
 
+# Build-essentials
```
mhamilton723, line 31: Group run steps together for smaller docker image

mhamilton723 (changes requested):

imatiach-msft (approved):

PR #1019, opened 2021-04-13
feat: Add MMLSpark Logs

serena-ruan:


src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala
```diff
@@ -100,6 +102,7 @@ class LightGBMClassificationModel(override val uid: String)
     * @return transformed dataset
     */
   override def transform(dataset: Dataset[_]): DataFrame = {
+    logInfo("Calling function transform --- telemetry record")
```
imatiach-msft, line 105: do you need "--- telemetry record"?  Also can you include the class name, like "Calling LightGBMClassificationModel transform"?

src/test/scala/com/microsoft/ml/spark/logging/LoggingTest.scala
```diff
@@ -0,0 +1,337 @@
...
+import org.apache.spark.sql.functions.{col, collect_list, lit, struct}
+import org.apache.spark.sql.{DataFrame, Row}
+
+import java.net.URL
+
+class LoggingTest extends TestBase with TextKey with CognitiveKey with AnomalyKey{
+
+  import spark.implicits._
+
+  test("TextSentiment test"){
```
imatiach-msft, line 22: how is this testing logging?

serena-ruan: Hi this is just a draft for test, please ignore it. Thanks still!

mhamilton723:
One thing that might make this a bit easier to edit down the road is to abstract the information you want to log into a trait

```scala
trait BasicLogging {

  def logFit(): Unit = {
      stuff
  }

  def logTransform(): Unit = {
     stuff
  }

}```

this way when you want to change the structure of the log output you only need to change it in one place later.

**Comment by serena-ruan** [comment]
> One thing that might make this a bit easier to edit down the road is to abstract the information you want to log into a trait
> 
> ```scala
> trait BasicLogging {
> 
>   def logFit(): Unit = {
>       stuff
>   }
> 
>   def logTransform(): Unit = {
>      stuff
>   }
> 
> }
> ```
> 
> this way when you want to change the structure of the log output you only need to change it in one place later.

File: src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala (hunk 1 above)
Comment (serena-ruan) [reply src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMClassifier.scala]: Actually logInfo automatically fetches the classname, that's why I don't add it in the message.

File: src/main/scala/com/microsoft/ml/spark/automl/FindBestModel.scala [hunk 3]
```diff
@@ -46,7 +47,8 @@ trait FindBestModelParams extends Wrappable with ComplexParamsWritable with HasE
 }
 
 /** Evaluates and chooses the best model from a list of models. */
-class FindBestModel(override val uid: String) extends Estimator[BestModel] with FindBestModelParams {
+class FindBestModel(override val uid: String) extends Estimator[BestModel] with FindBestModelParams with BasicLogging{
```
eisber, line 50: space missing
```suggestion
class FindBestModel(override val uid: String) extends Estimator[BestModel] with FindBestModelParams with BasicLogging {
```

src/main/scala/com/microsoft/ml/spark/automl/FindBestModel.scala
```diff
@@ -176,7 +180,10 @@ class BestModel(val uid: String) extends Model[BestModel]
 
   override def copy(extra: ParamMap): BestModel = defaultCopy(extra)
 
-  override def transform(dataset: Dataset[_]): DataFrame = getBestModel.transform(dataset)
+  override def transform(dataset: Dataset[_]): DataFrame = {
+    logTransform(uid)
```
eisber, line 184: could this be achieved using the [Mixin Composition Stacks](http://jonasboner.com/aop-style-mixin-composition-stacks-in-scala/)

src/main/scala/com/microsoft/ml/spark/cognitive/AnamolyDetection.scala
```diff
@@ -196,6 +200,7 @@ class SimpleDetectAnomalies(override val uid: String) extends AnomalyDetectorBas
   }
 
   override def transform(dataset: Dataset[_]): DataFrame = {
+    logTransform(uid)
```
eisber, line 203: curious if it would be good to pass the dataset to logTransform to enable richer logging (e.g. colums names, row counts,...)

src/main/scala/com/microsoft/ml/spark/cognitive/CognitiveServiceBase.scala
```diff
@@ -25,7 +25,7 @@ import spray.json.DefaultJsonProtocol._
 import scala.collection.JavaConverters._
 import scala.language.existentials
 
-trait HasServiceParams extends Params {
+trait HasServiceParams extends Params{
```
eisber, line 28:
```suggestion
trait HasServiceParams extends Params {
```

src/test/scala/com/microsoft/ml/spark/core/test/base/SparkSessionFactory.scala
```diff
@@ -47,7 +47,7 @@ object SparkSessionFactory {
     val sess = SparkSession.builder()
       .config(conf)
       .getOrCreate()
-    sess.sparkContext.setLogLevel(logLevel)
+    sess.sparkContext.setLogLevel("INFO")
```
eisber, line 50: done on purpose?

src/main/scala/com/microsoft/ml/spark/logging/BasicLogging.scala
```diff
@@ -0,0 +1,42 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark.logging
+
+import org.apache.spark.internal.Logging
+
+trait BasicLogging extends Logging{
```
eisber, line 8:
```suggestion
trait BasicLogging extends Logging {
```

src/main/scala/com/microsoft/ml/spark/automl/FindBestModel.scala
serena-ruan: Do you mean that we could override the transform function by extending the trait/class? I log all transform functions for different classes so it looks not applicable to extend them one by one. Pls correct me if I didn't get you!

src/main/scala/com/microsoft/ml/spark/cognitive/AnamolyDetection.scala
serena-ruan: This sounds good! I'll look into that.

src/test/scala/com/microsoft/ml/spark/core/test/base/SparkSessionFactory.scala
serena-ruan: Yep, otherwise during tests at some point the root log level will be set as warning and I can't see info-level logs.

src/main/scala/com/microsoft/ml/spark/automl/FindBestModel.scala
eisber: I think it's worth a try. if you define override def transform(dataset: Dataset[_]): DataFrame = { ... }  in the Loggable trait, which method is called first? The Loggable or the FindBestModel? if it's the Loggable trait, I think you won, as you don't need the individual logTransform(uid) calls...

src/test/scala/com/microsoft/ml/spark/core/test/base/SparkSessionFactory.scala
mhamilton723: I dont think this will affect the deployed Jar so I would suggest keeping it at the logLevel for now. This only affects tests in the build

src/main/scala/com/microsoft/ml/spark/logging/BasicLogging.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import LogJsonProtocol._
+import spray.json._
+
+trait BasicLogging extends Logging {
+
+  val uid: String
+
+  protected def logBase(methodName: String): Unit = {
+    logInfo("metrics/ " + BasicLogInfo(uid, getClass.toString, methodName).toJson.toString())
```
mhamilton723, line 27:
```suggestion
    logInfo("metrics/ " + BasicLogInfo(uid, getClass.toString, methodName).toJson.compactPrint())
```
not sure if toString is properly supported in all cases. this is the explicit method that stringifyies in spray

mhamilton723 (changes requested):
Nice!! Looks great one minor nit (sorry to be a pain in the ass :) )

src/main/scala/com/microsoft/ml/spark/logging/BasicLogging.scala
serena-ruan: I tried compactPrint previously and I'm not sure what's the issue that causes the logs can't be properly output to Kusto, in Kusto table they show as 'metrics/ {"[REDACTED]"}'.. So I'm trying to see if toString will work.

serena-ruan: Figured out that this has nothing to do with compactPrint, changed it back. That issue is related to log4j properties setup.

mhamilton723 (approved):

PR #1024, opened 2021-04-16
feat: add num iteration and start iteration to lightgbm model

imatiach-msft:
- add num iteration and start iteration to lightgbm model
- set num iteration to be best iteration when early stopping criteria reached
resolves issue https://github.com/Azure/mmlspark/issues/1006


src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMModelMethods.scala
```diff
@@ -46,6 +46,22 @@ trait LightGBMModelMethods extends LightGBMModelParams {
     getLightGBMBooster.featuresShap(Vectors.sparse(size, indices, values))
   }
 
+  /** Public method to sets the start index of the iteration to predict.
+    * If <= 0, starts from the first iteration.
+    * @param startIteration The start index of the iteration to predict.
+    */
+  def setStartIteration(startIteration: Int): Unit = {
```
mhamilton723, line 53: Could you please try to make this a standard sparkml param so we don't need to add to python bindings? Thanks!

mhamilton723 (changes requested):

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMModelMethods.scala
imatiach-msft: done!

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMBooster.scala
```diff
@@ -192,6 +192,10 @@ class LightGBMBooster(val model: String) extends Serializable {
     new BoosterHandler(model)
   }
 
+  var bestIteration: Int = -1
```
mhamilton723, line 195: Can you make these sparkML params pleas?

mhamilton723 (changes requested):

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMModelMethods.scala
```diff
@@ -63,4 +106,13 @@ trait LightGBMModelMethods extends LightGBMModelParams {
   protected def featuresShap(features: Vector): Vector = {
     Vectors.dense(getLightGBMBooster.featuresShap(features))
   }
+
+  protected def updateBoosterParamsBeforePredict(): Unit = {
+    if (getNumIterations != LightGBMConstants.DefaultNumIterations) {
```
mhamilton723, line 111: i dont think these ifs are saving much. COnsider simplifying to just the setters

mhamilton723 (changes requested):
One small nit but other than that LGTM

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMModelMethods.scala
```diff
@@ -63,4 +106,13 @@ trait LightGBMModelMethods extends LightGBMModelParams {
...
+      getModel.setNumIterations(getNumIterations)
+    }
+    if (getStartIteration != LightGBMConstants.DefaultStartIteration) {
+      getModel.setStartIteration(getStartIteration)
+    }
+  }
```
mhamilton723, line 117:
```suggestion
  protected def updateBoosterParamsBeforePredict(): Unit = {
      getModel.setNumIterations(getNumIterations)
      getModel.setStartIteration(getStartIteration)
  }
```

Wouldnt this yield the same behavior?

mhamilton723 (changes requested):

src/main/scala/com/microsoft/ml/spark/lightgbm/LightGBMModelMethods.scala
imatiach-msft: ah, I don't want to set numIterations if the best iteration was set on the booster... let me think, maybe I can clean this up a bit more... certainly for startIteration it is fine..

imatiach-msft: I think I have a more readable solution: I set
```
new LightGBMClassificationModel(uid)
...
.setNumIterations(lightGBMBooster.bestIteration)
```
on each of the models when they are created by the learner, so I can remove this somewhat confusing check here

mhamilton723 (approved):

PR #1052, opened 2021-05-17
feat: add matrix type parameter and improve auto logic

imatiach-msft:
- add matrix type parameter to specify whether matrix is sparse, dense, or auto inferred
- allow user to override default auto logic to specify whether sparse or dense matrix must be constructed in lightgbm native code
- improve autologic to use first 10 sampled rows instead of just the head row when trying to determine whether constructed matrix should be sparse or dense

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
@@ -29,14 +29,23 @@ private object TrainUtils extends Serializable {
   def generateDataset(rowsIter: Iterator[Row], columnParams: ColumnParams,
                       referenceDataset: Option[LightGBMDataset], schema: StructType,
                       log: Logger, trainParams: TrainParams): Option[LightGBMDataset] = {
-    val hrow = rowsIter.next()
+    val (concatRowsIter: Iterator[Row], isSparse: Boolean) =
+      if (trainParams.executionParams.matrixType == "auto") {
+        sampleRowsForArrayType(rowsIter, schema, columnParams)
+      } else if (trainParams.executionParams.matrixType == "sparse") {
+        (rowsIter: Iterator[Row], true)
+      } else if (trainParams.executionParams.matrixType == "dense") {
+        (rowsIter: Iterator[Row], false)
+      } else {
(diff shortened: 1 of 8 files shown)


src/test/scala/com/microsoft/ml/spark/lightgbm/split1/VerifyLightGBMClassifier.scala
```diff
@@ -320,12 +320,15 @@ class VerifyLightGBMClassifier extends Benchmarks with EstimatorFuzzing[LightGBM
   test("Verify LightGBM Classifier with dart mode parameters") {
     // Assert the dart parameters work without failing and setting them to tuned values improves performance
     val Array(train, test) = pimaDF.randomSplit(Array(0.8, 0.2), seed)
-    val scoredDF1 = baseModel.setBoostingType("dart").fit(train).transform(test)
+    val scoredDF1 = baseModel.setBoostingType("dart").
```
imatiach-msft, line 323: note: this is just fixing a flaky test in the build, unrelated to this PR

src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
```diff
@@ -60,6 +69,34 @@ private object TrainUtils extends Serializable {
...
+  /**
+    * Sample the first several rows to determine whether to construct sparse or dense matrix in lightgbm native code.
+    * @param rowsIter  Iterator of rows.
+    * @param schema The schema.
+    * @param columnParams The column parameters.
+    * @return A reconstructed iterator with the same original rows and whether the matrix should be sparse or dense.
+    */
+  def sampleRowsForArrayType(rowsIter: Iterator[Row], schema: StructType,
+                             columnParams: ColumnParams): (Iterator[Row], Boolean) = {
+    var sampleRow = 0
```
mhamilton723, line 81: Do we need to look at all 10? perhaps just using the first value would give the same clue and allow a cleaner implementation

mhamilton723 (changes requested):

src/main/scala/com/microsoft/ml/spark/lightgbm/TrainUtils.scala
imatiach-msft: discussed offline, cleaned up the code a bit more

mhamilton723 (approved):

PR #1054, opened 2021-05-20
feat: add custom objective function to lightgbm learners

imatiach-msft:
add support for custom objective function in lightgbm learners

Users in the scala API will now be able to supply a custom loss function for training, which will need to output the gradient and hessian matrix.

As part of this PR, the LightGBMBooster and LightGBMDataset classes have been improved and moved under the booster and dataset folders, and some of the files related to parameters have been moved to a separate params folder in order to organize the code better since there are so many files in lightgbm directory.

TODO: add pyspark support, in a separate PR

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
+++ b/src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
@@ -1,12 +1,15 @@
 // Copyright (C) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in project root for information.
 
-package com.microsoft.ml.spark.lightgbm
+package com.microsoft.ml.spark.lightgbm.booster
 
 import com.microsoft.ml.lightgbm._
+import com.microsoft.ml.spark.lightgbm.{LightGBMConstants, LightGBMUtils}
 import com.microsoft.ml.spark.lightgbm.LightGBMUtils.getBoosterPtrFromModelString
+import com.microsoft.ml.spark.lightgbm.dataset.LightGBMDataset
 import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector}
 import org.apache.spark.sql.{SaveMode, SparkSession}
+import org.slf4j.Logger
 
 //scalastyle:off
 protected abstract class NativePtrHandler[T](val ptr: T) {
@@ -32,13 +35,17 @@ protected class LongLongNativePtrHandler(ptr: SWIGTYPE_p_long_long) extends Nati
 
 /** Wraps the boosterPtr and guarantees that Native library is initialized
  * everytime it is needed
- * @param model The string serialized representation of the learner
+ * @param boosterPtr The pointer to the native lightgbm booster
  */
-protected class BoosterHandler(model: String) {
-  LightGBMUtils.initializeNativeLibrary()
+protected class BoosterHandler(var boosterPtr: SWIGTYPE_p_void) {
 
-  var boosterPtr: SWIGTYPE_p_void = {
-    getBoosterPtrFromModelString(model)
+  /** Wraps the boosterPtr and guarantees that Native library is initialized
+    * everytime it is needed
+    *
+    * @param model The string serialized representation of the learner
+    */
+  def this(model: String) = {
+    this(getBoosterPtrFromModelString(model))
   }
 
   val scoredDataOutPtr: ThreadLocal[DoubleNativePtrHandler] = {
@@ -127,6 +134,13 @@ protected class BoosterHandler(model: String) {
   lazy val dataInt32bitType: Int = lightgbmlibConstants.C_API_DTYPE_INT32
   lazy val data64bitType: Int = lightgbmlibConstants.C_API_DTYPE_FLOAT64
 
+  def freeNativeMemory(): Unit = {
+    if (boosterPtr != null) {
+      LightGBMUtils.validate(lightgbmlib.LGBM_BoosterFree(boosterPtr), "Finalize Booster")
+      boosterPtr = null
+    }
+  }
+
   private def getNumClasses: Int = {
     val numClassesOut = lightgbmlib.new_intp()
     LightGBMUtils.validate(
@@ -167,13 +181,6 @@ protected class BoosterHandler(model: String) {
     out
   }
 
-  private def freeNativeMemory(): Unit = {
-    if (boosterPtr != null) {
-      LightGBMUtils.validate(lightgbmlib.LGBM_BoosterFree(boosterPtr), "Finalize Booster")
-      boosterPtr = null
-    }
-  }
-
   override protected def finalize(): Unit = {
     freeNativeMemory()
     super.finalize()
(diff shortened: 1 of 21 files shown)


src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
```diff
@@ -167,35 +200,129 @@ protected class BoosterHandler(model: String) {
...
+      newParameters), "Booster Reset learning_rate Param")
+  }
+
+  def innerPredict(dataIndex: Int, classification: Boolean): Array[Array[Double]] = {
+    val kind = boosterHandler.normalScoreConstant
+    lightgbmlib.LGBM_BoosterGetPredict(boosterHandler.boosterPtr, dataIndex,
+      boosterHandler.customObjectiveScoredDataLengthLongPtr.get().ptr,
+      boosterHandler.customObjectiveScoredDataOutPtr.get().ptr)
+    // TODO: fixme
+    Array(predScoreToArray(classification, boosterHandler.customObjectiveScoredDataOutPtr.get().ptr, kind))
```
imatiach-msft, line 323: TODO: fixme

src/test/scala/com/microsoft/ml/spark/lightgbm/split1/VerifyLightGBMClassifier.scala
```diff
@@ -86,6 +87,7 @@ trait LightGBMTestUtils extends TestBase {
   }
 
   def assertProbabilities(tdf: DataFrame, model: LightGBMClassifier): Unit = {
+    println(model.getModelString)
```
imatiach-msft, line 90: remove

src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
imatiach-msft: done

src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
```diff
@@ -1,12 +1,15 @@
 // Copyright (C) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in project root for information.
 
-package com.microsoft.ml.spark.lightgbm
+package com.microsoft.ml.spark.lightgbm.booster
```
mhamilton723, line 4: why the namespace change?

src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
```diff
@@ -167,35 +182,155 @@ protected class BoosterHandler(model: String) {
...
+    if (trainDataset.isEmpty) {
+      new BoosterHandler(model)
+    } else {
+      LightGBMUtils.validate(lightgbmlib.LGBM_BoosterCreate(trainDataset.map(_.datasetPtr).get,
+        parameters.get, boosterOutPtr), "Booster")
+      new BoosterHandler(lightgbmlib.voidpp_value(boosterOutPtr))
+    }
   }
 
+  val model = modelStr.getOrElse("")
```
mhamilton723, line 230: what is the behavior if this is ""? Seems this should be kept as option especially if code elsewhere checks for ""

src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
```diff
@@ -167,35 +182,155 @@ protected class BoosterHandler(model: String) {
...
 /** Represents a LightGBM Booster learner
-  * @param model The string serialized representation of the learner
+  * @param trainDataset The training dataset
+  * @param parameters The booster initialization parameters
+  * @param modelStr Optional parameter with the string serialized representation of the learner
   */
 @SerialVersionUID(777L)
-class LightGBMBooster(val model: String) extends Serializable {
-  /** Transient variable containing local machine's pointer to native booster
+class LightGBMBooster(val trainDataset: Option[LightGBMDataset] = None, val parameters: Option[String] = None,
```
mhamilton723, line 197: will this break serialization?

mhamilton723 (dismissed):

src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
imatiach-msft: moving params, booster and dataset into their own folders, because there are way too many files in lightgbm folder

src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
imatiach-msft: model string should never be empty

src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
imatiach-msft: it's still Serializable, see line below

src/main/scala/com/microsoft/ml/spark/lightgbm/booster/LightGBMBooster.scala
imatiach-msft: removed the model

src/main/scala/com/microsoft/ml/spark/lightgbm/params/FObjTrait.scala
```diff
@@ -0,0 +1,17 @@
...
+
+trait FObjTrait extends Serializable {
+  /**
+    * User defined objective function, returns gradient and second order gradient
+    *
+    * @param predictions untransformed margin predicts
+    * @param trainingData training data
+    * @return List with two float array, correspond to grad and hess
+    */
+  def getGradient(predictions: Array[Array[Double]], trainingData: LightGBMDataset): List[Array[Float]]
```
mhamilton723, line 16: Might want to make this return two objects instead of a list. Also in general the grad and hessian wont be the same shape...

src/main/scala/com/microsoft/ml/spark/lightgbm/params/FObjTrait.scala
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark.lightgbm.params
+
+import com.microsoft.ml.spark.lightgbm.dataset.LightGBMDataset
+
+trait FObjTrait extends Serializable {
```
mhamilton723, line 8: Perhaps we should rename this to CustomObjective

mhamilton723 (changes requested):

src/main/scala/com/microsoft/ml/spark/lightgbm/params/FObjTrait.scala
imatiach-msft: I think that's what it is called in XGBoost actually, but in lightgbm the parameter is called fobj, so I wanted to keep in line with the terminology in lightgbm

src/main/scala/com/microsoft/ml/spark/lightgbm/params/FObjTrait.scala
imatiach-msft: interesting idea, in xgboost I see they return one list:
https://github.com/dmlc/xgboost/blob/4224c08cacceba3f83f90e387c07aa6205a83bfa/jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/scala/ObjectiveTrait.scala#L32
I can return two objects but for those familiar with xgboost it might be harder to get used to a slightly different API

imatiach-msft: done!  changes it to return a tuple

src/main/scala/com/microsoft/ml/spark/lightgbm/params/FObjTrait.scala
mhamilton723: SG

mhamilton723 (dismissed):

PR #1061, opened 2021-05-27
feat: add singleton dataset mode for faster performance

imatiach-msft:
Adding single (or "singleton") dataset mode to lightgbm learners.
User can enable this new mode by setting the parameter useSingleDatasetMode=True (it is false by default).
In this mode, each executor creates a single LightGBMDataset.  By default, currently each task within an executor creates a dataset:

image

In this PR, a new mode is added to only create one dataset per executor:

image

This means that there is lower network communication overhead since fewer nodes are initialized and more parallelization is done within the machine in the native code with default number of threads.  This also seems to reduce memory usage significantly for some datasets.

Note in most cluster configurations there is usually only one executor per machine anyway.

In performance tests, we've found this mode sometimes outperforms the default in certain scenarios, both in terms of memory and execution time.

On a sparse dataset with 9 GB of data and large parameter values (num_leaves=768, num_trees=1000, min_data_in_leaf=15000, max_bin=512) and 5 machines with 8 cores and 28 GB of RAM, runtime was 17.54 minutes with this new mode. When specifying tasks=5 it took 106 minutes and in default mode it failed with OOM.  

However in other scenarios the default mode is much faster.
On dense Higgs dataset (4GB) with default parameters and 8 workers with 14 GB memory, 4 cores each the default run took 54 seconds but new single dataset mode took 2.2 minutes, twice as slow.

For this reason we will keep this mode as non-default for now as we continue to do more benchmarking/experimentation.

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/io/http/SharedVariable.scala
+++ b/src/main/scala/com/microsoft/ml/spark/io/http/SharedVariable.scala
@@ -49,7 +49,6 @@ class SharedSingleton[T: ClassTag](constructor: => T) extends AnyRef with Serial
   }
 
   def get: T = instance
-
 }
 
 object SharedSingleton {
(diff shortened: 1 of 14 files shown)


mhamilton723 (changes requested):
This seems like it complicates the implementation a lot. Perhaps we can decouple the dataset aggregation from lightgbm. In particular, this is doing something very similar to the partition consolidator. COuld the logic be held in that class and then applied for the specific case of LGBM. Seems if we decouple than all lightGBM needs to do is skip partitions without rows.

imatiach-msft:
closing as superseded by https://github.com/Azure/mmlspark/pull/1066

PR #1077, opened 2021-06-09
feat: new LIME and KernelSHAP explainers

memoryz:
In this PR, we rewrote the LIME explainers and added KernelSHAP explainers in the `com.microsoft.ml.spark.explainers` package.

New features:
- KernelSHAP explainer for tabular, vector, image and text models.
- LIME explainer now supports kernel width and sample weights.
- Both explainer support categorical variable (in tabular explainer).
- Both explainers report r-squared metric from the underlying regression model.
- Both explainers support explaining multiple classes output in one run.
- For tabular and vector models, both explainers support passing in a background dataframe. ~~If one is not given, the dataframe used for local interpretation will be used as background data.~~

Sample notebooks will be included in the next PR.


src/main/scala/com/microsoft/ml/spark/codegen/Wrappable.scala
```diff
@@ -134,10 +134,14 @@ trait PythonWrappable extends BaseWrappable {
...
 
+  private def escape(raw: String): String = {
+    StringEscapeUtils.escapeJava(raw)
+  }
+
   protected lazy val pyParamsDefinitions: String = {
     this.params.map { p =>
       val typeConverterString = getParamInfo(p).pyTypeConverter.map(", typeConverter=" + _).getOrElse("")
-      s"""|${p.name} = Param(Params._dummy(), "${p.name}", "${p.doc}"$typeConverterString)
+      s"""|${p.name} = Param(Params._dummy(), "${p.name}", "${escape(p.doc)}"$typeConverterString)
```
mhamilton723, line 144: TY!

src/main/scala/com/microsoft/ml/spark/core/schema/DatasetExtensions.scala
```diff
@@ -51,13 +48,8 @@ object DatasetExtensions {
...
     */
   def findUnusedColumnName(prefix: String)(columnNames: scala.collection.Set[String]): String = {
-    var counter = 2
-    var unusedColumnName = prefix
-    while (columnNames.contains(unusedColumnName)) {
-      unusedColumnName += "_" + counter
-      counter += 1
-    }
-    unusedColumnName
+    val stream = Iterator(prefix) ++ Iterator.from(1, 1).map(prefix + "_" + _)
```
mhamilton723, line 51: 😎

src/main/python/mmlspark/explainers/TabularLIME.py
```diff
@@ -0,0 +1,8 @@
+from mmlspark.explainers.HasBackgroundData import HasBackgroundData
+from mmlspark.explainers._TabularLIME import _TabularLIME
+from pyspark.ml.common import inherit_doc
+
+
+@inherit_doc
```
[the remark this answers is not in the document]
memoryz: Removed.

src/test/scala/com/microsoft/ml/spark/explainers/LassoRegressionSuite.scala
```diff
@@ -0,0 +1,103 @@
...
+  // clf.fit(X, y)
+  // print(clf.coef_)
+  // print(clf.intercept_)
+  test("LassoRegression should regress correctly without weights with intercept") {
+    val a = BDM((1.0, 1.0), (2.0, -2.0), (3.0, 3.0), (4.0, 5.0))
+    val b = BDV(1.01, 1.98, 3.03, 4.05)
+
+    val result = new LassoRegression(0.05d).fit(a, b, fitIntercept = true)
+
+    assert(norm(result.coefficients - BDV(0.94072731, 0.02135768)) < 1e-3, "coefficients disagreed")
```
mhamilton723, line 25: Perhaps we can use tolerant equality in scalatest for these

src/test/scala/com/microsoft/ml/spark/explainers/LeastSquaresRegressionSuite.scala
```diff
@@ -0,0 +1,115 @@
...
+    assert(abs(result.coefficients(0) - 0.01518619) < 1e-6)
+    assert(abs(result.coefficients(1) - 0.08135246) < 1e-6)
+    assert(abs(result.coefficients(2) - 0.082494) < 1e-6)
+    assert(abs(result.rSquared - 0.43565371571037637) < 1e-6)
+    assert(abs(result.loss - 8.394650978808151) < 1e-6)
+  }
+
+  test("LeastSquaresRegression should regress correctly with weights without intercept.") {
+    // Validated with the following code in Jupyter notebook:
+    // import numpy as np
```
mhamilton723, line 38: not sure what you mean by this.

mhamilton723 (changes requested):
So great @memoryz ! Only minor nits and comments

src/main/scala/com/microsoft/ml/spark/explainers/KernelSHAPBase.scala
```diff
@@ -0,0 +1,130 @@
...
+      val denominator = comb(coalition.size, activeSize.toInt) * activeSize * inactiveSize
+      numerator / denominator
+    }
+  }
+
+  private[explainers] def getEffectiveNumSamples(numSamplesParam: Option[Int], numFeature: Int): Int = {
+    val minSamplesNeeded = numFeature
+    val maxSamplesNeeded = math.pow(2, numFeature)
+
+    val value = numSamplesParam.getOrElse(2 * numFeature + 2048)
```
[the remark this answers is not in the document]
memoryz: done.

src/test/scala/com/microsoft/ml/spark/core/test/fuzzing/FuzzingTest.scala
```diff
@@ -178,7 +182,6 @@ class FuzzingTest extends TestBase {
     pipelineStages.foreach { pipelineStage =>
       pipelineStage.params.foreach { param =>
         assertOrLog(!param.name.contains(badChars), param.name)
-        assertOrLog(!param.doc.contains("\""), param.doc)
```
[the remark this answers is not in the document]
memoryz: the badChars are used for checking param names. I feel that is still necessary.

src/main/scala/com/microsoft/ml/spark/explainers/TextSHAP.scala
```diff
@@ -0,0 +1,97 @@
...
+    df.withColumn("samples", explode(samplesUdf(col(getTokensCol))))
+      .select(
+        col(idCol),
+        col("samples.coalition").alias(coalitionCol),
+        col("samples.sample").alias(getInputCol)
+      )
+  }
+
+  private def getSampleSchema: DataType = {
+    ArrayType(StructType(Seq(
```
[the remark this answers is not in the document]
memoryz: done.

src/main/scala/com/microsoft/ml/spark/explainers/TextLIME.scala
```diff
@@ -0,0 +1,104 @@
...
+class TextLIME(override val uid: String)
+  extends LIMEBase(uid) with TextLIMEParams {
+
+  logClass()
+
+  def this() = {
+    this(Identifiable.randomUID("TextLIME"))
+  }
+
+  override protected def preprocess(df: DataFrame): DataFrame = {
```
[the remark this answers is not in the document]
memoryz: done.

src/main/scala/com/microsoft/ml/spark/explainers/TextLIME.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.functions.{col, explode}
+import org.apache.spark.sql.types._
+
+trait TextLIMEParams extends LIMEParams with HasSamplingFraction with HasInputCol {
+  self: TextLIME =>
+
+  def setInputCol(value: String): this.type = this.set(inputCol, value)
+
+  val tokensCol = new Param[String](this,
```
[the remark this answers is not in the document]
memoryz: done.

src/main/scala/com/microsoft/ml/spark/explainers/KernelSHAPBase.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+object KernelSHAPBase {
+  private[explainers] def kernelWeight(coalition: SV): Double = {
+    val activeSize = sum(coalition.toBreeze)
+    val inactiveSize = coalition.size - activeSize
+
+    if (activeSize == 0 || inactiveSize == 0) {
+      1E6
```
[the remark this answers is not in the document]
memoryz: done.

src/test/scala/com/microsoft/ml/spark/core/test/fuzzing/FuzzingTest.scala
```diff
@@ -231,8 +234,17 @@ class FuzzingTest extends TestBase {
       "com.microsoft.ml.spark.lightgbm.LightGBMRegressionModel",
       "com.microsoft.ml.spark.vw.VowpalWabbitClassificationModel",
       "com.microsoft.ml.spark.vw.VowpalWabbitRegressionModel",
-      "com.microsoft.ml.spark.vw.VowpalWabbitContextualBanditModel"
+      "com.microsoft.ml.spark.vw.VowpalWabbitContextualBanditModel",
+      "com.microsoft.ml.spark.explainers.ImageLIME",
```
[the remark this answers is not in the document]
memoryz: These ones are about the HasInputCol mixins. I intentionally used the Spark's built-in `HasInputCol` trait, rather than mmlspark's override `com.microsoft.ml.spark.core.contracts.HasInputCol`. The difference is really just the `setInputCol` method, but I feel using Spark's built-in trait is safer, in the case where a customer may want to use the Spark's built-in `HasInputCol` trait to do pattern matching, or reflection.

src/test/scala/com/microsoft/ml/spark/explainers/LassoRegressionSuite.scala
memoryz: done.

src/test/scala/com/microsoft/ml/spark/explainers/LeastSquaresRegressionSuite.scala
memoryz: You mean the comment or this test itself?
This test itself validates that the LeastSquaresRegression returns correct result when we specify non-trivial sample weights but without fitting intercept.

The comment here just documents the python code that runs the same regression with same inputs/outputs/weights, but runs the regression through sklearn. We expect the results to match between scala implementation and sklearn.

src/test/scala/com/microsoft/ml/spark/explainers/HasExplainTargetSuite.scala
```diff
@@ -0,0 +1,53 @@
...
+      .getExplainTarget(df.schema)
+
+    val Tuple1(v1) = df.select(target1).as[Tuple1[SV]].head
+    assert(v1 == SVS.dense(1d, 3d))
+
+    // vector
+    val target2 = LocalExplainer.LIME.vector
+      .setTargetCol("label2")
+      .setTargetClasses(Array(0, 2))
+      .getExplainTarget(df.schema)
```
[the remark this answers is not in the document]
memoryz: This is to test the `getExplainTarget` function in the `HasExplainTarget` trait. The `getExplainTarget` function supports a variety of target column types: 
- `NumericType`: in the case of a regression model
- `VectorType`: in the case of a typical classification model with probability output
- `ArrayType(NumericType)`: in the case where the output was converted to an array of numeric types.
- `MapType(IntegerType, NumericType)`: this is to support ZipMap type of output for sklearn models via ONNX runtime.

The testing code here tests that each type can be successfully parsed and doesn't trigger error at runtime.

src/test/scala/com/microsoft/ml/spark/core/test/fuzzing/FuzzingTest.scala
```diff
@@ -84,32 +80,40 @@ class FuzzingTest extends TestBase {
...
-      "com.microsoft.ml.spark.train.ComputePerInstanceStatistics",
-      "com.microsoft.ml.spark.cognitive.LocalNER",
+      "com.microsoft.ml.spark.automl.TuneHyperparametersModel",
       "com.microsoft.ml.spark.cntk.train.CNTKLearner",
+      "com.microsoft.ml.spark.cognitive.LocalNER",
+      "com.microsoft.ml.spark.core.serialize.ComplexParamTest",
+      "com.microsoft.ml.spark.core.serialize.MixedParamTest",
+      "com.microsoft.ml.spark.core.serialize.StandardParamTest",
       "com.microsoft.ml.spark.core.serialize.TestEstimatorBase",
+      "com.microsoft.ml.spark.explainers.ImageLIME",
```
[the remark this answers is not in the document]
memoryz: Yes indeed. The SHAP sample coalition contains some randomness, and seems causes some small numerical difference in the coefficients and r2 metrics.

The LIME ones does not have this issue, so I removed them from the exemption list.

Let me look into the SHAP ones a bit more.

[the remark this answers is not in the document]
memoryz: I removed all these exemptions and enabled TransformerFuzzing for all of them.

src/main/scala/com/microsoft/ml/spark/explainers/BreezeUtils.scala
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.ml.spark.explainers
+
+import breeze.linalg.{DenseVector => BDV, SparseVector => BSV, Vector => BV, DenseMatrix => BDM}
+import org.apache.spark.ml.linalg.{Vector => SV, Vectors => SVS, Matrix => SM, Matrices => SMS}
```
imatiach-msft, line 7: these abbreviations are confusing, how about just adding Breeze to the breeze.linalg names, and keeping the ml linalg names as is?

src/main/scala/com/microsoft/ml/spark/explainers/ImageSHAP.scala
```diff
@@ -0,0 +1,121 @@
...
+        this.binarySamplesUdf
+      case t if ImageSchemaUtils.isImage(t) =>
+        this.imageSamplesUdf
+    }
+
+    df.withColumn("samples", explode(samplingUdf(col(getInputCol), col(getSuperpixelCol))))
+      .select(
+        col(idCol),
+        col("samples.coalition").alias(coalitionCol),
+        col("samples.sample").alias(getInputCol)
```
imatiach-msft, line 99: I see some of these strings in a few places, might be a good idea to put them in some constant file and reference from there

src/main/scala/com/microsoft/ml/spark/explainers/KernelSHAPBase.scala
```diff
@@ -0,0 +1,150 @@
...
+        val wlsResults = outputsBV(::, *).toIndexedSeq.map {
+          new LeastSquaresRegression().fit(inputsBV, _, weightsBV, fitIntercept = true)
+        }
+
+        val coefficientsMatrix = wlsResults.map(r => SVS.dense(r.intercept, r.coefficients.toArray: _*))
+        val metrics = BDV(wlsResults.map(_.rSquared): _*)
+        (id, coefficientsMatrix, metrics.toSpark)
+    }.toDF(idCol, this.getOutputCol, this.getMetricsCol)
+
+    preprocessed.hint("broadcast").join(fitted, Seq(idCol), "inner").drop(idCol)
```
imatiach-msft, line 92: is this the same as calling:
```
import org.apache.spark.sql.functions.broadcast
broadcast(preprocessed).join(fitted, Seq(idCol), "inner").drop(idCol)
```
https://spark.apache.org/docs/2.4.0/sql-performance-tuning.html

src/main/scala/com/microsoft/ml/spark/explainers/LassoRegression.scala
```diff
@@ -0,0 +1,80 @@
...
+    if (x > 0 && lambda < math.abs(x))
+      x - lambda
+    else if (x < 0 && lambda < math.abs(x))
+      x + lambda
+    else
+      0d
+  }
+
+  private def fitIteration(x: BDM[Double], y: BDV[Double])(currBeta: BDV[Double]): BDV[Double] = {
+    val newBeta = currBeta.copy
```
imatiach-msft, line 24: can't the data also be sparse, how do we handle sparse vectors?

src/main/scala/com/microsoft/ml/spark/explainers/SharedParams.scala
```diff
@@ -0,0 +1,193 @@
...
+
+  final def getTargetClassesCol: String = $(targetClassesCol)
+  final def setTargetClassesCol(value: String): this.type = this.set(targetClassesCol, value)
+
+  private def slice[T](values: Int => T, indices: Seq[Int])(num: Numeric[_]): SV = {
+    val n = num.asInstanceOf[Numeric[T]]
+    SVS.dense(indices.map(values.apply).map(n.toDouble).toArray)
+  }
+
+  private val dataTypeToNumericMap: Map[NumericType, Numeric[_]] = Map(
```
imatiach-msft, line 135: interesting utility, I wonder if we should put other core utilities like this in a more core-place in the repo (instead of under explainers), what do you think?

src/main/scala/com/microsoft/ml/spark/explainers/TabularLIME.scala
```diff
@@ -0,0 +1,153 @@
...
+      },
+      returnDataType
+    )
+
+    df.withColumn("samples", explode(samplesUdf(struct(getInputCols.map(col): _*))))
+      .select(
+        col(idCol),
+        col("samples.distance").alias(distanceCol),
+        col("samples.state").alias(stateCol),
+        col("samples.sample.*")
```
imatiach-msft, line 81: move to common constants file?

src/main/scala/com/microsoft/ml/spark/explainers/TabularSHAP.scala
```diff
@@ -0,0 +1,97 @@
...
+  override protected def createSamples(df: DataFrame, idCol: String, coalitionCol: String): DataFrame = {
+    val instances = df.select(col(idCol), struct(getInputCols.map(col): _*).alias("instance"))
+    val background = this.get(backgroundData).getOrElse(df)
+      .select(struct(getInputCols.map(col): _*).alias("background"))
+
+    val featureSize = this.getInputCols.length
+
+    val effectiveNumSamples = KernelSHAPBase.getEffectiveNumSamples(this.getNumSamplesOpt, featureSize)
+
+    // println(s"effectiveNumSamples: $effectiveNumSamples")
```
imatiach-msft, line 38: remove print?

src/main/scala/com/microsoft/ml/spark/explainers/VectorLIME.scala
```diff
@@ -0,0 +1,94 @@
...
+          val sampler = new LIMEVectorSampler(vector, featureStats)
+          (1 to numSamples).map(_ => sampler.sample)
+      },
+      returnDataType
+    )
+
+    df.withColumn("samples", explode(samplesUdf(col(getInputCol))))
+      .select(
+        col(idCol),
+        col("samples.distance").alias(distanceCol),
```
imatiach-msft, line 60: move to constants?

src/main/scala/org/apache/spark/ml/param/DataFrameParam.scala
```diff
@@ -109,7 +109,8 @@ class DataFrameParam(parent: Params, name: String, doc: String, isValid: DataFra
   override def assertEquality(v1: Any, v2: Any): Unit = {
     (v1, v2) match {
       case (df1: Dataset[_], df2: Dataset[_]) =>
-        assert(df1.toDF() === df2.toDF())
+        // assert(df1.toDF() === df2.toDF())
```
imatiach-msft, line 112: remove?

src/test/scala/com/microsoft/ml/spark/explainers/ImageExplainersSuite.scala
```diff
@@ -0,0 +1,159 @@
...
+  import spark.implicits._
+
+  test("ImageLIME can explain a model locally for image type observation") {
+    val (image, superpixels, weights, r2) = lime
+      .transform(imageDf)
+      .select("image", "superpixels", "weights", "r2")
+      .as[(ImageFormat, SuperpixelData, Seq[SV], SV)]
+      .head
+
+    // println(weights)
```
imatiach-msft, line 112: remove the printlns?

src/test/scala/com/microsoft/ml/spark/explainers/ImageExplainersSuite.scala
```diff
@@ -0,0 +1,159 @@
...
+    // import com.microsoft.ml.spark.io.image.ImageUtils
+    // import com.microsoft.ml.spark.lime.{Superpixel, SuperpixelData}
+    // import java.awt.image.BufferedImage
+    // val originalImage = ImageUtils.toBufferedImage(image.data, image.width, image.height, image.nChannels)
+    // val censoredImage: BufferedImage = Superpixel.maskImage(originalImage, superpixels, spStates)
+    // Superpixel.displayImage(censoredImage)
+    // Thread.sleep(100000)
+  }
+
+  // Do not run in CI pipeline due to high memory needs
```
imatiach-msft, line 128: it looks like this is run anyway?  it's not disabled?

src/test/scala/com/microsoft/ml/spark/explainers/SamplerSuite.scala
```diff
@@ -0,0 +1,316 @@
...
+    val featureStats = DiscreteFeatureStats(freqTable)
+
+    val (samples, distances) = (1 to 1000).map {
+      _ =>
+        val sample = featureStats.sample(3.0)
+        val distance = featureStats.getDistance(3.0, sample)
+        (sample, distance)
+    }.unzip
+
+//    println(samples(0 to 100))
```
imatiach-msft, line 58: remove the printlns

imatiach-msft (dismissed):
looks great, added some minor comments

imatiach-msft:
I didn't notice this in the code:

"For tabular and vector models, both explainers support passing in a background dataframe. If one is not given, the dataframe used for local interpretation will be used as background data."

but it seems like a bad idea to implicitly use the local interpretation data as background.  I didn't quite understand where this is done though, so maybe I am misreading/misunderstanding.

src/test/scala/com/microsoft/ml/spark/core/test/fuzzing/FuzzingTest.scala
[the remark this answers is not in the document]
mhamilton723: SG!

src/main/scala/com/microsoft/ml/spark/explainers/RegressionBase.scala
```diff
@@ -0,0 +1,108 @@
...
+    val normalizedWeights = normalizeSampleWeights(sampleWeights)
+
+    val (x, y, xOffset, yOffset) = if (fitIntercept) {
+      // step 1: center x and y.
+      val x_weighted = BDM.tabulate(data.rows, data.cols) {
+        (i, j) =>
+          normalizedWeights(i) * data(i, j)
+      }
+
+      val x_offset = (sum(x_weighted(::, *)) /:/ sum(normalizedWeights)).t
```
mhamilton723, line 40: nit: can we try to make these inner vars have a more idiomatic naming scheme?

src/main/scala/com/microsoft/ml/spark/explainers/BreezeUtils.scala
mhamilton723: I dont mind the brevity though ;)

src/main/scala/com/microsoft/ml/spark/explainers/ImageSHAP.scala
mhamilton723: Yes agree with this, someplace that crystalizes this new API in a SparkType might reduce duplication and clarify things

src/main/scala/com/microsoft/ml/spark/explainers/SamplerSupport.scala
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  override def nextState: SV = this.randomStateGenerator.next
+
+  private def generateSampleSizes(m: Int, nSamples: Int): List[Int] = {
+    assert(m > 0)
+    assert(nSamples >= 0)
+
+    def generateCoalitionLengths(n: Int, k: Int, b: Int): List[Int] = {
```
mhamilton723, line 64: I think this function can be replaced with only a few lines sampling from a particular categorical dist. Will talk more today

mhamilton723 (changes requested):

mhamilton723:
> I didn't notice this in the code:
> 
> "For tabular and vector models, both explainers support passing in a background dataframe. If one is not given, the dataframe used for local interpretation will be used as background data."
> 
> but it seems like a bad idea to implicitly use the local interpretation data as background. I didn't quite understand where this is done though, so maybe I am misreading/misunderstanding.

What is the default behavior of SHAP for these scenarios? Does it allow you to explain without background?

src/main/scala/com/microsoft/ml/spark/explainers/KernelSHAPBase.scala
memoryz: Spark 3 seems to recommend `df.hint("broadcast")` syntax:
https://spark.apache.org/docs/3.1.1/sql-performance-tuning.html#join-strategy-hints-for-sql-queries

`df.hint("broadcast")` creates an UnresolvedHint logical plan, while `broadcast(df)` creates a ResolvedHint logical plan. I cannot find any document regarding the difference between the two, so I'm sticking with what Spark 3 recommends.

memoryz:
> I didn't notice this in the code:
> 
> "For tabular and vector models, both explainers support passing in a background dataframe. If one is not given, the dataframe used for local interpretation will be used as background data."
> 
> but it seems like a bad idea to implicitly use the local interpretation data as background. I didn't quite understand where this is done though, so maybe I am misreading/misunderstanding.



> > I didn't notice this in the code:
> > "For tabular and vector models, both explainers support passing in a background dataframe. If one is not given, the dataframe used for local interpretation will be used as background data."
> > but it seems like a bad idea to implicitly use the local interpretation data as background. I didn't quite understand where this is done though, so maybe I am misreading/misunderstanding.
> 
> What is the default behavior of SHAP for these scenarios? Does it allow you to explain without background?

In TabularLIME, TabularSHAP, VectorLIME and VectorSHAP, it's done by:
```scala
val background = this.get(backgroundData).getOrElse(df)
```
`backgroundData` is a DataFrameParam. `this.get(backgroundData)` returns Option[DataFrame], and it's none if this param is not specified. `df` is the local interpretation data.

In the original SHAP kernel explainer code, it is assumed that the background data is specified. I did it this way because the old LIME explainer in MMLSpark sort of uses the local interpretation data as background data (used StandardScaler to find out mean and standard deviation), and I wanted to support this old behavior. If you don't think it's necessary, please let me know and I can remove it.

src/main/scala/com/microsoft/ml/spark/explainers/BreezeUtils.scala
memoryz: ok so I removed all the aliasing for spark.ml.linalg vector classes, but kept the BDV/BSV aliasing for breeze vectors. I also like the brevity especially when the DenseVector class is used a lot.

src/main/scala/com/microsoft/ml/spark/explainers/ImageSHAP.scala
memoryz: done.

src/main/scala/com/microsoft/ml/spark/explainers/LassoRegression.scala
memoryz: I don't know if there is an algorithm that optimizes for sparse vectors. For now, I changed the interface inside `RegressionBase` (that's the public facing API) to support the more general Vector type, and just convert the vector inputs to dense type, so at least functional-wise, we can handle sparse vectors.

src/main/scala/com/microsoft/ml/spark/explainers/TabularLIME.scala
memoryz: done.

src/main/scala/com/microsoft/ml/spark/explainers/TabularSHAP.scala
memoryz: done.

src/main/scala/com/microsoft/ml/spark/explainers/VectorLIME.scala
memoryz: done.

src/main/scala/org/apache/spark/ml/param/DataFrameParam.scala
memoryz: done.

src/test/scala/com/microsoft/ml/spark/explainers/ImageExplainersSuite.scala
memoryz: done.

src/test/scala/com/microsoft/ml/spark/explainers/ImageExplainersSuite.scala
memoryz: Actually it shouldn't be disabled. I removed the comment.

src/test/scala/com/microsoft/ml/spark/explainers/SamplerSuite.scala
memoryz: done.

src/main/scala/com/microsoft/ml/spark/explainers/RegressionBase.scala
memoryz: done.

imatiach-msft:
@memoryz I think this behavior is a bit misleading.  If a user forgets to specify background data, we should error out, not use the evaluation examples.  Using evaluation examples as background seems like a rare edge case to me.  It's much more likely the user somehow forgot or didn't understand what it means to supply background data.  What do you think?  Maybe for LIME I can see how it could make sense, but for SHAP KernelExplainer I see this as confusing.

src/main/scala/com/microsoft/ml/spark/explainers/LassoRegression.scala
imatiach-msft: @memoryz yes, we should be able to support sparse vector, please see PR:
https://github.com/slundberg/shap/pull/200
lasso regression should support sparse data, eg see:
https://scikit-learn.org/stable/auto_examples/linear_model/plot_lasso_dense_vs_sparse_data.html

we can discuss this more and improve in separate PR if you want

memoryz:
> @memoryz I think this behavior is a bit misleading. If a user forgets to specify background data, we should error out, not use the evaluation examples. Using evaluation examples as background seems like a rare edge case to me. It's much more likely the user somehow forgot or didn't understand what it means to supply background data. What do you think? Maybe for LIME I can see how it could make sense, but for SHAP KernelExplainer I see this as confusing.

I'll modify the code. Thank you for the comment!

src/main/scala/com/microsoft/ml/spark/explainers/SharedParams.scala
memoryz: done. Extracted to `com.microsoft.ml.spark.core.utils.SlicerFunctions` and added unit tests.

src/main/scala/com/microsoft/ml/spark/explainers/LassoRegression.scala
memoryz: Agree. Let's improve that in a separate PR.

src/main/scala/com/microsoft/ml/spark/explainers/SamplerSupport.scala
memoryz: I rewrote the shap sampler to align with the original shap implementation.

src/main/scala/com/microsoft/ml/spark/core/utils/RowUtils.scala
```diff
@@ -6,10 +6,13 @@ package com.microsoft.ml.spark.core.utils
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.catalyst.expressions.GenericRow
 
+// This class currently has no usage. Should we just remove it?
+@deprecated("This is a copy of Row.merge function from Spark, which was marked deprecated.", "1.0.0-rc3")
```
mhamilton723, line 10: yes we can remove

mhamilton723 (approved):

PR #1086, opened 2021-06-15
update: update spark version to 3.1.2

xuwq1993:
Update Spark Version to 3.1.2 (Synapse Spark3 GA version)


src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitContextualBandit.scala
```diff
@@ -275,7 +275,7 @@ class VowpalWabbitContextualBandit(override val uid: String)
     })
   }
 
-  override def fit(dataset: Dataset[_], paramMaps: Array[ParamMap]): Seq[VowpalWabbitContextualBanditModel] = {
```
eisber, line 278: why is the override removed?

eisber (approved):

src/main/scala/com/microsoft/ml/spark/vw/VowpalWabbitContextualBandit.scala
xuwq1993: Spark ML Lib changed Array[ParamMap] to Seq[ParamMap]
Update: keep "override", changed Array[] to Seq[]

mhamilton723 (approved):

PR #1108, opened 2021-06-30
feat: add translator

serena-ruan:
Add translator into mmlspark


cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/DocumentTranslator.scala
```diff
@@ -0,0 +1,172 @@
...
+            sourceUrl = getValue(r, sourceUrl)),
+            storageType = getValueOpt(r, storageType),
+            targets = getValue(r, targets).asInstanceOf[Seq[Row]].map(
+              row => TargetInput(Option(row.getString(0)),
+                fetchGlossaries(row),
+                row.getString(2), row.getString(3), Option(row.getString(4))))
+          ))).toJson.compactPrint, ContentType.APPLICATION_JSON))
+  }
+
+  private def queryForResult(key: Option[String],
```
mhamilton723, line 119: Are there any similarities that would allow us to abstract this and other async querying logic into the same function?

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
```diff
@@ -0,0 +1,406 @@
...
+        getValueOpt(r, text)
+          .map(x => x.map(y => Map("Text" -> y))).toJson.compactPrint, ContentType.APPLICATION_JSON))
+  }
+}
+
+abstract class TextTranslatorBase(override val uid: String) extends CognitiveServicesBase(uid)
+  with HasInternalJsonOutputParser with HasCognitiveServiceInput with HasSubscriptionRegion
+  with HasSetLocation {
+
+  protected val subscriptionRegionHeaderName = "Ocp-Apim-Subscription-Region"
```
mhamilton723, line 89: this is so strange i cant believe they make you specify this in a header lol

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
```diff
@@ -0,0 +1,406 @@
...
+    " For example, use to=de to translate to German. It's possible to translate to multiple languages simultaneously" +
+    " by repeating the parameter in the query string. For example, use to=de&to=it to translate to German and Italian.",
+    isRequired = true, isURLParam = true,
+    toValueString = { seq => seq.mkString(",") })
+
+  def setToLanguage(v: Seq[String]): this.type = setScalarParam(toLanguage, v)
+
+  def setToLanguageCol(v: String): this.type = setVectorParam(toLanguage, v)
+
+  val fromLanguage = new ServiceParam[String](this, "fromLanguage", "Specifies the language of the input" +
```
mhamilton723, line 176: Can abstract the fromLanguage and toLanguage methods into separate traits and "mix together" to reduce code. Also be on the lookout for other params that can be "factored" out in this manner as its less maintenance work for us later on ;)

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,434 @@
...
+
+  lazy val textDf5: DataFrame = Seq(List("The word <mstrans:dictionary translation=wordomatic>word " +
+    "or phrase</mstrans:dictionary> is a dictionary entry.")).toDF("text")
+
+}
+
+class TranslateSuite extends TransformerFuzzing[Translate]
+  with TranslatorKey with Flaky with TranslatorUtils {
+
+  lazy val translate: Translate = new Translate()
```
mhamilton723, line 42: Can factor out common setters into a base method then just .set only the differing params. Isnt the fluent API nifty!

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,434 @@
...
+
+trait TranslatorUtils extends TestBase {
+
+  import spark.implicits._
+
+  lazy val textDf1: DataFrame = Seq(List("Hello, what is your name?")).toDF("text")
+
+  lazy val textDf2: DataFrame = Seq(List("Hello, what is your name?", "Bye")).toDF("text")
+
+  lazy val textDf3: DataFrame = Seq(List("This is bullshit.")).toDF("text")
```
mhamilton723, line 29: lol

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,434 @@
...
+    .setSubscriptionKey(translatorKey)
+    .setLocation("eastus")
+    .setFromLanguage("en")
+    .setToLanguage(Seq("fr"))
+    .setIncludeSentenceLength(true)
+    .setTextCol("text")
+    .setOutputCol("translation")
+    .setConcurrency(5)
+
+  test("Translate multiple pieces of text with language autodetection") {
```
mhamilton723, line 79: all of these tests have a similiar structure, might want to factor this structure out so that the tests are smaller to write. I know it's just test code and who cares, but it will make your life easier when you need to update I promise

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,434 @@
...
+      .transform(textDf1)
+      .withColumn("translation", flatten(col("translation.translations")))
+      .withColumn("translation", col("translation.text"))
+      .select("translation").collect()
+    val headStr = results.head.getSeq(0).mkString("\n")
+    assert(headStr === "你好，你叫什么名字？\nHallo, wie heißt du?")
+  }
+
+  test("Handle profanity") {
+    val results = translate
```
mhamilton723, line 111: when re-using the same estimator make translate a def not a lazy val. I know its weird but the setters actually MODIFY state globally which is a weird thing on SparkML's part

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,434 @@
...
+  override def reader: MLReadable[_] = DictionaryExamples
+}
+
+class DocumentTranslatorSuite extends TransformerFuzzing[DocumentTranslator]
+  with TranslatorKey with Flaky {
+
+  import spark.implicits._
+
+  // TODO: Replace all of those SAS urls after 2022-07-07
+  lazy val sourceUrl: String = "https://mmlspark.blob.core.windows.net/datasets?sp=rl&st=2021-07-06T06" +
```
mhamilton723, line 320: are you able to get away without SAS URLs? the datasets blob is public I believe. Also you might want to pull the root URL into a val to keep code DRY

core/src/test/scala/com/microsoft/ml/spark/Secrets.scala
```diff
@@ -48,6 +48,8 @@ object Secrets {
   lazy val AnomalyApiKey: String = getSecret("anomaly-api-key")
   lazy val AzureSearchKey: String = getSecret("azure-search-key")
   lazy val BingSearchKey: String = getSecret("bing-search-key")
+  lazy val TranslatorKey: String = getSecret("translator-key")
+  lazy val TranslatorName: String = getSecret("translator-name")
```
mhamilton723, line 52: What does the translator name do? is it state or a secret?

mhamilton723 (changes requested):

mhamilton723:
Awesome stuff, so great to see this flying out of your fingertips!

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
serena-ruan: Yes it's quite strange... the document says it's optional for global translator resource, but if we don't add it into header the response will be '{"error":{"code":401000,"message":"The request is not authorized because credentials are missing or invalid."}}'

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
serena-ruan: Emmm I tried but seems can't access it directly using '"https://mmlspark.blob.core.windows.net/datasets", it returns "error": {
        "code": "InvalidRequest",
        "message": "Cannot access source document location with the current permissions.",
        "target": "Operation",
        "innerError": {
            "code": "InvalidDocumentAccessLevel",
            "message": "Cannot access source document location with the current permissions."
        }
    },

core/src/test/scala/com/microsoft/ml/spark/Secrets.scala
serena-ruan: It's the service name 
image

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/DocumentTranslator.scala
serena-ruan: Done.

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
serena-ruan: Done.

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
serena-ruan: Done.

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
serena-ruan: Done. That solves my pain on creating multiple translators lol!

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
serena-ruan: Although there're several fromLanguage & toLanguage in this file, the 'isRequired' parameter and type might be different, so didn't factor these two within Translate out. Or do you have other approaches to mix them together?

mhamilton723: Okay! One idea would be to add protected fromLanguageRequired: Boolean = true on base class and just override that

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
mhamilton723: Arent cognitive services so great!

core/src/test/scala/com/microsoft/ml/spark/Secrets.scala
mhamilton723: Weird

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
mhamilton723: On folders it might not work, but it should work on files. Is this manually inspecting a folder and loading content from it, or is it taking it URL by URL?

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,405 @@
...
+}
+
+class TransliterateSuite extends TransformerFuzzing[Transliterate]
+  with TranslatorKey with Flaky with TranslatorUtils {
+
+  import spark.implicits._
+
+  lazy val transDf: DataFrame = Seq(List("こんにちは", "さようなら")).toDF("text")
+
+  lazy val transliterate: Transliterate = new Transliterate()
```
mhamilton723, line 161: can you change lazy val -> def here and elsewhere?

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,405 @@
...
+  override def testObjects(): Seq[TestObject[Transliterate]] =
+    Seq(new TestObject(transliterate, transDf))
+
+  override def reader: MLReadable[_] = Transliterate
+}
+
+class DetectSuite extends TransformerFuzzing[Detect]
+  with TranslatorKey with Flaky with TranslatorUtils {
+
+  lazy val detect: Detect = new Detect()
```
mhamilton723, line 188: lazy val -> def

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,405 @@
...
+
+  lazy val targetRoot: String = "?sp=racwl&st=2021-07-06T06:29:05Z&se=2022-07-07T06:29:00Z" +
+    "&sv=2020-08-04&sr=c&sig=tk62GpHoRb5CcojmyQammMbnYICAsTdgQqAeCikbtKg%3D"
+
+  lazy val sourceUrl: String = "https://mmlspark.blob.core.windows.net/datasets" + sourceRoot
+
+  lazy val fileSourceUrl: String = "https://mmlspark.blob.core.windows.net/datasets/Translator/" +
+    "source/document-translation-sample.pdf" + sourceRoot
+
+  lazy val targetUrl: String = "https://mmlspark.blob.core.windows.net/translator-target/test-zh-Hans-" +
```
mhamilton723, line 309: It looks like this is a new container as opposed to using the datasets container, are we unable to re-use datasets container for everything?

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,405 @@
...
+  override def reader: MLReadable[_] = DictionaryExamples
+}
+
+class DocumentTranslatorSuite extends TransformerFuzzing[DocumentTranslator]
+  with TranslatorKey with Flaky {
+
+  import spark.implicits._
+
+  // TODO: Replace root SAS urls after 2022-07-07
+  lazy val sourceRoot: String = "?sp=rl&st=2021-07-06T06:28:26Z&se=2022-07-07T06:28:00Z" +
```
mhamilton723, line 298: might want to consider factor this in the following way

lazy val containerSasToken = ... (If this needs to be a SAS otherwise we should try to remove SAS)
lazy val urlRoot = "https://mmlspark.blob.core.windows.net/"

and use the same container as for all experiments. If you find the container SAS to be used all over we might consider breaking this into a single trait for all cog service tests so that we only need to update in one location if it expires

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -0,0 +1,405 @@
...
+    ))), targetUrl3, "zh-Hans", None))))
+    .toDF("sourceUrl", "filterPrefix", "targets")
+
+  lazy val docTranslationDf4: DataFrame = Seq((fileSourceUrl,
+    "File",
+    Seq(TargetInput(None, None, targetFileUrl1, "zh-Hans", None),
+      TargetInput(None, None, targetFileUrl2, "de", None))))
+    .toDF("sourceUrl", "storageType", "targets")
+
+  lazy val documentTranslator: DocumentTranslator = new DocumentTranslator()
```
mhamilton723, line 350: lazy val -> def and factor out shared structure

mhamilton723 (changes requested):
Thanks for making these changes have a few more ideas for how to continue tidying, awesome work and appreciate the iterations on this :)

mhamilton723 (approved):

PR #1110, opened 2021-06-30
Func: Basic multi-language capabilities (without mapping)

preetipidatala:
This is the multi language base code which creates the language column trait to be used by all endpoints. This version passes in some secondary input to the endpoint (ex. language, country hint) one at a time.

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
+++ b/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
@@ -1,6 +1,7 @@
 package com.microsoft.ml.spark.cognitive
 
-import com.azure.ai.textanalytics.models.{ExtractKeyPhraseResult, KeyPhrasesCollection, TextAnalyticsRequestOptions, TextAnalyticsWarning}
+import com.azure.ai.textanalytics.models.{ExtractKeyPhraseResult, KeyPhrasesCollection, TextAnalyticsRequestOptions,
+  TextAnalyticsWarning}
 import com.azure.ai.textanalytics.{TextAnalyticsClient, TextAnalyticsClientBuilder}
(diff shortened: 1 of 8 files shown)


src/test/scala/com/microsoft/ml/spark/cognitive/split1/TextAnalyticsSuite.scala
```diff
@@ -324,7 +324,8 @@ class NERSuite extends TransformerFuzzing[NERV2] with TextKey {
 
   lazy val df: DataFrame = Seq(
     ("1", "en", "Jeff bought three dozen eggs because there was a 50% discount."),
-    ("2", "en", "The Great Depression began in 1929. By 1933, the GDP in America fell by 25%.")
+    ("2", "en", "In 1929, The Great Depression began. By 1933, the GDP in America fell by 25%."),
```
HanLiMS, line 327: this is not needed. remove this file from your changes.

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
```diff
@@ -1,34 +1,29 @@
...
-import com.azure.ai.textanalytics.models
-
-import java.net.URI
 import scala.collection.JavaConverters._
 
 abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnalyticsRequestOptions] = None)
   extends Transformer
-  with HasInputCol with HasErrorCol
-  with HasEndpoint with HasSubscriptionKey
+  with HasInputCol with HasOutputCol with HasErrorCol
```
HanLiMS, line 22: Great. Could you work on the example https://github.com/HanLiMS/mmlspark/tree/hali2/intern/callfunc to update the HasOutCol etc  & add the multi lang support? Let's work on directly with calling the function.

PR #1125, opened 2021-07-16
feat: fixed mini batching added

skonigs:
Added FixedMiniBatching to the 3 endpoints

### DIFF
--- a/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
+++ b/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
@@ -1,5 +1,6 @@
 package com.microsoft.ml.spark.cognitive
-import com.azure.ai.textanalytics.models.{AssessmentSentiment, DetectLanguageInput, DocumentSentiment, SentenceSentiment, SentimentConfidenceScores, TargetSentiment, TextDocumentInput}
+import com.azure.ai.textanalytics.models.{AssessmentSentiment, DetectLanguageInput,
+  DocumentSentiment, SentenceSentiment, SentimentConfidenceScores, TargetSentiment, TextDocumentInput}
 import com.azure.ai.textanalytics.{TextAnalyticsClient, TextAnalyticsClientBuilder}
 import com.azure.core.credential.AzureKeyCredential
 import com.microsoft.ml.spark.core.contracts.{HasConfidenceScoreCol, HasInputCol, HasLangCol, HasOutputCol, HasTextCol}
--- a/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
+++ b/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
@@ -1,26 +1,22 @@
 package com.microsoft.ml.spark.cognitive
-import com.azure.ai.textanalytics.models.{AssessmentSentiment, DetectLanguageInput, DocumentSentiment, SentenceSentiment, SentimentConfidenceScores, TargetSentiment, TextDocumentInput}
+import com.azure.ai.textanalytics.models.{AssessmentSentiment, DetectLanguageInput,
+  DocumentSentiment, SentenceSentiment, SentimentConfidenceScores, TargetSentiment, TextDocumentInput}
 import com.azure.ai.textanalytics.{TextAnalyticsClient, TextAnalyticsClientBuilder}
 import com.azure.core.credential.AzureKeyCredential
 import com.microsoft.ml.spark.core.contracts.{HasConfidenceScoreCol, HasInputCol, HasLangCol, HasOutputCol, HasTextCol}
 import com.microsoft.ml.spark.core.schema.SparkBindings
 import com.microsoft.ml.spark.io.http.HasErrorCol
 import com.microsoft.ml.spark.logging.BasicLogging
-import org.apache.spark.ml.param.{BooleanParam, IntParam, Param, ParamMap, Params}
+import org.apache.spark.ml.param.ParamMap
 import org.apache.spark.ml.util.Identifiable._
 import org.apache.spark.ml.{ComplexParamsReadable, ComplexParamsWritable, Transformer}
-import org.apache.spark.sql.types.{ArrayType, DataTypes, StringType, StructType}
-import org.apache.spark.sql.{DataFrame, Dataset, Row}
+import org.apache.spark.sql.types.{DataTypes, StructType}
+import org.apache.spark.sql.{DataFrame, Dataset, Row, SparkSession}
 import com.azure.core.util.Context
-import com.microsoft.ml.spark.stages.{FixedBatcher, FixedBufferedBatcher, FixedMiniBatchTransformer, HasBatchSize, MiniBatchBase}
-import com.sun.tools.javac.code.TypeTag
-import org.apache.spark.injections.UDFUtils
-import org.apache.spark.ml.util.{DefaultParamsReadable, Identifiable}
+import com.microsoft.ml.spark.stages.FixedMiniBatchTransformer
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
 
-import java.util.concurrent.{BlockingQueue, CountDownLatch, LinkedBlockingQueue}
 import scala.collection.JavaConverters._
-import scala.reflect.runtime.universe.{MethodSymbol, typeOf}
 
 abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnalyticsRequestOptionsV4] = None)
   extends Transformer
@@ -33,7 +29,11 @@ abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnal
   protected def outputSchema: StructType
 
   val responseTypeBinding: SparkBindings[TAResponseV4[T]]
-
+  val spark = SparkSession
+    .builder
+    .appName("SparkSQL")
+    .master("local[*]")
+    .getOrCreate()
   def invokeTextAnalytics(text: Seq[String], lang: Seq[String]): TAResponseV4[T]
 
   protected lazy val textAnalyticsClient: TextAnalyticsClient =
@@ -41,6 +41,11 @@ abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnal
       .credential(new AzureKeyCredential(getSubscriptionKey))
       .endpoint(getEndpoint)
       .buildClient()
+
+  import spark.implicits._
+  lazy val df: DataFrame = Seq(
+    (Seq(""),Seq("")),
+  ).toDF("lang", "text")
 
   protected def transformTextRows(toRow: TAResponseV4[T] => Row)
                                  (rows: Iterator[Row]): Iterator[Row] = {
@@ -51,8 +56,10 @@ abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnal
     }}
 
   override def transform(dataset: Dataset[_]): DataFrame = {
+    val batchedDF = new FixedMiniBatchTransformer().setBatchSize(9).transform(dataset.coalesce(1))
+    val finaldataset = spark.createDataFrame(batchedDF.rdd, df.schema)
     logTransform[DataFrame]({
-      val df = dataset.toDF
+      val df = finaldataset.toDF
       val enc = RowEncoder(df.schema.add(getOutputCol, responseTypeBinding.schema))
       val toRow = responseTypeBinding.makeToRowConverter
       df.mapPartitions(transformTextRows(
--- a/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
+++ b/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
@@ -1,6 +1,5 @@
 package com.microsoft.ml.spark.cognitive
-import com.azure.ai.textanalytics.models.{AssessmentSentiment, DetectLanguageInput,
-  DocumentSentiment, SentenceSentiment, SentimentConfidenceScores, TargetSentiment, TextDocumentInput}
+import com.azure.ai.textanalytics.models.{AssessmentSentiment, DetectLanguageInput, DocumentSentiment, SentenceSentiment, SentimentConfidenceScores, TargetSentiment, TextDocumentInput}
 import com.azure.ai.textanalytics.{TextAnalyticsClient, TextAnalyticsClientBuilder}
 import com.azure.core.credential.AzureKeyCredential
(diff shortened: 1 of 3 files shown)


src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
```diff
@@ -1,20 +1,26 @@
...
 import com.azure.core.util.Context
+import com.microsoft.ml.spark.stages.{FixedBatcher, FixedBufferedBatcher, FixedMiniBatchTransformer, HasBatchSize, MiniBatchBase}
+import com.sun.tools.javac.code.TypeTag
+import org.apache.spark.injections.UDFUtils
+import org.apache.spark.ml.util.{DefaultParamsReadable, Identifiable}
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
+
+import java.util.concurrent.{BlockingQueue, CountDownLatch, LinkedBlockingQueue}
 import scala.collection.JavaConverters._
+import scala.reflect.runtime.universe.{MethodSymbol, typeOf}
```
assafi, line 23: Is this needed?

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
```diff
@@ -1,20 +1,26 @@
...
 import com.microsoft.ml.spark.logging.BasicLogging
-import org.apache.spark.ml.param.{ParamMap}
+import org.apache.spark.ml.param.{BooleanParam, IntParam, Param, ParamMap, Params}
 import org.apache.spark.ml.util.Identifiable._
 import org.apache.spark.ml.{ComplexParamsReadable, ComplexParamsWritable, Transformer}
-import org.apache.spark.sql.types.{DataTypes, StructType}
+import org.apache.spark.sql.types.{ArrayType, DataTypes, StringType, StructType}
 import org.apache.spark.sql.{DataFrame, Dataset, Row}
 import com.azure.core.util.Context
+import com.microsoft.ml.spark.stages.{FixedBatcher, FixedBufferedBatcher, FixedMiniBatchTransformer, HasBatchSize, MiniBatchBase}
```
assafi, line 15: There's a lot of new imports but I don't see new corresponding code.

src/test/scala/com/microsoft/ml/spark/cognitive/split1/TextAnalyticsSDKSuite.scala
```diff
@@ -103,6 +135,31 @@ class TextSentimentSuiteV4 extends TestBase with DataFrameEquality with TextKey
...
+
+  test("Sentiment - batch usage"){
+    lazy val detector3: TextSentimentV4  = new TextSentimentV4(options)
+      .setSubscriptionKey(textKey)
+      .setEndpoint("https://eastus.api.cognitive.microsoft.com/")
+      .setInputCol("text")
+      .setLangCol("lang")
+      .setOutputCol("output2")
+
+    val batchedDF2 = new FixedMiniBatchTransformer().setBatchSize(10).transform(df2.coalesce(1))
```
assafi, line 154: I'm confused here. Shouldn't the change be in the TextSentimentV4 class, and not outside of it.
Looks like this PR just show-case how a customer might use the `FixedMiniBatchTransformer` to do batching but doesn't change the connector to do it instead.

Shouldn't `TextAnalyticsSDKBase` inherit or use `FixedMiniBatchTransformer` somehow instead?

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
skonigs: Ooops forgot to take those out. Were for a different attempt at batching that ended up not being used.

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
```diff
@@ -36,35 +42,39 @@ abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnal
...
                                  (rows: Iterator[Row]): Iterator[Row] = {
+
     rows.map { row =>
       val results = invokeTextAnalytics(getValue(row, text), getValue(row,lang))
       Row.fromSeq(row.toSeq ++ Seq(toRow(results))) // Adding a new column
     }}
 
   override def transform(dataset: Dataset[_]): DataFrame = {
+    val batchedDF = new FixedMiniBatchTransformer().setBatchSize(10).transform(dataset.coalesce(1))
+    val finaldataset = spark.createDataFrame(batchedDF.rdd, df.schema)
```
HanLiMS, line 60: move this in the logTransform[DataFrame]

HanLiMS, line 60: I would suggest instead of creating a temp df here, creating the schema directly: 

https://sparkbyexamples.com/spark/spark-schema-explained-with-examples/

  val inputSchema = StructType(Array(
    StructField("lang",ArrayType(StringType,true),true),
    StructField("text",ArrayType(StringType,true),true)
  ))

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
```diff
@@ -36,35 +42,39 @@ abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnal
...
   protected def transformTextRows(toRow: TAResponseV4[T] => Row)
                                  (rows: Iterator[Row]): Iterator[Row] = {
+
     rows.map { row =>
       val results = invokeTextAnalytics(getValue(row, text), getValue(row,lang))
       Row.fromSeq(row.toSeq ++ Seq(toRow(results))) // Adding a new column
     }}
 
   override def transform(dataset: Dataset[_]): DataFrame = {
+    val batchedDF = new FixedMiniBatchTransformer().setBatchSize(10).transform(dataset.coalesce(1))
```
HanLiMS, line 59: This number 10 --> pass in as a configuration number && set up a default number as 5 or 10.

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
```diff
@@ -44,23 +45,28 @@ abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnal
...
+      }(ExecutionContext.global)
+    }
+    bufferedAwait(futures, getConcurrency, Duration(getTimeout,SECONDS))(ExecutionContext.global)
+  }
 
   override def transform(dataset: Dataset[_]): DataFrame = {
-    val batchedDF = new FixedMiniBatchTransformer().setBatchSize(10).transform(dataset.coalesce(1))
-    val finaldataset = spark.createDataFrame(batchedDF.rdd, df.schema)
     logTransform[DataFrame]({
+      val batchedDF = new FixedMiniBatchTransformer().setBatchSize(getBatchSize).transform(dataset.coalesce(1))
```
assafi, line 68: Do we really need the `coalesce(1)`?

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
```diff
@@ -44,23 +45,28 @@ abstract class TextAnalyticsSDKBase[T](val textAnalyticsOptions: Option[TextAnal
...
+      val finaldataset = spark.createDataFrame(batchedDF.rdd, inputSchema)
```
assafi, line 69: I'm not sure I understand how this works. Is this how the `FixedMiniBatchTransformer` is used elsewhere? Is there a pattern you're following?

src/test/scala/com/microsoft/ml/spark/cognitive/split1/TextAnalyticsSDKSuite.scala
```diff
@@ -74,11 +77,55 @@ class DetectedLanguageSuitev4 extends TestBase with DataFrameEquality with TextK
...
     val language = tdf.map(row => row.getList(0))
     assert(language(0).get(0).toString == "English" && language(0).get(1).toString == "Spanish")
-    assert(language(0).get(2).toString == "French" && language(0).get(3).toString == "Chinese")
+    assert(language(1).get(0).toString == "French" && language(1).get(1).toString == "Chinese")
 
     val iso = tdf.map(row => row.getList(1))
     assert(iso(0).get(0).toString == "en" && iso(0).get(1).toString == "es" &&
-      iso(0).get(2).toString == "fr" && iso(0).get(3).toString == "zh")
+      iso(1).get(0).toString == "fr" && iso(1).get(1).toString == "zh")
+  test("Asynch Functionality with Parameters") {
```
assafi, line 85: nit: Async not Asynch

assafi, line 85: Could you rename this test? The current name is very generic.

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
skonigs: This is how it is used in past scenarios (TA.scala). The only differences is our schema is structured slightly differently so we have to convert it back to our schema after calling fixedminibatch for us to get the proper output

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
skonigs: fails without it --> this is how mark called it in TASuite

src/test/scala/com/microsoft/ml/spark/cognitive/split1/TextAnalyticsSDKSuite.scala
```diff
@@ -150,6 +181,32 @@ class TextSentimentSuiteV4 extends TestBase with DataFrameEquality with TextKey
...
+    ("en", "Hello world. This is some input text that I love."),
+    ("fr", "Bonjour tout le monde"),
+    ("es", "La carretera estaba atascada. Había mucho tráfico el día de ayer."),
+    ("es", "La carretera estaba atascada."),
+    ("es", "Había mucho tráfico el día de ayer."),
+    ("es", "La carretera estaba atascada. Había mucho tráfico el día de ayer."),
+    (null, "ich bin ein berliner")
+  ).toDF("lang", "text")
+
+  test("Sentiment - batch usage"){
```
HanLiMS, line 195: what is the purpose of this testing case here? if we would like to test the current default batch setup, then we might want to add some assertions here to show it actually batched according to the default value? Let's say we have 5 as default, then the input is 7, the batched result will be two rows of 5 first and 2 second? something like this?

src/test/scala/com/microsoft/ml/spark/cognitive/split1/TextAnalyticsSDKSuite.scala
```diff
@@ -194,4 +251,27 @@ class KeyPhraseExtractionSuiteV4 extends TestBase with DataFrameEquality with Te
...
+    ("fr", "Bonjour tout le monde")
+  ).toDF("lang","text" )
+
+  test("Keyphrase - batch usage"){
+    lazy val extractor2: TextAnalyticsKeyphraseExtraction = new TextAnalyticsKeyphraseExtraction(options)
+      .setSubscriptionKey(textKey)
+      .setEndpoint("https://eastus.api.cognitive.microsoft.com/")
+      .setInputCol("text")
+      .setLangCol("lang")
+      .setBatchSize(8)
```
HanLiMS, line 267: if the input contains only 3 documents and the batch size is set as 8,  there is no difference comparing with the default setting of 5? So the result will look the same.

src/test/scala/com/microsoft/ml/spark/cognitive/split1/TextAnalyticsSDKSuite.scala
```diff
@@ -150,6 +181,33 @@ class TextSentimentSuiteV4 extends TestBase with DataFrameEquality with TextKey
...
+    (null, "ich bin ein berliner")
+  ).toDF("lang", "text")
+
+  test("Sentiment - batch usage"){
+    lazy val detector3: TextSentimentV4  = new TextSentimentV4(options)
+      .setSubscriptionKey(textKey)
+      .setEndpoint("https://eastus.api.cognitive.microsoft.com/")
+      .setInputCol("text")
+      .setLangCol("lang")
+      .setBatchSize(2)
```
HanLiMS, line 201: if we set batchsize as 2 here --> then ideally each row should only contain 2 input documents/sentences. So the test here should check --> whether the output is splitted into each row containing 2 documents with results.   You can check the details in val tdf = detector3.transform(df2)
      .select("output2.result.sentiment") --> without calling.collect().   Try to .show to see the details.

src/test/scala/com/microsoft/ml/spark/cognitive/split1/TextAnalyticsSDKSuite.scala
```diff
@@ -194,4 +252,25 @@ class KeyPhraseExtractionSuiteV4 extends TestBase with DataFrameEquality with Te
...
+      .setSubscriptionKey(textKey)
+      .setEndpoint("https://eastus.api.cognitive.microsoft.com/")
+      .setInputCol("text")
+      .setLangCol("lang")
+      .setOutputCol("output2")
+
+    val tdf = extractor2.transform(df3)
+      .select("output2.result.keyPhrases")
+      .collect()
+    assert(tdf.length == 1)
```
HanLiMS, line 273: What is this assertion checking here?

skonigs: it's checking that it is only batching into one row (expected because default is 5 and this is 3 inputs)

src/test/scala/com/microsoft/ml/spark/cognitive/split1/TextAnalyticsSDKSuite.scala
skonigs: Got it! Will change that. was checking by asserting length to 4 meaning first 3 batches would have 2 and last would have 1 amounting in 4 rows

skonigs: output is correct with .show() will push changes now

src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalyticsSDK.scala
```diff
@@ -1,289 +1,301 @@
...
-
-    TAResponseV4[SentimentScoredDocumentV4](
-      sentimentResult,
-      error,
-      stats,
-      Some(resultCollection.getModelVersion))
-  }
-  override def outputSchema: StructType = SentimentResponseV4.schema
-}
+package com.microsoft.ml.spark.cognitive
```
HanLiMS, line 1: Can we make sure the previous codes still present? Only submit the changes you modified?

PR #1132, opened 2021-07-20
fix: test case asserting confidence score, sentiment score and model version format

victoriajmicrosoft:


victoriajmicrosoft:
Created a PR on the wrong branch.

PR #1138, opened 2021-07-26
Fix SpeechToTextSDK setLinedService Error

xuwq1993:
Error:

SpeechToText didn't support endpoint like: https://wenqxcogserv.cognitiveservices.azure.com/

It only support https://{location}.api,cognitive.microsoft.com/ like: https://eastus.api.cognitive.microsoft.com/

Also, SpeechToText don't support DEP Vnet, therefore, we don't need to consider the PE problems.


mhamilton723 (approved):

mhamilton723 (changes requested):
can we make setLinkedService a trait and mix in in both places?

mhamilton723 (approved):

PR #1154, opened 2021-08-05
Documentation on an example notebook for PII analysis

preetipidatala:
Edits the cognitive services docs to include healthcare and PII endpoints, and links to an example notebook for PII redaction within synapse.


mhamilton723:
- You cannot commit keys within notebooks 
- Consider adding this to the Cognitive Service Overview notebook
- no need to put `.setUrl("https://eastus.api.cognitive.microsoft.com/")` as that is already taken care of by set location
- consider making just a short sample dataframe manually so we can remove any dependence on a file. This also has the added benefit of showing what the input data is. theres a few examples in the cogntive service overview doc you can look at

PR #1156, opened 2021-08-09
feat: .NET binding support

serena-ruan:
AB#1821715


core/src/main/scala/org/apache/spark/ml/param/PythonWrappableParam.scala
```diff
@@ -41,7 +41,7 @@ trait PythonWrappableParam[T] extends Param[T] {
 
   val name: String
 
-  type InnerType = T
+  type pyInnerType = T
```
mhamilton723, line 44: Will we need separate values per language?

core/src/main/scala/org/apache/spark/ml/param/UDFParam.scala
```diff
@@ -30,4 +30,17 @@ class UDFParam(parent: Params, name: String, doc: String, isValid: UserDefinedFu
         throw new AssertionError("Values did not have UserDefinedFunction type")
     }
   }
+
+  override def dotnetValue(v: UserDefinedFunction): String =
+    throw new NotImplementedError("No translation found for complex parameter")
+
+  override def dotnetParamInfo: String = "UserDefinedFunction"
+
+//  override def dotnetLoadLine(modelNum: Int): String = {
```
mhamilton723, line 39: Remove?

core/src/test/scala/com/microsoft/ml/spark/automl/VerifyFindBestModel.scala
```diff
@@ -98,6 +98,6 @@ class VerifyFindBestModel extends EstimatorFuzzing[FindBestModel]{
     val randomForestClassifier = createRF.setLabelCol(mockLabelColumn)
     val model = randomForestClassifier.fit(createMockDataset)
     new FindBestModel()
-      .setModels(Array(model, model))
+      .setModels(Array(model.asInstanceOf[Transformer], model))
```
mhamilton723, line 101: this now needed?

core/src/test/scala/com/microsoft/ml/spark/codegen/DotnetTestGen.scala
```diff
@@ -0,0 +1,124 @@
...
+         |    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+         |      <PrivateAssets>all</PrivateAssets>
+         |      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+         |    </PackageReference>
+         |    <PackageReference Include="Microsoft.Spark" Version="2.0.0" />
+         |    <PackageReference Include="IgnoresAccessChecksToGenerator" Version="0.4.0" PrivateAssets="All" />
+         |  </ItemGroup>
+         |
+         |  <ItemGroup>
+         |    <ProjectReference Include="..\\..\\..\\..\\..\\..\\..\\..\\core\\src\\main\\dotnet\\dotnetBase.csproj" />
```
mhamilton723, line 96: Generate this path? Also this depends on core is that intended?

project/CodegenPlugin.scala
```diff
@@ -33,10 +34,12 @@ object CodegenPlugin extends AutoPlugin {
 
   val RInstallTag = Tags.Tag("rInstall")
   val TestGenTag = Tags.Tag("testGen")
+  val DotnetTestGenTag = Tags.Tag("dotnetTestGen")
```
mhamilton723, line 37: We should make a pyTestGen as well and just unify them with testGen so that testGen does it for all languages which I think more closely aligns with the other commands

core/src/main/scala/com/microsoft/ml/spark/codegen/DotnetWrappable.scala
```diff
@@ -0,0 +1,522 @@
...
+           |    return result;
+           |}
+           |""".stripMargin
+      case _: EstimatorParam | _: ModelParam | _: TransformerParam | _: EvaluatorParam | _: PipelineStageParam =>
+        val dType = p match {
+          case _: EstimatorParam | _: ModelParam =>
+            getParamInfo(p).dotnetType.substring(5, getParamInfo(p).dotnetType.length - 3) + "<object>"
+          case _ => getParamInfo(p).dotnetType
+        }
+        s"""
```
mhamilton723, line 359: We might want to put this info into the WrappableParam trait so that we don't need to pollute the core generator code

core/src/main/scala/com/microsoft/ml/spark/codegen/DotnetWrappable.scala
```diff
@@ -0,0 +1,522 @@
...
+            |{
+            |    var arrayList = new ArrayList(SparkEnvironment.JvmBridge);
+            |    foreach (var v in value)
+            |    {
+            |        arrayList.Add(v);
+            |    }
+            |    return $dotnetClassWrapperName(Reference.Invoke(\"set$capName\", (object)arrayList));
+            |}
+            |""".stripMargin
+      case _: ArrayMapParam =>
```
mhamilton723, line 181: Likewise here: we might want to put this into WrappableParam so that the code generator is slim and the implementations are distributed

core/src/main/scala/com/microsoft/ml/spark/codegen/DotnetWrappable.scala
```diff
@@ -0,0 +1,522 @@
...
+  def dotnetAdditionalMethods: String = {
+    ""
+  }
+
+  //noinspection ScalaStyle
+  protected def dotnetParamSetter(p: Param[_]): String = {
+    val capName = p.name match {
+      case "xgboostDartMode" => "XGBoostDartMode"
+      case "parallelism" => dotnetClassName match {
+        case "VowpalWabbitContextualBandit" => "ParallelismForParamListFit"
```
mhamilton723, line 124: Could you explain why these things need to be here? Perhaps a comment would help

mhamilton723 (commented):
You're a real hero Serena :)

core/src/main/dotnet/Base/Base.cs
```diff
@@ -0,0 +1,278 @@
...
+
+            return (T)constructor.Invoke(new object[] { reference });
+        }
+
+    }
+
+    /// <summary>
+    /// <see cref="ScalaPipelineStage"/> A stage in a pipeline, either an Estimator or a Transformer.
+    /// </summary>
+    public class ScalaPipelineStage : Params
```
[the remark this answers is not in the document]
serena-ruan: Ah yes they should be abstract!

core/src/main/dotnet/E2ETestUtils/E2ETestUtils.cs
```diff
@@ -0,0 +1,406 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+using Microsoft.Spark.Interop.Ipc;
+using Microsoft.Spark.Sql;
+using Microsoft.Spark.Services;
+using Microsoft.Spark.ML.Feature;
+using Microsoft.Spark.ML.Feature.Param;
+using Xunit;
+
+namespace MMLSparktest.Utils
```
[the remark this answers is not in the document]
serena-ruan: Yes

core/src/main/scala/com/microsoft/ml/spark/featurize/CleanMissingData.scala
```diff
@@ -158,6 +159,8 @@ class CleanMissingDataModel(val uid: String)
 
   def setFillValues(v: Array[Any]): this.type = set(fillValues, v)
 
+  def setFillValues(v: java.util.ArrayList[Any]): this.type = set(fillValues, v.asScala.toArray)
```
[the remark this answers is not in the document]
serena-ruan: Yes we need it. We need to convert the dotnet array into java.util.ArrayList so that scala can recognize it, and a function converting scala array to java ArrayList in scala side doesn't help, so we need exactly that type passed in the function to solve the problem.. We might discuss this in detail if you have any better ideas!

core/src/main/scala/com/microsoft/ml/spark/featurize/text/MultiNGram.scala
```diff
@@ -32,7 +32,7 @@ class MultiNGram(override val uid: String)
 
   setDefault(outputCol, uid + "_output")
 
-  val lengths = new TypedArrayParam[Int](
+  val lengths = new TypedIntArrayParam(
```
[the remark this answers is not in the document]
serena-ruan: params already has IntArrayParam : Param[Array[Int]] defined, so I guess this Typed is needed to distinguish between that one?

core/src/main/scala/org/apache/spark/ml/param/ArrayParamMapParam.scala
```diff
@@ -9,9 +9,15 @@ import com.microsoft.ml.spark.core.serialize.ComplexParam
...
+  extends ComplexParam[Array[ParamMap]](parent, name, doc, isValid)
+    with WrappableParam[Array[ParamMap]] {
 
   def this(parent: Params, name: String, doc: String) =
     this(parent, name, doc, ParamValidators.alwaysTrue)
 
+  override def dotnetValue(v: Array[ParamMap]): String =
+    throw new NotImplementedError("No translation found for complex parameter")
+
+  override def dotnetParamInfo: String = "ParamMap[]"
```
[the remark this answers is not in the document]
serena-ruan: Done. Guess it's dotnetType? (should the N be capitalized emmm..)

core/src/main/scala/org/apache/spark/ml/param/MapParam.scala
```diff
@@ -29,3 +30,43 @@ class MapParam[K, V](parent: Params, name: String, doc: String, isValid: Map[K,
...
+  override def jsonEncode(value: Map[String, String]): String = {
+    value.toJson.compactPrint
+  }
+
+  override def jsonDecode(json: String): Map[String, String] = {
+    json.parseJson.convertTo[Map[String, String]]
+  }
+}
+
+class StringIntMapParam(parent: Params, name: String, doc: String, isValid: Map[String, Int] => Boolean)
```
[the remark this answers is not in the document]
serena-ruan: Yes. I'll change them to extend MapParam

core/src/main/scala/org/apache/spark/ml/param/PythonWrappableParam.scala
serena-ruan: Actually no, and dotnet didn't used it.. I'll change it back!

core/src/test/scala/com/microsoft/ml/spark/automl/VerifyFindBestModel.scala
serena-ruan: Emm yes.. Otherwise it has `cannot resolve overloaded method` error

core/src/test/scala/com/microsoft/ml/spark/codegen/DotnetTestGen.scala
serena-ruan: Sure. Yes because I put those base files of dotnet into core.. Should we move it to an independent folder and let each package depends on it?

project/CodegenPlugin.scala
serena-ruan: Done!

core/src/main/scala/com/microsoft/ml/spark/codegen/DotnetWrappable.scala
serena-ruan: That's a great point! I should add another field to capture the different return type

core/src/main/scala/com/microsoft/ml/spark/codegen/DotnetWrappable.scala
serena-ruan: Ah do you mean putting the whole function into WrappableParam?

TomFinley (commented):
Hi @serena-ruan just did a first pass. I focused more on the .NET side of things, since I figure @mhamilton723 probably has the Scala side well in hand.

I have a few first reactions, which I might say are these:

1. It seems like much of this might benefit from being made `internal`, since it's always easier to go from internal to public (if absolutely necessary) than the reverse.
2. One thing I am not certain about is whether we're going for something that looks like Spark code as it appears in Scala or Python with PySpark, or whether we're going for something that looks more idiomatically like .NET?
3. It seems like rephrasing the interface in terms of the interfaces (which can be covariant/contravariant) is necessary if we want to enjoy some of the same benefits that Scala and Spark enjoys with its own constructs (I expand more on this).

I might have further comments later if you like, but those are just the top level things that occur to me. Great work so far incidentally, this is very exciting!

core/src/main/dotnet/Base/Base.cs
```diff
@@ -0,0 +1,278 @@
...
+using Microsoft.Spark.Sql.Types;
+using Microsoft.Spark.ML.Feature;
+using Microsoft.Spark.ML.Feature.Param;
+using Microsoft.Spark.Interop;
+using Microsoft.Spark.Interop.Ipc;
+
+namespace MMLSpark.Dotnet.Wrapper
+{
+
+    public class Params : Identifiable, IJvmObjectReferenceProvider
```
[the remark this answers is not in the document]
serena-ruan: Ah it should be abstract! Changed it.

core/src/main/scala/com/microsoft/ml/spark/codegen/CodeGen.scala
```diff
@@ -171,6 +179,45 @@ object CodeGen {
          |""".stripMargin)
   }
 
+  //noinspection ScalaStyle
+  def generateDotnetProjFile(conf: CodegenConfig): Unit = {
```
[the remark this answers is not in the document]
serena-ruan: Yes definitely, I can reserve a time to walk you through my test process and also discuss about those detailed comments you have!

core/src/main/dotnet/Base/Utils.cs
```diff
@@ -0,0 +1,132 @@
...
+    public interface ScalaMLWritable
+    {
+        /// <returns>a <see cref=\"ScalaMLWriter\"/> instance for this ML instance.</returns>
+        ScalaMLWriter Write();
+
+        /// <summary>Saves this ML instance to the input path</summary>
+        void Save(string path) => Write().Save(path);
+    }
+
+    public interface MLReader<T>
```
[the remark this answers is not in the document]
serena-ruan: I think we should follow the same pattern as origin spark ReadWrite, that the MLReader is constrained with type `T` and each model's reader can only load this type of model not others?

core/src/main/dotnet/Params/TransformerArray.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Spark.Interop;
+using Microsoft.Spark.Interop.Ipc;
+using MMLSpark.Dotnet.Wrapper;
+
+namespace Microsoft.Spark.ML.Feature.Param
+{
+    // <summary>
+    /// Array of Transformers.
+    /// </summary>
+    public class TransformerArray : IJvmObjectReferenceProvider
```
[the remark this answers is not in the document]
serena-ruan: I agree, this is used to reference the jvm under beneath, but since currently the `IJvmObjectReferenceProvider` is made public in dotnet/spark repo we need to make it public as well.

core/src/main/dotnet/Base/HashMap.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.Spark.Interop.Ipc;
+
+namespace Microsoft.Spark.Interop.Internal.Java.Util
+{
+    /// <summary>
+    /// HashMap class represents a <c>java.util.HashMap</c> object.
+    /// </summary>
+    public sealed class HashMap : IJvmObjectReferenceProvider
```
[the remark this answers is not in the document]
serena-ruan: This is used to cast .NET Dictionary to java hashMap so that scala can recognize it. I'll make it internal, because we only call it internally!

core/src/main/dotnet/Base/Schemas.cs
```diff
@@ -0,0 +1,121 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Spark.Interop;
+using Microsoft.Spark.Interop.Ipc;
+
+namespace MMLSpark.Dotnet.Utils
+{
+    public sealed class TimeSeriesPoint : IJvmObjectReferenceProvider
+    {
+        public readonly string timestamp;
```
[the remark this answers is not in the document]
serena-ruan: Thanks for those suggestions! But I'm still a little bit confused after seeing different samples for the `init`; should we always keep the field private even we only expose `init` other than `set` for the field?
To be specific:
1. ` private string _timestamp; 
 public string Timestamp 
 { 
   get { return _timestamp; } 
   init { _timestamp = value;}
 }`
2. ` public string Timestamp { get; init; }`
which option follows the c# convention more?

core/src/main/dotnet/Params/TransformerArray.cs
```diff
@@ -0,0 +1,42 @@
...
+        public TransformerArray() : this(SparkEnvironment.JvmBridge.CallConstructor(s_TransformerArrayClassName))
+        {
+        }
+
+        internal TransformerArray(JvmObjectReference jvmObject)
+        {
+            Reference = jvmObject;
+        }
+
+        public JvmObjectReference Reference { get; private set; }
```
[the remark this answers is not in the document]
serena-ruan: Yes make sense.

core/src/main/dotnet/Base/Base.cs
```diff
@@ -0,0 +1,278 @@
...
+        {
+        }
+
+        /// <summary>Evaluates the model output.</summary>
+        /// <param name=\"dataset\">The <see cref=\"DataFrame\"/> to evaluate the model against.</param>
+        /// <returns>double, evaluation result</returns>
+        public virtual double Evaluate(DataFrame dataset) =>
+            (double)Reference.Invoke("evaluate", dataset);
+
+        public Boolean IsLargerBetter => true;
```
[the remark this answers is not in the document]
serena-ruan: Done.

core/src/main/dotnet/Base/Base.cs
```diff
@@ -0,0 +1,278 @@
...
+            WrapAsType<M>((JvmObjectReference)Reference.Invoke("fit", dataset));
+
+    }
+
+    /// <summary>
+    /// <see cref="ScalaModel"/> Class for models that are fitted by estimators.
+    /// </summary>
+    public abstract class ScalaModel<M> : ScalaTransformer, Model<M> where M : ScalaModel<M>
+    {
+        public ScalaModel(string className) : base(className)
```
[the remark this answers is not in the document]
serena-ruan: Yes absolutely! I'll change them to internal.

core/src/main/dotnet/Params/TransformerArray.cs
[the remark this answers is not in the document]
TomFinley: Ah, I see. Still making the implementations [explicit](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation) would stop it from showing up, which seems desirable.

core/src/main/dotnet/Base/Schemas.cs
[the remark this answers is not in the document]
TomFinley: I'd say using the second one, with autoimplemented properties seems to be favored nowadays.

```csharp
public string Timestamp { get; init set; }
```

serena-ruan:
Pending on dotnet Spark team to release v2.1 in order to test on Spark 3.2 :)

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/DefaultParamInfo.scala
```diff
@@ -0,0 +1,121 @@
...
+      case cp: ComplexParam[_] => getComplexParamInfo(cp)
+      case p => {
+        logWarning(s"unsupported type $p")
+        UnknownInfo
+      }
+    }
+  }
+
+  //noinspection ScalaStyle
+  def getServiceParamInfo(dataType: ServiceParam[_]): ParamInfo[_] = {
```
mhamilton723, line 88: Int he future we might want to figure out a good way to abstract this stuff into "plugins" so that the core code is clean and each added parameter class has to do the legwork to hook into codegen/testgen

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/DotnetWrappable.scala
```diff
@@ -0,0 +1,519 @@
...
+           |    JvmObjectReference[] jvmObjects = (JvmObjectReference[])Reference.Invoke(\"get$capName\");
+           |    ParamMap[] result = new ParamMap[jvmObjects.Length];
+           |    for (int i=0; i < jvmObjects.Length; i++)
+           |    {
+           |        result[i] = new ParamMap(jvmObjects[i]);
+           |    }
+           |    return result;
+           |}
+           |""".stripMargin
+      case _: TransformerArrayParam | _: EstimatorArrayParam =>
```
mhamilton723, line 381: likewise here, in the future perhaps this is the content that we try to push into the class def of a new param class

mhamilton723 (changes requested):
Wowee Serena really big one! I left a few comments and perhaps we can also tour this as its so big I want to make sure I can give you good feedback. If I mention in the comment this is good future work feel free to leave as optional ;)

core/src/main/scala/com/microsoft/azure/synapse/ml/recommendation/SAR.scala
```diff
@@ -257,49 +257,4 @@ trait SARParams extends Wrappable with RecommendationParams {
   setDefault(timeDecayCoeff -> 30, activityTimeFormat -> "yyyy/MM/dd'T'h:mm:ss", supportThreshold -> 4,
     ratingCol -> C.RatingCol, userCol -> C.UserCol, itemCol -> C.ItemCol, similarityFunction ->
       "jaccard", timeCol -> "time", startTimeFormat -> "EEE MMM dd HH:mm:ss Z yyyy")
-
```
mhamilton723, line 260: are these removed on purpose?

serena-ruan: Yes, I'm moving them to org.apache.spark.ml.recommendation.RecommendationHelper's RecommendationParams since there're other classes need these.

mhamilton723 (changes requested):
Wowee this was a big one!

core/src/main/dotnet/src/Base/Schemas.cs
```diff
@@ -0,0 +1,309 @@
...
+            {
+                dict[(string)key.Invoke("toString")] = (string)((JvmObjectReference)parameters.Invoke("get", key)).Invoke("get");
+            }
+            this.Parameters = dict;
+        }
+
+        public JvmObjectReference Reference { get; init; }
+    }
+
+// /// TODO: Fix DiagnosticsInfo
```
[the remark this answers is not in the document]
serena-ruan: Deleted.

core/src/main/dotnet/src/org/apache/spark/ml/feature/StringIndexerModel.cs
```diff
@@ -0,0 +1,227 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+
+using System;
```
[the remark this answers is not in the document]
serena-ruan: These are generated but manually changed some like the constructors. These are needed because some of our tests depend on them and the official Microsoft.Spark hasn't incorporated them. We might want to auto-generate most of ML classes for Microsoft.Spark and once they publish them we can remove these from our repo and depend on Microsoft.Spark.

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/DotnetWrappable.scala
```diff
@@ -0,0 +1,322 @@
...
+          |/// Gets ${p.name} value for <see cref=\"${p.name}\"/>
+          |/// </summary>
+          |/// <returns>
+          |/// ${p.name}: ${p.doc}
+          |/// </returns>""".stripMargin
+    p match {
+      case sp: ServiceParam[_] =>
+        val dType = getGeneralParamInfo(sp).dotnetType
+        dType match {
+          case "TimeSeriesPoint[]" |
```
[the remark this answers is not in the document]
serena-ruan: I moved this inside ServiceParam but we can't change the string matching to real type matching, because the params exist in cognitive package and we can't import them into core package.

core/src/main/scala/com/microsoft/azure/synapse/ml/param/ModelParam.scala
```diff
@@ -0,0 +1,50 @@
...
+        |                typeof($parentClassType),
+        |                "s_className");
+        |    JvmObjectUtils.TryConstructInstanceFromJvmObject(
+        |                jvmObject,
+        |                classMapping,
+        |                out $dotnetReturnType instance);
+        |    return instance;
+        |}
+        |""".stripMargin
+  }
```
mhamilton723, line 48: Seen this code in a few places, might want to provide a trait that fills this in with something useful

mhamilton723, line 48: Seen this code in a few places, might want to provide a trait that fills this in with something useful

dotnetSetup.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
mhamilton723, line 1: Perhaps put this file in tools directory

mhamilton723, line 1: Perhaps put this file in tools directory

project/CodegenPlugin.scala
```diff
@@ -166,6 +213,10 @@ object CodegenPlugin extends AutoPlugin {
         version.value
       }
     },
+    dotnetVersion := {
+      val versionArray = version.value.split("-".toCharArray)
+      versionArray.head + "-rc" + versionArray.drop(1).head
+    },
```
mhamilton723, line 219: I added this for python because python and R dont allow fancy versions like ours. Is this also the case with .NET?

mhamilton723, line 219: I added this for python because python and R dont allow fancy versions like ours. Is this also the case with .NET?

serena-ruan: The error that I discussed with Mark N. the other day was due to the long SNAPSHOT version, and I tried removing all dashes in between still makes the version too long to recognize, so I cut it at the first number to make the assembly recognizable from projects.

core/src/main/scala/com/microsoft/azure/synapse/ml/param/ModelParam.scala
serena-ruan: Done.

project/CodegenPlugin.scala
```diff
@@ -166,6 +213,10 @@ object CodegenPlugin extends AutoPlugin {
         version.value
       }
     },
+    dotnetVersion := {
```
mhamilton723, line 216: Im so sorry to keep bugging you on this but i still think thise code might have some suprises. In particular i think you'll want to check whether the version is a release, snapshot, or dirty snapshot and combine accordingly. I think as it stands this will error on a release version because it wont have any "-" characters. I think its ok for dirty though because of your drop and dropRight calls.

project/CodegenPlugin.scala
```diff
@@ -166,6 +213,10 @@ object CodegenPlugin extends AutoPlugin {
...
+      val versionArray = version.value.split("-".toCharArray)
+      versionArray.head + "-rc" + versionArray.drop(1).dropRight(1).mkString("")
+    },
```
mhamilton723, line 219: Im so sorry to keep bugging you on this but i still think thise code might have some suprises. In particular i think you'll want to check whether the version is a release, snapshot, or dirty snapshot and combine accordingly. I think as it stands this will error on a release version because it wont have any "-" characters. I think its ok for dirty though because of your drop and dropRight calls.
```suggestion
    dotnetVersion := {
      if (version.value.contains("-") ){
        val versionArray = version.value.split("-".toCharArray)
        versionArray.head + "-rc" + versionArray.drop(1).dropRight(1).mkString("")
      } else {
       version.value
      }
    },
```

build.sbt
```diff
@@ -99,6 +99,66 @@ genBuildInfo := {
...
+       |""".stripMargin
+  val dotnetTestBaseDir = join(baseDirectory.value, "core", "src", "main", "dotnet", "test")
+  val dotnetHelperFile = join(dotnetTestBaseDir, "SynapseMLVersion.cs")
+  if (dotnetHelperFile.exists()) FileUtils.forceDelete(dotnetHelperFile)
+  FileUtils.writeStringToFile(dotnetHelperFile, fileContent, "utf-8")
+  runCmd(
+    Seq("dotnet", "pack", "--output", join(dotnetTestBaseDir, "target").getAbsolutePath),
+    dotnetTestBaseDir
+  )
+  val packagePath = join(dotnetTestBaseDir, "target", s"SynapseML.DotnetE2ETest.0.9.1.nupkg").getAbsolutePath
```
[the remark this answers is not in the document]
serena-ruan: I fix this version on purpose, because the only change to the dotnetTestBase is the version of synapseml jar that it depends on. This is only used for our internal tests, so I'm open to either, releasing different snapshot versions is also fine.

PR #1157, opened 2021-08-09
fix: Code Cleanup: naming conventions

victoriajmicrosoft:


suhas92:
The PR needs quite some work. Please look at this PR as here we are adding options property to be added.

https://github.com/Azure/mmlspark/pull/1162/files

Please make sure only opinion mining changes are reflected in commits. i am seeing multiple entities changes as well which should be removed.

PR #1160, opened 2021-08-10
fix: improve LGBM error message for invalid slot names

mhamilton723:


imatiach-msft (approved):

lightgbm/src/test/scala/com/microsoft/ml/spark/lightgbm/split2/VerifyLightGBMRegressor.scala
```diff
@@ -41,6 +42,20 @@ class VerifyLightGBMRegressor extends Benchmarks
     super.testSerialization()
   }
 
+  test("foo") {
```
imatiach-msft, line 45: remove?

mhamilton723: TY!

PR #1164, opened 2021-08-14
docs: Adding document and notebooks for ONNXModel

memoryz:


deep-learning/src/main/scala/com/microsoft/ml/spark/onnx/ONNXModel.scala
```diff
@@ -262,16 +264,32 @@ object ONNXModel extends ComplexParamsReadable[ONNXModel] with Logging {
...
+      case x: T => bufferWrite(x)
       case s: Seq[_] =>
-        flattenNested(s)
-      case a: Array[_] =>
-        flattenNested(a)
+        writeNestedSeqToBuffer(s, bufferWrite)
     }
   }
 
+  private def writeNestedSeqToStringBuffer(nestedSeq: Seq[_], size: Int): ArrayBuffer[String] = {
```
mhamilton723, line 275: Could you share a little details on why this stuff is necessary?

docs/onnx.md
```diff
@@ -0,0 +1,55 @@
...
+
+[ONNX](https://onnx.ai/) is an open format to represent both deep learning and traditional machine learning models. With ONNX, AI developers can more easily move models between state-of-the-art tools and choose the combination that is best for them.
+
+MMLSpark now includes a Spark transformer to bring an trained ONNX model to Apache Spark, so you can run inference on your data with Spark's large-scale data processing power.
+
+## Setup and installation
+
+In addition to the mmlspark-core and mmlspark-deep-learning jars, the ONNX model transformer also depends on the ONNX runtime library.
+
+For CPU clusters, the maven coordinate is `com.microsoft.onnxruntime:onnxruntime:1.8.1` and for GPU clusters, the maven coordinate is `com.microsoft.onnxruntime:onnxruntime_gpu:1.8.1`.
```
mhamilton723, line 18: I thought this was included in the deps?

core/src/test/scala/com/microsoft/ml/spark/nbtest/DatabricksUtilities.scala
```diff
@@ -50,7 +50,9 @@ object DatabricksUtilities extends HasHttpClient {
     Map("pypi" -> Map("package" -> "nltk")),
     Map("pypi" -> Map("package" -> "bs4")),
     Map("pypi" -> Map("package" -> "plotly")),
-    Map("pypi" -> Map("package" -> "Pillow"))
+    Map("pypi" -> Map("package" -> "Pillow")),
+    Map("pypi" -> Map("package" -> "onnxmltools")),
+    Map("pypi" -> Map("package" -> "lightgbm"))
```
mhamilton723, line 55: Are we able to switch this back to using our LGBM?

mhamilton723 (changes requested):

core/src/test/scala/com/microsoft/ml/spark/nbtest/DatabricksUtilities.scala
memoryz: Unfortunately no. Our LGBM outputs the string representation for a "Booster" model, and I cannot find an API to convert it to a "Classifier" model. `onnxmltools` (as well as Hummingbird) only supports conversion of a "Classifier" model, but not a "Booster" model.

deep-learning/src/main/scala/com/microsoft/ml/spark/onnx/ONNXModel.scala
memoryz: This method `writeNestedSeqToStringBuffer`, together with `writeNestedSeqToBuffer`, are used to replace the `flattenNested` method.

`flattenNested` flattens a nested Seq to a 1-d array, but is super slow due to repeated memory allocation.
`writeNestedSeqToStringBuffer` and `writeNestedSeqToBuffer` both allocates the memory buffer one time, and write the nested `Seq` to the buffer. I did some local testing, and they appear to be 5-10 times faster than `flattenNested`.

The reason I need two methods is because there is not a nio.Buffer class for string type, so I had to replace it with a ArrayBuffer[String].

core/src/test/scala/com/microsoft/ml/spark/nbtest/DatabricksUtilities.scala
imatiach-msft: @memoryz you can use loadNativeModelFromString or loadNativeModelFromFile to load a native lightgbm model file to create an mmlspark LightGBMClassificationModel:

https://github.com/Azure/mmlspark/blob/master/lightgbm/src/main/python/mmlspark/lightgbm/LightGBMClassificationModel.py#L25

docs/onnx.md
memoryz: But they're not packaged in the mmlspark jars right? So in Databricks clusters, I always had to manually install external maven dependencies like `com.microsoft.cntk:cntk:2.4`, `com.microsoft.ml.lightgbm:lightgbmlib`, `io.spray:spray-json:1.3.2` and `org.openpnp:opencv:3.2.0-1`. In this case, you can install either "onnxruntime" or "onnxruntime_gpu". The APIs are exactly the same, but the "onnxruntime" jar doesn't have CUDA dependencies packaged together, so won't be able to utilize GPU, and will fallback to use CPU on a GPU cluster.

core/src/test/scala/com/microsoft/ml/spark/nbtest/DatabricksUtilities.scala
memoryz: > @memoryz you can use loadNativeModelFromString or loadNativeModelFromFile to load a native lightgbm model file to create an mmlspark LightGBMClassificationModel:
> 
> https://github.com/Azure/mmlspark/blob/master/lightgbm/src/main/python/mmlspark/lightgbm/LightGBMClassificationModel.py#L25

But what I need is a non-mmlspark `LGBMClassifier`: https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMClassifier.html#lightgbm.LGBMClassifier.

What I'm doing here is to use onnxmltools to convert it to a ONNX model. I don't think onnxmltools works with MMLSpark's `LightGBMClassificationModel` class.

memoryz: Let me look into this a bit more.

imatiach-msft: so you mean that you want to create a lightgbm model, specifically LGBMClassifier, from a native lightgbm file?
Interesting, you can unofficially do it like this:
https://github.com/interpretml/interpret-community/blob/0923ad59f104f788941f39a197387748f8bfbef4/python/interpret_community/mimic/models/lightgbm_model.py#L397


see issues for reference:
https://github.com/Microsoft/LightGBM/issues/1217
https://github.com/Microsoft/LightGBM/issues/1942

I'm not sure how that relates to the mmlspark lightgbm model that I thought Mark was referring to.

memoryz: > @memoryz so you mean that you want to create a lightgbm model, specifically LGBMClassifier, from a native lightgbm file?
> Interesting, you can unofficially do it like this:
> https://github.com/interpretml/interpret-community/blob/0923ad59f104f788941f39a197387748f8bfbef4/python/interpret_community/mimic/models/lightgbm_model.py#L397
> 
> see issues for reference:
> [microsoft/LightGBM#1217](https://github.com/microsoft/LightGBM/issues/1217)
> [microsoft/LightGBM#1942](https://github.com/microsoft/LightGBM/issues/1942)
> 
> I'm not sure how that relates to the mmlspark lightgbm model that I thought Mark was referring to.

I'll take a look and get back to this.

docs/onnx.md
memoryz: Turns out Databricks installs the transitive dependencies. I was under the impression it wouldn't do so. Anyway I removed this section from the doc. It no longer applies.

core/src/test/scala/com/microsoft/ml/spark/nbtest/DatabricksUtilities.scala
memoryz: Turns out the onnxmltools package does support converting a lightgbm booster model directly. It's Hummingbird that doesn't support the conversion of booster. Since we're not using Hummingbird for now, I modified the notebook to use mmlspark's LightGBM to train the model, then export the a booster, then convert to onnx. We still need the lightgbm python package though, because the booster object is defined there.

mhamilton723 (approved):

PR #1173, opened 2021-08-29
Solving the penalisation issue through a new objective function 

navya002:
Solved the issue of penalising the negative errors more than the positive ones by creating an objective function for the lightGBM Regression task. I have also included a working case on a sample dataset (just for demonstration  purpose) and did all types of testing on it and came down to 96-97% accuracy for that problem statement .

### DIFF
--- a/lightGBMRegressor.py
+++ b/lightGBMRegressor.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[ ]:
+
+
+
+import numpy as np
+import matplotlib.pyplot as plt
+import pandas as pd
+from datetime import datetime 
+from sklearn import metrics
+from sklearn.metrics import roc_auc_score
+from sklearn.metrics import confusion_matrix
+import seaborn as sns
+import keras.backend as k
+penaliser_controller=0.25
+# Importing the dataset
+df = pd.read_csv("data/wisconsin_breast_cancer_dataset.csv")
+
+#Rename Dataset to Label to make it easy to understand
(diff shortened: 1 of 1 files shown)


imatiach-msft:
@navya002 the example you wrote just uses single node lightgbm and not distributed lightgbm in mmlspark, did you mean to post this example in:

https://github.com/Microsoft/LightGBM

this repo is for the spark-based distributed lightgbm only, sorry for any confusion

imatiach-msft:
also in the example it looks like you are specifying custom metric and not objective:
```
'metric':['auc', 'penalisation_loss'],
```
I think you want to specify custom objective instead of metric?

navya002:
Hi ,

Yes that is correct, I have submitted it for the problem :" I have a simple
regression task (using a LightGBMRegressor) where I want to penalize
negative predictions more than positive ones. Is there a way to achieve
this with the default regression LightGBM objectives  If not, is it somehow
possible to define (many example for default LightGBM model) and pass a
custom regression objective?"


I wanted to say that for solving underestimation and overestimation
quartile can be used and for that case I customised the loss
function(wherein the error could be better adjusted to the mean by squaring
the penalisation error) and yes you are correct , I will look into it.
Thank you for the clarification.


Thankfully
Navya Singh

imatiach-msft:
@navya002 sorry I think there is still some confusion based on your reply.  The original question was in this repository for mmlspark distributed lightgbm, not single node lightgbm.  If you want to add this to lightgbm repository (https://github.com/Microsoft/LightGBM) please feel free to do so.  This repository is for the distributed spark-based version of lightgbm.

PR #1176, opened 2021-09-01
fix: fix cognitive service errors

serena-ruan:


cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/CognitiveServiceBase.scala
```diff
@@ -91,7 +91,9 @@ trait HasServiceParams extends Params {
   }
 
   protected def shouldSkip(row: Row): Boolean = getRequiredParams.exists { p =>
-    emptyParamData(row, p)
+    if (emptyParamData(row, p))
+      throw new NullPointerException(s"required param undefined: $p")
```
mhamilton723, line 95: This will never throw false if the error is in here. What I'm thinking is that we look at the required params in the transformSchema function to ensure all required params are set with either a value or a column. We can then call transformSchema before we transform to add validation there too

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
```diff
@@ -76,9 +78,15 @@ trait TextAsOnlyEntity extends HasTextInput with HasCognitiveServiceInput {
 
   override protected def prepareEntity: Row => Option[AbstractHttpEntity] = {
     r =>
-      Some(new StringEntity(
-        getValueOpt(r, text)
-          .map(x => x.map(y => Map("Text" -> y))).toJson.compactPrint, ContentType.APPLICATION_JSON))
+      val textVal = getValueOpt(r, text)
+      if (textVal.nonEmpty) {
+        val content = textVal.get.getClass.getName match {
```
mhamilton723, line 83: Chris' second issue was caused by having nulls within a batch. Perhaps we should add that case as a test and ensure this new func can handle. I think in my TA batching logic it had to get pretty hairy to handle that unfortunately. Perhaps we can use similar logic. Also we might want to consider a better solution that just batching elements into single arrays. But we can push that to a later PR as that is current behavior in TA

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
```diff
@@ -171,6 +211,8 @@ class Translate(override val uid: String) extends TextTranslatorBase(uid)
 
   def setToLanguage(v: Seq[String]): this.type = setScalarParam(toLanguage, v)
 
+  def setToLanguage(v: String): this.type = setScalarParam(toLanguage, Seq(v))
```
mhamilton723, line 214: Nice! We might need to add this to python methods too

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/CognitiveServiceBase.scala
mhamilton723: nit: Should be an IllegalArgumentException

mhamilton723 (changes requested):
Thank you!!

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
serena-ruan: I guess this will automatically work? We're calling `self._java_obj = self._java_obj.setXXX(value)` and the java object will find the corresponding set function?

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/CognitiveServiceBase.scala
mhamilton723: We could have the logic in getInternalTransformer. I think the logic should live in the "control-plane" which executes on the head node rather than code inside a mapPartitions like shouldSkip

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
mhamilton723: I think the error comes from when there are nulls of text which would ordinarily be skipped but they are in a batch so the mess up the whole batch

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
mhamilton723: Youre right!

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
```diff
@@ -387,20 +488,66 @@ trait HasTextAndTranslationInput extends HasServiceParams {
...
+            ""
+          }
+
+          val post = new HttpPost(base + appended)
+          getValueOpt(row, subscriptionKey).foreach(post.setHeader("Ocp-Apim-Subscription-Key", _))
+          getValueOpt(row, subscriptionRegion).foreach(post.setHeader("Ocp-Apim-Subscription-Region", _))
+          post.setHeader("Content-Type", "application/json; charset=UTF-8")
+
+          val json = textAndTranslations.head.getClass.getTypeName match {
+            case "scala.Tuple2" => textAndTranslations.map(
```
mhamilton723, line 534: This looks a little fishy, is there a reason we cant use regular pattern matching here?

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -47,23 +52,46 @@ class TranslateSuite extends TransformerFuzzing[Translate]
     .setConcurrency(5)
 
   def translationTextTest(translator: Translate,
-                          df: DataFrame,
-                          expectString: String): Boolean = {
-    val results = translator
+                          df: DataFrame): DataFrame = {
```
mhamilton723, line 55: Might want to rename this method to better reflect what it does

cognitive/src/test/scala/com/microsoft/ml/spark/cognitive/split1/TranslatorSuite.scala
```diff
@@ -10,7 +10,6 @@ import com.microsoft.ml.spark.core.test.fuzzing.{TestObject, TransformerFuzzing}
 import org.apache.spark.ml.util.MLReadable
 import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.functions.{col, flatten}
-import org.scalactic.Equality
 
 trait TranslatorKey {
```
mhamilton723, line 14: We might want to add a test case for the auto-batching stuff you added

mhamilton723 (changes requested):
Love how you have applied these fixes to whole lib

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextTranslator.scala
serena-ruan: Yes, because we added `def setTextAndTranslation(v: (String, String)): this.type = setScalarParam(textAndTranslation, Seq(v))` function and in this case even though in `getValue` function we cast it into Seq[(String, String)] the underlying type is still Tuple2, and for the Tuple case I can't cast it into Seq[Row] and use `Map("Text" -> s.getString(0), "Translation" -> s.getString(1))` directly...

mhamilton723 (approved):

PR #1178, opened 2021-09-07
feat: add NERPii

xuwq1993:


cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalytics.scala
```diff
@@ -293,6 +293,18 @@ class NER(override val uid: String) extends TextAnalyticsBase(uid) with BasicLog
   def urlPath: String = "/text/analytics/v3.0/entities/recognition/general"
 }
 
+object NERPii extends ComplexParamsReadable[NERPii]
```
mhamilton723, line 296: Can we rename to PII?

mhamilton723 (changes requested):
Looks good one small nit!

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalytics.scala
xuwq1993: done

mhamilton723 (approved):

PR #1190, opened 2021-09-17
Feat: Build our new website

serena-ruan:


mhamilton723:
Awesome first draft of this!!! Here's some thoughts

1) We should clean out all the content from the template that's not used

2) Theres a few TODOs on main page that we should clean up: Boringly Easy -> Install

3) We should make a section on the main page that shows how simple it is to apply a variety of algos. Like one tab for Text Analytics, one for Deep Learning with new ONNX Stuff, one for model interpretability, one for LightGBM etc. These would be short little simple examples that show off how simple these experiments are with mmlspark

4) We should build out the automated creation of this. In particular, if you had to run nbconvert to generate this content we should make a script to do this automatically from the notebooks. If you had to do any post-processing we should do this in the upstream notebooks and then use an automated tool to regen the notebooks. This tool might also add a link to each generated MD with a button that says "download as a notebook" or points to the original notebook for people to download and move to databricks

PR #1196, opened 2021-09-27
Added TransformerFuzzing

suhas92:


cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalytics.scala
```diff
@@ -344,18 +344,6 @@ class PII(override val uid: String)
   def urlPath: String = "/text/analytics/v3.1/entities/recognition/pii"
 }
 
-object PII extends ComplexParamsReadable[PII]
```
mhamilton723, line 347: Any reason for removing this?

mhamilton723 (changes requested):

cognitive/src/main/scala/com/microsoft/ml/spark/cognitive/TextAnalytics.scala
suhas92: this PII object is duplicate, line 324.

PR #1221, opened 2021-10-20
perf: website enhancement

serena-ruan:


serena-ruan:
Issue with docSearch -- https://github.com/algolia/docsearch-configs/issues/4758

PR #1222, opened 2021-10-20
fix: add logging for number of columns and rows when creating datasets, set useSingleDatasetMode=True by default, fix validation dataset handling

imatiach-msft:
- add back some logging that was removed previously for tracking generated dataset size
- set useSingleDatasetMode=True by default
- fix issue with validation dataset failing in single dataset mode


lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/DatasetAggregator.scala
```diff
@@ -434,9 +438,12 @@ private[lightgbm] abstract class BaseSparseAggregatedColumns(chunkSize: Int)
     }
   }
 
-  def generateDataset(referenceDataset: Option[LightGBMDataset], datasetParams: String): LightGBMDataset = {
+  def generateDataset(referenceDataset: Option[LightGBMDataset],
+                      datasetParams: String, log: Logger): LightGBMDataset = {
     indexPointerArrayIncrement(getIndexPointers.array)
     val pointer = lightgbmlib.voidpp_handle()
+    val numRows = indptrCount.get() - 1
+    log.info(s"LightGBM task generating sparse dataset with $numRows rows and $numCols columns")
```
mhamilton723, line 446: Can lightGBM dataset just extend logging instead of passing this object around?

mhamilton723 (changes requested):
One minor Question

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/DatasetAggregator.scala
imatiach-msft: done!

mhamilton723 (approved):

PR #1226, opened 2021-10-22
docs: Adding document and notebook for Data Balance Analysis

ms-kashyap:


memoryz:
> @memoryz @mhamilton723 Where do you guys think I should host the Data Balance Analysis md? Currently, I have it in "features" under a new folder named "exploratory". However, do you guys have suggestions for better places than this?

I think having the folder structure matching the package name is a good idea. In the future, we might want to bundle this with the explainers into a rai module, so the doc can be moved together.

memoryz (approved):

mhamilton723 (approved):

PR #1241, opened 2021-11-04
docs: Introduce Responsible AI section on website (Interpretability + DataBalanceAnalysis)

ms-kashyap:
I am making these changes for the "Next" section of the website. Did the following:
- Renaming the "Model Interpretability" section on the website to "Responsible AI" for both Features and Examples
- Additionally, moving Data Balance Analysis into this section
- Generated "Interpretability - Explanation Dashboard" markdown from notebook which was added a couple days ago
- Generated "DataBalanceAnalysis - Adult Census Income" markdown from notebook + pictures that it references

The "DataBalanceAnalysis - Adult Census Income" notebook in `notebooks/` contains cell outputs (matplotlib figures), so I had to make some changes to `notebookconvert.py` to:
- From the notebook, generate a markdown file with no spaces in the name
- Generate the extra folder and files of cell outputs with no spaces in their names
- Rename the markdown file to match original notebook name

Also note: In `website/src/pages/index.js`, I had to point to `docs/next/features` instead of `docs/features` because the folder structure changes are reflected in Next, not 0.9.1. I wonder if there is a better way for me to do this.

The Responsible AI section under Features and Examples will look like this:
image
image


memoryz:
Can we also do these two things:
1. Move the Snow Leopard Detection notebook to Examples -> Responsible AI section.
2. Rename Features -> Responsible AI -> "About" to "Model Interpretation on Spark"

website/docs/features/responsible_ai/Model Interpretation on Spark.md
```diff
@@ -26,9 +26,9 @@ Both explainers extends from `org.apache.spark.ml.Transformer`. After setting up
 
 To see examples of model interpretability on Spark in action, take a look at these sample notebooks:
 
-- [Tabular SHAP explainer](/docs/examples/model_interpretability/Interpretability%20-%20Tabular%20SHAP%20explainer)
-- [Image explainers](/docs/examples/model_interpretability/Interpretability%20-%20Image%20Explainers)
-- [Text explainers](/docs/examples/model_interpretability/Interpretability%20-%20Text%20Explainers)
+- [Tabular SHAP explainer](../../../examples/responsible_ai/Interpretability%20-%20Tabular%20SHAP%20explainer)
```
ms-kashyap, line 29: Reason for absolute -> relative file path change:
- `/docs/examples/model_interpretability/` points to versioned docs (`website/versioned_docs/version-0.9.1/`)
- `/docs/next/examples/model_interpretability/` points to latest docs (`website/docs/`)

So when Next becomes 0.9.2, then we would have to remove the `/next/` in each path if we went with absolute paths

But if we do relative, then we can avoid this change.

website/src/pages/index.js
```diff
@@ -41,7 +41,7 @@ model_prediction_df = (ONNXModel()
   },
   {
     label: "Model Interpretability",
-    further: "/docs/features/model_interpretability/about",
+    further: "/docs/next/features/responsible_ai/Model%20Interpretation%20on%20Spark",
```
ms-kashyap, line 44: This is the only place we will have to remove `/next/` from when Next becomes 0.9.2.

This is the main landing page, and since the `responsible_ai` section is not in 0.9.1 versioned folder, we have to point to the latest one using `docs/next/features/responsible_ai`.

website/src/plugins/examples/index.js
```diff
@@ -12,11 +12,13 @@ function all_examples_for_type(folder, type) {
   let examples = [];
   let dir = path.join(__dirname, `../../../docs/${folder}/${type}`);
   fs.readdirSync(dir).forEach(function (file) {
-    let name = file.split(".").slice(0, -1).join(".");
-    let data = fs.readFileSync(path.join(dir, file));
-    const { frontMatter } = parseMarkdownString(data);
-    frontMatter["name"] = name;
-    examples.push(frontMatter);
+    if (file.endsWith(".md")) {
```
ms-kashyap, line 15: Originally, this iterated over not only files, but also folders that `jupyter nbconvert` generates for supporting images in generated markdown files. 

However, we only need to iterate over markdown files, so I made this change.

mhamilton723 (dismissed):

website/notebookconvert.py
```diff
@@ -1,39 +1,57 @@
...
 
     dic = {
         "CognitiveServices - Overview": os.path.join(outputdir, "features"),
         "Classification": os.path.join(outputdir, "examples", "classification"),
         "CognitiveServices": os.path.join(outputdir, "examples", "cognitive_services"),
+        "DataBalanceAnalysis": os.path.join(outputdir, "examples", "responsible_ai"),
         "DeepLearning": os.path.join(outputdir, "examples", "deep_learning"),
-        "Interpretability": os.path.join(outputdir, "examples", "model_interpretability"),
+        "Interpretability": os.path.join(outputdir, "examples", "responsible_ai"),
+        "ModelInterpretability": os.path.join(outputdir, "examples", "responsible_ai"),
```
serena-ruan, line 46: Should "ModelInterpretability" be put into "features"?

website/notebookconvert.py
```diff
@@ -1,39 +1,57 @@
...
-    print("Converting {} into markdown \n".format(file_path))
-    convert_cmd = 'jupyter nbconvert --output-dir=\"{}\" --to markdown \"{}\"'.format(outputdir, file_path)
+
+def convert_notebook_to_markdown(folder, nb, outputdir):
+    file_path = os.path.join(folder, nb)
+    print(f"Converting {file_path} into markdown")
+
+    # If the notebook contains cell outputs such as figures, a folder containing cell output images is generated alongside the markdown file
+    # By default, both the folder and files contain the notebook name. But spaces in the notebook name create linking errors in the generated markdown
+    # Therefore, we first generate the markdown file, output folder, and output files with no spaces
```
serena-ruan, line 23: Love this! But for figures could we upload them to our public blob (https://mmlspark.blob.core.windows.net/graphics) and use the url directly? Just want to be consistent with the structure of /notebooks folder.

website/src/plugins/examples/index.js
serena-ruan: If you replace the images with url links, this won't be needed anymore. But pls feel free to keep it if you prefer.

serena-ruan (changes requested):

website/src/plugins/examples/index.js
```diff
@@ -28,9 +30,10 @@ function all_examples() {
     `features/CognitiveServices - Overview.md`,
     `examples/ConditionalKNN - Exploring Art Across Cultures.md`,
     `examples/CyberML - Anomalous Access Detection.md`,
+    `examples/responsible_ai/DataBalanceAnalysis - Adult Census Income.md`,
+    `examples/responsible_ai/ModelInterpretability - Snow Leopard Detection.md`,
```
serena-ruan, line 34: Likewise here, I would like to keep one example "ModelInterpretability - Snow Leopard Detection.md" in features folder  😃

website/notebookconvert.py
ms-kashyap: In an above comment in this PR, @memoryz requested that we move the Mode Interpretability notebooks (more specifically "ModelInterpretability - Snow Leopard Detection.md" to `examples` instead of `features` because it is more like an example :smile:

website/notebookconvert.py
ms-kashyap: I definitely don't have an issue with this; however, I think that keeping the inline images in the notebook is advantageous because:
- The notebook creator doesn't have to export the inline images and upload them to blob each time they either add images or edit images
- The notebook creator doesn't have to add an extraneous notebook cell that points to the image from blob
- When a user runs the notebook, the unintended effect will be this (user will see the graph twice -- once from generated code and once from markdown link):

image

What are your thoughts? :smile:

website/src/plugins/examples/index.js
ms-kashyap: I think Jason Wang suggested that we move ModelInterpretability - Snow Leopard Detection.md to examples. However, the Responsible AI folder now contains:
- `Model Interpretation on Spark.md`  (renamed from about.md)
- `Data Balance Analysis.md`

serena-ruan:
> Can we also do these two things:
> 
> 1. Move the Snow Leopard Detection notebook to Examples -> Responsible AI section.
> 2. Rename Features -> Responsible AI -> "About" to "Model Interpretation on Spark"

Hi Jason, could we keep the Snow Leopard Detection in Features still? There should be one outstanding notebook examples inside each sub folders under features. Or do you prefer putting another one?

serena-ruan:
> > Can we also do these two things:
> > 
> > 1. Move the Snow Leopard Detection notebook to Examples -> Responsible AI section.
> > 2. Rename Features -> Responsible AI -> "About" to "Model Interpretation on Spark"
> 
> Hi Jason, could we keep the Snow Leopard Detection in Features still? There should be one outstanding notebook examples inside each sub folders under features. Or do you prefer putting another one?

@memoryz

memoryz:
> > Can we also do these two things:
> > 
> > 1. Move the Snow Leopard Detection notebook to Examples -> Responsible AI section.
> > 2. Rename Features -> Responsible AI -> "About" to "Model Interpretation on Spark"
> 
> Hi Jason, could we keep the Snow Leopard Detection in Features still? There should be one outstanding notebook examples inside each sub folders under features. Or do you prefer putting another one?

Hi Serena, the snow leopard detection notebook uses the deprecated API. If there should be one outstanding notebook example, can we replace it with "Interpretability - Image Explainers.ipynb" instead?

serena-ruan:
> > > Can we also do these two things:
> > > 
> > > 1. Move the Snow Leopard Detection notebook to Examples -> Responsible AI section.
> > > 2. Rename Features -> Responsible AI -> "About" to "Model Interpretation on Spark"
> > 
> > 
> > Hi Jason, could we keep the Snow Leopard Detection in Features still? There should be one outstanding notebook examples inside each sub folders under features. Or do you prefer putting another one?
> 
> Hi @serena-ruan, the snow leopard detection notebook uses the deprecated API. If there should be one outstanding notebook example, can we replace it with "Interpretability - Image Explainers.ipynb" instead?

Sure! Thanks for the comment. @ms-kashyap Kashyap could you pls change it?

website/notebookconvert.py
mhamilton723: @ms-kashyap, how do you get it so that its the only cell of output shown? do you manually delete all of the other output?

ms-kashyap: If you want to clear output of all cells, then there's a button for that in VS Code
image

However, if you want to keep the output of one cell, then yeah you need to go to every other cell and clear its output manually and painstakingly 😅 
image

ms-kashyap: For @memoryz and @mhamilton723 's context, @serena-ruan and I talked offline and we are going with pointing to the matplotlib figures via blob urls in markdown cells below the code (and clearing all outputs).

If we encounter a notebook where we HAVE to keep cell outputs (which would in turn generate the extra folder + files), then we can rethink this at that point.

memoryz:
Thank you @ms-kashyap

memoryz (approved):

serena-ruan (approved):
Thanks for the changes! Looks perfect!

PR #1279, opened 2021-11-30
feat: blob help for anomaly detection

eisber:


mhamilton723 (changes requested):
Looks great! Just have a security related ask about blob

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/cognitive/split1/MultivariateAnamolyDetectionSuite.scala
```diff
@@ -36,7 +38,13 @@ object MADListModelsProtocol extends DefaultJsonProtocol {
 
 import com.microsoft.azure.synapse.ml.cognitive.split1.MADListModelsProtocol._
 
-object MADUtils extends AnomalyKey {
+trait storageCredentials {
+  lazy val connectionString: String = sys.env.getOrElse("STORAGE_CONNECTION_STRING", Secrets.StorageConnectionString)
+  lazy val storageKey: String = sys.env.getOrElse("STORAGE_KEY", Secrets.StorageKey)
```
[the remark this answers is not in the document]
serena-ruan: Then I guess I'll create another storage account, because in order to support using `endpoint and sas token` instead of connectionString, we need storage key to authenticate the client thus it could generate sas token afterwards.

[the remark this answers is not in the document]
serena-ruan: Or could we just support authenticate through connection string? It's way more simpler than the endpoint and sas token one:
`
def this(storageConnectionString: String, containerName: String) =
    this {
      val blobContainerClient = new BlobServiceClientBuilder()
        .connectionString(storageConnectionString)
        .credential(StorageSharedKeyCredential.fromConnectionString(storageConnectionString))
        .buildClient()
        .getBlobContainerClient(containerName.toLowerCase())
      if (!blobContainerClient.exists()) {
        blobContainerClient.create()
      }
      blobContainerClient
    }

  def this(storageName: String, storageKey: String, endpoint: String, sasToken: String, containerName: String) =
    this {
      val blobContainerClient = new BlobServiceClientBuilder()
        .endpoint(endpoint)
        .sasToken(sasToken)
        .credential(new StorageSharedKeyCredential(storageName, storageKey))
        .buildClient()
        .getBlobContainerClient(containerName.toLowerCase())
      if (!blobContainerClient.exists()) {
        blobContainerClient.create()
      }
      blobContainerClient
    }
`

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -300,6 +300,165 @@ class MultivariateAnomalyEstimator(override val uid: String) extends Estimator[D
   }
 }
 
+trait HasBlobHelper extends HasServiceParams {
```
mhamilton723, line 303: nit: we should probably rename this to something like SimpleMultiADParams as it has a bit more params than just for the blob helper

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -300,6 +300,165 @@ class MultivariateAnomalyEstimator(override val uid: String) extends Estimator[D
...
+  def getSASToken: String = getScalarParam(sasToken)
+
+  val containerName = new ServiceParam[String](this, "containerName", "Container that will be " +
+    "used to upload files to.")
+
+  def setContainerName(v: String): this.type = setScalarParam(containerName, v)
+
+  def getContainerName: String = getScalarParam(containerName)
+
+  val sourceDF = new DataFrameParam(this, "sourceDF", "DataFrame as multivariate " +
```
mhamilton723, line 355: Shouldnt this be what is passed into fit? Or is this some other dataframe needed for the fitting?

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -300,6 +300,165 @@ class MultivariateAnomalyEstimator(override val uid: String) extends Estimator[D
...
+    )
+
+    NamespaceInjections.pipelineModel(stages)
+  }
+
+  override def fit(dataset: Dataset[_]): DetectMultivariateAnomaly = {
+    logFit({
+      import MADJsonProtocol._
+
+      val df = getInternalTransformer(dataset.schema)
```
mhamilton723, line 437: nit: might want to rename to rows as it is no longer a dataframe

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/cognitive/split4/MultivariateAnamolyDetectionSuite.scala
```diff
@@ -34,9 +37,15 @@ object MADListModelsProtocol extends DefaultJsonProtocol {
   implicit val MADLMRespEnc: RootJsonFormat[MADListModelsResponse] = jsonFormat4(MADListModelsResponse)
 }
 
-import com.microsoft.azure.synapse.ml.cognitive.split1.MADListModelsProtocol._
+import com.microsoft.azure.synapse.ml.cognitive.split4.MADListModelsProtocol._
 
-object MADUtils extends AnomalyKey {
+trait storageCredentials {
```
mhamilton723, line 42: rename to StorageCredentials

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/cognitive/split4/MultivariateAnamolyDetectionSuite.scala
```diff
@@ -198,3 +207,128 @@ class MultiAnomalyEstimatorSuite extends EstimatorFuzzing[MultivariateAnomalyEst
...
+      .transform(emptyDf)
+      .withColumn("status", col("result.summary.status"))
+      .withColumn("variableStates", col("result.summary.variableStates"))
+      .select("status", "variableStates")
+      .collect()
+
+    assert(result.head.getString(0).equals("READY"))
+    assert(result.head.getSeq(1).length.equals(3))
+
+    madDelete(model.getModelId)
```
mhamilton723, line 250: Right now cleanup won't run unless tests pass (which can cause bloat of of objects). Consider using a try finally to ensure cleanup runs. Also, it might be good to put the delete model func on the model itself for easy cleanup.

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/cognitive/split4/MultivariateAnamolyDetectionSuite.scala
```diff
@@ -198,3 +207,128 @@ class MultiAnomalyEstimatorSuite extends EstimatorFuzzing[MultivariateAnomalyEst
...
+  override def testSerialization(): Unit = {
+    println("ignore the Serialization Fuzzing test because fitting process takes more than 3 minutes")
+  }
+
+  override def testExperiments(): Unit = {
+    println("ignore the Experiment Fuzzing test because fitting process takes more than 3 minutes")
+  }
+
+  override def afterAll(): Unit = {
+    val models = madListModels().parseJson.convertTo[MADListModelsResponse].models.map(_.modelId)
```
mhamilton723, line 320: this could mess up other concurrent running instances in the build if this cleans up while another instance is running. Consider keeping some hidden global variable to keep track of which specific models are created in this JVM or robustly clean up each individual one created.

pipeline.yaml
```diff
@@ -542,6 +542,10 @@ jobs:
         PACKAGE: "cognitive.split3"
         FFMPEG: "true"
         FLAKY: "true"
+      cognitive4:
+        PACKAGE: "cognitive.split4"
+        FFMPEG: "true"
```
mhamilton723, line 547: dont need ffmpeg

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/AnomalyDetectionBlobHelpers.scala
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/** Helper to upload data from dataframe to Azure Blob storage. Stored as CSV and zipped.
+  *
+  * @param blobContainerClient configured with credentials.
+  * @param timestampCol        timestamp column of the source dataframe.
+  */
+class AnomalyDetectionBlobHelpers(blobContainerClient: BlobContainerClient, timestampCol: String) {
+
+  var blobName = ""
```
mhamilton723, line 55: this mutable var worries me, seems like you could just return this from the upload function

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -300,6 +300,165 @@ class MultivariateAnomalyEstimator(override val uid: String) extends Estimator[D
...
+        .setSubscriptionKey(getSubscriptionKey)
+        .setLocation(getUrl.split("/".toCharArray)(2).split(".".toCharArray).head)
+        .setModelId(modelId)
+    })
+  }
+
+  override def copy(extra: ParamMap): Estimator[DetectMultivariateAnomaly] = defaultCopy(extra)
+
+  override def transformSchema(schema: StructType): StructType = {
+    getInternalTransformer(schema).transformSchema(schema)
```
mhamilton723, line 457: this wont yield the correct final schema because the pipeline returned doesn't include the DetectMultivariateAnomaly stuff. One option is to just implement all the code in fit, and manually implement the transform schema. Another is to put the  fit code into getInternalTransformer and then no need to override this.

mhamilton723 (changes requested):

mhamilton723 (approved):

PR #1281, opened 2021-12-01
fix: Fix error in R test path

mhamilton723:


mhamilton723:
Thankfully @serena-ruan has saved the day!!

PR #1282, opened 2021-12-02
perf: improve lightgbm training performance 4x-10x by setting num_threads to be cores-1 by default for single dataset mode

imatiach-msft:
In benchmarking, it was discovered that lightgbm training time could be reduced 4X-10X by setting the num_threads to be equal to the # of machine cores - 1 for the single dataset mode (which is now the default mode).
This actually was already suggested in the lightgbm docs:
https://lightgbm.readthedocs.io/en/latest/Parameters.html#num_threads
>>> for distributed learning, do not use all CPU cores because this will cause poor performance for the network communication

Poor network communication can lead to very slow training execution time.

On one scenario with a 37 GB dataset on disk with parameters:
```learning_rate = 0.1, num_leaves = 768, num_trees = 1000, min_data_in_leaf = 15000, max_bin = 512```
the training time without setting num_threads was 5.5 hours, while setting num_threads to (number of machine cores)-1 reduced the training time to 1.2 hours.  The change in performance will vary depending on the parameters used.

This PR sets the distributed lightgbm number of threads by default to be (num cores)-1  if the parameter is not specified by the user.  The user can still override the parameter.

### DIFF
--- a/lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
+++ b/lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
@@ -247,8 +247,12 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
     *
     * @return ExecutionParams object containing parameters related to LightGBM execution.
     */
-  protected def getExecutionParams: ExecutionParams = {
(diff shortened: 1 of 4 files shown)


imatiach-msft:
An added benefit seems to be that the lightgbm tests finish faster as well - the lightgbm1 unit tests finished in 11 min in this build and took 33 min in previous build :sweat_smile:

mhamilton723 (approved):

PR #1284, opened 2021-12-03
feat: ICE/PDP explainer 

ezherdeva:
In this PR, I'm introducing ICETransformer and adding it in the com.microsoft.ml.spark.explainers package.

ICETransformer displays the model dependence on specified features with the given data frame.

- ICETransformer supports categorical and numeric features.
- It supports 2 types of plots: "average" - PDP and "individual" - ICE
- This transformer only supports a one-way dependence plot.

Also, I added ICECategoricalFeature and ICENumericFeature classes which are used in ICETransformer.

All of these classes can be called from the python side.​


ezherdeva:
@ms-kashyap

src/test/scala/com/microsoft/azure/synapse/ml/core/test/fuzzing/FuzzingTest.scala
```diff
@@ -241,6 +241,7 @@ class FuzzingTest extends TestBase {
       "com.microsoft.azure.synapse.ml.explainers.TextSHAP",
       "com.microsoft.azure.synapse.ml.explainers.VectorLIME",
       "com.microsoft.azure.synapse.ml.explainers.VectorSHAP",
+      "com.microsoft.azure.synapse.ml.explainers.ICETransformer"
```
ms-kashyap, line 244:
```suggestion
      "com.microsoft.azure.synapse.ml.explainers.ICETransformer",
```

core/src/main/python/synapse/ml/explainers/ICETransformer.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in project root for information.
+
+from synapse.ml.explainers._ICETransformer import _ICETransformer
+from pyspark.ml.common import inherit_doc
+from pyspark import SparkContext
+
+@inherit_doc
+class ICETransformer(_ICETransformer):
+    def setCategoricalFeatures(self, value):
```
mhamilton723, line 10: I think these should be auto-genned by the codegen system

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,252 @@
...
+
+    val predicted = getModel.transform(df.withColumn(feature, explodeFunc))
+    val targetCol = DatasetExtensions.findUnusedColumnName("target", predicted)
+    val dependenceCol = DatasetExtensions.findUnusedColumnName("feature__dependence", predicted)
+
+    val explainTarget = extractTarget(predicted.schema, targetClassesColumn)
+    val result = predicted.withColumn(targetCol, explainTarget)
+
+    getKind.toLowerCase match {
+      case this.averageKind =>
```
mhamilton723, line 92: nit: idt you need the "this" here

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  def transform(ds: Dataset[_]): DataFrame = {
+    transformSchema(ds.schema)
+    val df = ds.toDF
+    val idCol = DatasetExtensions.findUnusedColumnName("idCol", df)
+    val targetClasses = DatasetExtensions.findUnusedColumnName("targetClasses", df)
+    val dfWithId = df
+      .withColumn(idCol, monotonically_increasing_id())
+      .withColumn(targetClasses, this.get(targetClassesCol).map(col).getOrElse(lit(getTargetClasses)))
```
mhamilton723, line 124: nit: use getters directly here

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,252 @@
...
+
+    // collect feature values for all features from original dataset - dfWithId
+    val (categoricalFeatures, numericFeatures) = (this.getCategoricalFeatures, this.getNumericFeatures)
```
mhamilton723, line 127: no need for this

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,252 @@
...
+  override def transformSchema(schema: StructType): StructType = {
+    // Check the data type for categorical features
+    val categoricalFeaturesTypes= getCategoricalFeatures.map(f => schema(f.name).dataType)
+    val allowedCategoricalTypes = Array(StringType, BooleanType, ByteType, ShortType, IntegerType, LongType)
+    require(categoricalFeaturesTypes.forall(allowedCategoricalTypes.contains),
+      s"Data type for categorical features must be ${allowedCategoricalTypes.mkString("[", ",", "]")}.")
+
+    // Check the data type for numeric features
+    val numericFeaturesTypes= getNumericFeatures.map(f => schema(f.name).dataType)
+    val allowedNumericTypes = Array(FloatType, DoubleType, DecimalType)
```
memoryz, line 232: Add ByteType, ShortType, IntegerType, LongType

memoryz (changes requested):

mhamilton723 (changes requested):
Thank you so much for this fantastic work! I put in a few nits but to give it a good review perhaps we should chat so I can get a better idea of what this does then

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,252 @@
...
+class ICETransformer(override val uid: String) extends Transformer
+  with HasExplainTarget
+  with HasModel
+  with ICEFeatureParams
+  with Wrappable
+  with ComplexParamsWritable {
+
+  override protected lazy val pyInternalWrapper = true
+
+  def this() = {
```
ms-kashyap, line 74: nit: you can remove the braces here

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,252 @@
...
+    "Whether to return the partial dependence plot (PDP) averaged across all the samples in the " +
+      "dataset or individual feature importance (ICE) per sample. " +
+      "Allowed values are \"average\" for PDP and \"individual\" for ICE.",
+    ParamValidators.inArray(Array(averageKind, individualKind))
+  )
+
+  def getKind: String = $(kind)
+  def setKind(value: String): this.type = set(kind, value)
+
+  setDefault(kind -> "individual", numericFeatures -> Seq.empty[ICENumericFeature],
```
ms-kashyap, line 54: nit: I suggest you do something like this to make it look cleaner
```
setDefault(
   kind -> "individual",
   numericFeatures -> Seq.empty[ICENumericFeature],
   categoricalFeatures -> Seq.empty[ICECategoricalFeature]
)
```

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,252 @@
...
+        dependenceDfs.reduceOption(_.join(_, Seq(idCol), "inner"))
+          .map(sampled.join(_, Seq(idCol), "inner").drop(idCol)).get
+      case this.averageKind =>
+        dependenceDfs.reduce(_ crossJoin _)
+    }
+  }
+
+  private def collectCategoricalValues[_](df: DataFrame, feature: ICECategoricalFeature): Array[_] = {
+    val featureCountCol = DatasetExtensions.findUnusedColumnName("__feature__count__", df)
+    val values = df
```
ms-kashyap, line 155: You can remove `val values = ` and just make this `df.xyz()` statement the last statement in the function

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
ms-kashyap: [A similar check happens in data balance analysis validateSchema, check it out here](https://github.com/microsoft/SynapseML/blob/ff2aa020693b6521da988250376557bf128301ef/core/src/main/scala/com/microsoft/azure/synapse/ml/exploratory/DataBalanceParams.scala#L37)

core/src/main/python/synapse/ml/explainers/ICETransformer.py
memoryz: We discovered the same via python unit test failures. Elena will redo this part in the next commit.

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
ezherdeva: targetClassesCol is Optional by design, that's why we're using get like this

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
ezherdeva: Added backticks, because it's looking for a stable identifier. Ref: https://stackoverflow.com/questions/7078022/why-does-pattern-matching-in-scala-not-work-with-variables

core/src/main/python/synapse/ml/explainers/ICETransformer.py
```diff
@@ -0,0 +1,48 @@
...
+                if isinstance(value, str):
+                    list_values.append({"name": value})
+                elif isinstance(value, dict):
+                    list_values.append(value)
+                else:
+                    pass
+            self._java_obj.setCategoricalFeaturesPy(list_values)
+        return self
+
+    def setNumericFeatures(self, values: List[Dict]):
```
mhamilton723, line 30: nit: SHouldnt values be the Union type as above

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,275 @@
...
+        val values = collectCategoricalValues(dfWithId, f)
+        calcDependence(sampled, idCol, targetClasses, f.name, values, f.getOutputColName)
+    }
+    val calcNumericFunc: ICENumericFeature => DataFrame = {
+      f: ICENumericFeature =>
+        val values = collectSplits(dfWithId, f)
+        calcDependence(sampled, idCol, targetClasses, f.name, values, f.getOutputColName)
+    }
+
+    val dependenceDfs = (categoricalFeatures map calcCategoricalFunc) ++ (numericFeatures map calcNumericFunc)
```
mhamilton723, line 155: Combine both sequences then map with a case match on the Feature Class. it will be a bit cleaner and will avoid functions as values

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,275 @@
...
+      case `individualKind` =>
+        dependenceDfs.reduceOption(_.join(_, Seq(idCol), "inner"))
+          .map(sampled.join(_, Seq(idCol), "inner").drop(idCol)).get
+      case `averageKind` =>
+        dependenceDfs.reduce(_ crossJoin _)
+    }
+  }
+
+  private def collectCategoricalValues[_](df: DataFrame, feature: ICECategoricalFeature): Array[_] = {
+    val featureCountCol = DatasetExtensions.findUnusedColumnName("__feature__count__", df)
```
mhamilton723, line 170: nit: might want to use featureCount to be consistent with previous usage of this func

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,275 @@
...
+    if (featureNames.isEmpty) {
+      throw new Exception("No categorical features or numeric features are set to the explainer. " +
+        "Call setCategoricalFeatures or setNumericFeatures to set the features to be explained.")
+    }
+    // Check for duplicate feature specification
+    val duplicateFeatureNames = featureNames.groupBy(identity).mapValues(_.length).filter(_._2 > 1).keys.toArray
+    if (duplicateFeatureNames.nonEmpty) {
+      throw new Exception(s"Duplicate features specified: ${duplicateFeatureNames.mkString(", ")}")
+    }
+    this.validateSchema(schema)
```
mhamilton723, line 270: dont need "this"

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEFeature.scala
```diff
@@ -0,0 +1,168 @@
...
+  def getNumTopValue: Int = {
+    this.numTopValues.getOrElse(defaultNumTopValue)
+  }
+}
+
+/**
+  * Companion object to provide JSON serializer and deserializer for ICECategoricalFeature.
+  */
+object ICECategoricalFeature {
+  implicit val JsonFormat: JsonFormat[ICECategoricalFeature] = new JsonFormat[ICECategoricalFeature] {
```
mhamilton723, line 39: Do the default case class serializers not work here?

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEFeature.scala
```diff
@@ -0,0 +1,168 @@
...
+  def getNumSplits: Int = {
+    this.numSplits.getOrElse(defaultNumSplits)
+  }
+}
+
+/**
+  * Companion object to provide JSON serializer and deserializer for ICENumericFeature.
+  */
+object ICENumericFeature {
+  implicit val JsonFormat: JsonFormat[ICENumericFeature] = new JsonFormat[ICENumericFeature] {
```
mhamilton723, line 109: likewise here, seems like this is alot of work for json parsing and the default impl for case classes should do the trick

core/src/test/scala/com/microsoft/azure/synapse/ml/explainers/split1/ICEExplainerSuite.scala
```diff
@@ -0,0 +1,154 @@
...
+    }.head
+
+    assert(outputCol1.size == iceAvg.getNumericFeatures.head.getNumSplits + 1)
+  }
+
+  test("The length of explainer map for categorical feature is less or equal to it's numTopValues.") {
+    val outputCol: Map[Double, Vector] = output.select("col3_dependence").collect().map {
+      case Row(map: Map[Double, Vector]) =>
+        map
+    }.head
```
mhamilton723, line 97: this code is repeated in a few places, consider pulling into a function to keep it dry

core/src/test/scala/com/microsoft/azure/synapse/ml/explainers/split1/ICEExplainerSuite.scala
```diff
@@ -0,0 +1,154 @@
...
+import com.microsoft.azure.synapse.ml.explainers.{ICECategoricalFeature, ICENumericFeature, ICETransformer}
+import org.apache.spark.ml.linalg.Vector
+import org.apache.spark.ml.util.MLReadable
+import scala.jdk.CollectionConverters._
+
+
+class ICEExplainerSuite extends TestBase with TransformerFuzzing[ICETransformer] {
+
+  import spark.implicits._
+  val dataDF: DataFrame = (1 to 100).flatMap(_ => Seq(
```
mhamilton723, line 22: please make all things in the test class body lazy, these will run during python test generation if they arent

core/src/test/scala/com/microsoft/azure/synapse/ml/explainers/split1/ICEExplainerSuite.scala
```diff
@@ -0,0 +1,154 @@
...
+  val pipeline: Pipeline = new Pipeline().setStages(Array(
+    new StringIndexer().setInputCol("col2").setOutputCol("col2_ind"),
+    new OneHotEncoder().setInputCol("col2_ind").setOutputCol("col2_enc"),
+    new VectorAssembler().setInputCols(Array("col1", "col2_enc", "col3", "col4")).setOutputCol("features"),
+    new LogisticRegression().setLabelCol("label").setFeaturesCol("features")
+  ))
+  val model: PipelineModel = pipeline.fit(data)
+
+  val ice = new ICETransformer()
+  ice.setModel(model)
```
mhamilton723, line 40: perhaps make these a def as they are mutable by default

mhamilton723 (changes requested):
Left a bit more detailed feedback. Looks awesome though and appreciate all the hard work and iterations!

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEFeature.scala
memoryz: Oh I wasn't aware of that. Elena and I will look into that.

core/src/main/python/synapse/ml/explainers/ICETransformer.py
mhamilton723: This still seems to be affected

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
mhamilton723: This still seems unresolved

mhamilton723 (changes requested):
A lot of the comments that are marked as resolved don't seem to be resolved perhaps I'm missing something or commenting too early

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -0,0 +1,278 @@
...
+      case `individualKind` =>
+        dependenceDfs.reduceOption(_.join(_, Seq(idCol), "inner"))
+          .map(sampled.join(_, Seq(idCol), "inner").drop(idCol)).get
+      case `averageKind` =>
+        dependenceDfs.reduce(_ crossJoin _)
+    }
+  }
+
+  private def collectCategoricalValues[_](df: DataFrame, feature: ICECategoricalFeature): Array[_] = {
+    val featureCount = DatasetExtensions.findUnusedColumnName("__feature__count__", df)
```
mhamilton723, line 167: might want to rename this to featureCount to be consistent with other added columns

mhamilton723 (approved):

PR #1286, opened 2021-12-06
feat: Allow FlattenBatch to propagate non-array values

### LINKED ISSUE #1285: Support propagating non-array values in FlattenBatch
**Is your feature request related to a problem? Please describe.**
When creating batches of data (e.g. with `FixedMiniBatchTransformer`, the result is a DataFrame where the values in each column is an array. The `FlattenBatch` transformer is happy to work with this format to de-batch a DataFrame.

However, when running with a Cognitive Services transformer (e.g. `TextAnalyze`) the resulting DataFrame has array values in each column _except_ for the error column, which contains either `undefined` or an error value. The scenario that I'm currently working with requires that we can de-batch and keep the error for any rows associated with a batch that had an error.

Currently, this requires us to split the DataFrame after the Cognitive Services transform into success and error DataFrames, apply de-batching, and finally join the results together.

**Describe the solution you'd like**
I'd like `FlattenBatch` to add a `propagateNonArrayValues()` option that would treat a non-array value in a row in the input DataFrame as a value that should be copied/propagated across all output rows that correspond to the input row.

e.g. with this input

| file | key_phrases | error | 
|-|-|-|
|["file1", "file2"]| [ ["hello", "world"], ["Welcome to Seattle"]] | undefined|
|["file3", "file4"] | undefined | {"code": 400}|

the result of `FlattenBatch` would be

| file | key_phrases | error | 
|-|-|-|
|"file1"| ["hello", "world"] | undefined|
|"file2"| ["Welcome to Seattle"] | undefined|
|"file3" | undefined | {"code": 400}|
| "file4" | undefined | {"code": 400}|

stuartleeks:
Closes #1285

### DIFF
--- a/core/src/main/scala/com/microsoft/azure/synapse/ml/stages/MiniBatchTransformer.scala
+++ b/core/src/main/scala/com/microsoft/azure/synapse/ml/stages/MiniBatchTransformer.scala
@@ -189,20 +189,23 @@ class FlattenBatch(val uid: String)
 
   def this() = this(Identifiable.randomUID("FlattenBatch"))
 
-  def transpose(nestedSeq: Seq[Seq[Any]]): Seq[Seq[Any]] = {
+  def transpose(nestedSeq: Seq[Any]): Seq[Seq[Any]] = {
 
     val innerLength = nestedSeq.filter {
       case null => false
-      case _ => true
-    }.head.length
+      case _: Seq[Any] => true
+      case _ => false
+    }.head.asInstanceOf[Seq[Any]].length
 
     assert(nestedSeq.forall{
       case null => true
-      case innerSeq => innerSeq.lengthCompare(innerLength) == 0
+      case innerSeq: Seq[Any] => innerSeq.lengthCompare(innerLength) == 0
+      case _ => true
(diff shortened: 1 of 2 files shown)


core/src/test/scala/com/microsoft/azure/synapse/ml/core/test/fuzzing/Fuzzing.scala
```diff
@@ -213,9 +213,9 @@ trait ExperimentFuzzing[S <: PipelineStage] extends TestBase with DataFrameEqual
     }
   }
 
-  test("Experiment Fuzzing") {
-    testExperiments()
-  }
+  // test("Experiment Fuzzing") {
```
mhamilton723, line 216: You can just delete these. Thanks for finding this!

stuartleeks: I've dropped these changes out as they were just a temporary test-speed up for my local dev cycle

mhamilton723 (approved):

PR #1326, opened 2022-01-03
fix: lazily creating a tempdir in NativeLoader

ocworld:
This pr is recreated. origin pr is https://github.com/microsoft/SynapseML/pull/945

---
When predicting with a lightgbm model, there are so many lightgbm's so files on runtime.
It made failure on allocating memory on tmpfs in my application.

So, I tried to use java.library.path that has already "so" files.
When trying this approach, I found that it is possible to create so many /tmp/mml-nativesXXXX directories. Though, copied "so" files were not used.

Therefore, this pr is created to create lazily tempdir in NativeLoader to using getOrCreateTempDir() instead of making it in class's initializer.


ocworld:
@mhamilton723 Semantic Pull Request is failed T.T How to fix it?

mhamilton723 (approved):

PR #1335, opened 2022-01-08
fix: Correct synpase typo

iemejia:


mhamilton723:
Thank you :) @iemejia

PR #1349, opened 2022-01-17
Log async call duration and include final status in retries exceeded error

stuartleeks:
- Add setAsyncCompletionLogMessagePrefixValue to enable logging of async operations on completion (with duration)
- When reporting the retries exceeded error, include the last retrieved status to aid debugging

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/ComputerVision.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/ComputerVision.scala
@@ -209,11 +209,11 @@ object RecognizeText extends ComplexParamsReadable[RecognizeText] {
   }
 }
 
-trait BasicAsyncReply extends HasAsyncReply {
+trait BasicAsyncReply extends HasAsyncReply with BasicLogging {
 
   protected def queryForResult(key: Option[String],
                                client: CloseableHttpClient,
-                               location: URI): Option[HTTPResponseData] = {
+                               location: URI): Either[String, HTTPResponseData] = {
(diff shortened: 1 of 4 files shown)


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/ComputerVision.scala
```diff
@@ -261,22 +266,36 @@ trait BasicAsyncReply extends HasAsyncReply {
...
+              }
+            }
+            queryForResult(key, client, location)
+          }
+        }
+      })
+      it match {
+        case Right(value) =>{
+          val asyncCompletionLogMessagePrefixValue = getAsyncCompletionLogMessagePrefix
+          if (asyncCompletionLogMessagePrefixValue != "") {
```
mhamilton723, line 285: Dont need to use "" to connote param not being set. get(paramName) will return none if param isn't set and can be used to turn off setting. However, it might be good to ask the broader existential question of why a user would need to set the log prefix manually. Are you trying to parse logs in a unified way? If so perhaps we can take a page from the messages output in logClass and logTransform and output a nice json payload that log parsers can inspect and display reasonably

stuartleeks: Simplified as we discussed

stuartleeks:
@mhamilton723 - I've just pushed a new commit that simplifies the approach in the PR based on my understanding of our conversation

stuartleeks:
@mhamilton723 - are you able to review this?

PR #1354, opened 2022-01-21
fix: missing org.json4s when saving model

dciborow:
Apache Spark 3.1
Python 3.8
Scala 2.12.10
Java 1.8.0_282

image

The above error was encountered while trying to save LightGBM on Azure Synapse.


imatiach-msft (approved):

mhamilton723 (changes requested):
Discussed offline, i dont think this is the right fix as it is probably a spark version mismatch

imatiach-msft (changes requested):
oh ok.  I'm removing my approval then :sweat_smile:

imatiach-msft:
@dciborow can this PR be closed?  I think @mhamilton723 is correct it isn't needed.

PR #781, opened 2020-01-23
WIP: Integration of LinkedIn's Isolation Forest

eisber:

PR #1361, opened 2022-01-25
fix: remove verbose CMS output and convert scoredLabelsCol to DoubleType

memoryz:


imatiach-msft:
nice, seems it was added as part of this PR by accident:
https://github.com/microsoft/SynapseML/pull/781/files#diff-ee0c252ce6e62822e495c7bdbf104068892335c8ee0f39bb8e8003159eddf02aL290

imatiach-msft (approved):

core/src/main/scala/com/microsoft/azure/synapse/ml/train/ComputeModelStatistics.scala
```diff
@@ -281,7 +281,7 @@ class ComputeModelStatistics(override val uid: String) extends Transformer
     // Calculate confusion matrix and output it as DataFrame
     // TODO: We call cache in order to avoid a bug with catalyst where CMS seems to get stuck in a loop
     // For future spark upgrade past 2.2.0, we should try to see if the cache() call can be removed
-    dataset.select(col(scoredLabelsColumnName), col(labelColumnName))
+    dataset.select(col(scoredLabelsColumnName).cast(DoubleType), col(labelColumnName))
       .cache()
```
imatiach-msft, line 285: might be interesting to see if this cache call can indeed finally be removed, I remember this was a pretty nasty bug

memoryz: Since I don't know anything about this bug, I'm not sure how to validate it. Do you have any ideas what this bug is?

PR #1362, opened 2022-01-25
fix: Make SynapseE2E Tests work now with Spark 3.2

riserrad:
## Problem

SynapseE2E tests likely stopped passing after two events:

- it started building & testing SynapseML 0.9.5, given it only supports Spark 3.2.
  - The reason is that the Synapse Analytics Workspace it was using to run the Jobs only had Apache Spark Pools using Spark 3.1.
- the notebooks it was using to submit the jobs were moved from the `/notebooks` folder to subfolders in `/notebooks/features`, and the current code did not support fetching files from folders recursively.

## Changes

Given these issues, the following changes are presented in this PR:

- `FileUtilities.scala` now presents a function to read files in a folder recursively. This function is later consumed by `SynapseUtilities.scala`;
- In `SynapseTests.scala`:
  - Made the tests now submit jobs to the workspace that supports Spark 3.2 (private preview);
  - To make sure the tests succeed and run in a timely fashion, it is now using 5 pools instead of 3.
    - Currently, since it has 13 Jobs, it requires 3 batches to complete. If we want this to run even faster, we could add 2 more pools and make it run in 2 baches only.
- In `SynapseUtilities.scala`:
  - Fixed the `livyPayload` to install `synapseml_2.12` instead of `synapseml`;
  - Fixed the `livyPayload` to also exclude `org.slf4j:slf4j-api` when configuring the session;
  - Increased the test timeout from 20 to 30 minutes - made this change by observing the average time jobs were taking and 20 minutes seemed not to be enough. Empirically, 30 minutes worked well.

Additionally, as I was going through this, I noticed some opportunities for improvement in `developer-readme`:

- Added some links to additional software it needs, and for a newcomer it might not be obvious to have installed (e.g., JDK 11, Miniconda);
- Added additional links to guidance that might be useful for folks getting started, such as "Forking a repo" and "Working with remotes";
- Added an additional step to prepare the Python/Conda environment
  - This one might be needed to run tests manually, run `nbconvert` and all that.

## Results

After these changes were pushed to the fork, we were able to see consistent success in the Pipeline execution.

## Opportunities

There are still opportunties for improvements here, but I'd rather have them in a separate PR/effort, such as:

- Understand the dependency on Azure CLI when setting up the local development environment.
  - Some folks went through issues when setting it up, when not having the cli installed/configured.
- Have a separate test result for each notebook run in SynapseE2E
  - Today, if any job fails in SynapseE2E, it's hard to identify which job(s) failed for troubleshooting.
  - We can leverage Scala's runtime test generation for that.
- Manage the Desired State Configuration of the SynapseE2E Workspace in code
  - It is risky not to store the Desired State Configuration of this workspace our SynapseE2E tests use to run the tests, since if anyone makes a change to it (delete it, deletes a pool, etc), tests are going to break.

### DIFF
--- a/core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/SynapseTests.scala
+++ b/core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/SynapseTests.scala
@@ -45,7 +45,7 @@ class SynapseTests extends TestBase {
         val poolName = SynapseUtilities.monitorPool(workspaceName, sparkPools)
         val livyUrl = "https://" +
           workspaceName +
-          ".dev.azuresynapse.net/livyApi/versions/2019-11-01-preview/sparkPools/" +
+          ".dev.azuresynapse-dogfood.net/livyApi/versions/2019-11-01-preview/sparkPools/" +
           poolName +
           "/batches"
         val livyBatch: LivyBatch = SynapseUtilities.uploadAndSubmitNotebook(livyUrl, f)
--- a/core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/SynapseTests.scala
+++ b/core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/SynapseTests.scala
@@ -81,4 +81,22 @@ class SynapseTests extends TestBase {
         throw t
     }
   }
+
+  test("listPythonFiles") {
+    val allPythonFiles = SynapseUtilities.listPythonFiles()
+
+    assert(allPythonFiles.length == 0)
+  }
+
+  test("listNoteBookFiles") {
+    val allPythonNotebooks = SynapseUtilities.listNoteBookFiles()
+
+    assert(allPythonNotebooks.length > 0)
+  }
+
+  test("listPythonJobFiles") {
+    val allPythonJobFiles = SynapseUtilities.listPythonJobFiles()
+
+    assert(allPythonJobFiles.length == 0)
+  }
 }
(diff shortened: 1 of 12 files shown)


svotaw:
@svotaw is added to the review. #Closed

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/SynapseTests.scala
```diff
@@ -78,4 +85,22 @@ class SynapseTests extends TestBase {
...
+    val allPythonNotebooks = SynapseUtilities.listNoteBookFiles()
+
+    allPythonNotebooks.foreach(file => println(file))
+  }
+
+  test("listPythonJobFiles") {
+    val allPythonJobFiles = SynapseUtilities.listPythonJobFiles()
+
+    allPythonJobFiles.foreach(file => println(file))
+  }
```
mhamilton723, line 105: Are these tests that you would like to keep or something that you use for debugging? If the former we might want to add some sort of assert in here so that we can ensure they don't break. If the latter we might want to set them to "ignore" with a comment

website/src/pages/index.js
```diff
@@ -287,7 +287,7 @@ function Home() {
 {
   "name": "synapseml",
   "conf": {
-      "spark.jars.packages": "com.microsoft.azure:synapseml_2.12:0.9.5",
+      "spark.jars.packages": "com.microsoft.azure:synapseml_2.12:0.9.4",
```
mhamilton723, line 290: I think this clarification was already added to website

mhamilton723 (changes requested):
Fantastique! Just minor comments

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/SynapseTests.scala
riserrad: I think it is now fine to remove them. Thanks for catching this one!

website/src/pages/index.js
riserrad: Very true!

Issue #1336, opened 2022-01-09
LightGBMRanker not repartition by grouping column when using BarrierExecutionMode 


liyzcj:
**Describe the bug**

when using BarrierExecutionMode, LightGBMRanker repartition again after repartition by grouping column, this will reshuffle the data and the final result is not repartition by grouping column.

**To Reproduce**

```scala
  override def prepareDataframe(dataset: Dataset[_], trainingCols: Array[(String, Seq[DataType])],
                                numWorkers: Int): DataFrame = {
    if (getRepartitionByGroupingColumn) {
      val repartitionedDataset = getOptGroupCol match {
        case None => dataset
        case Some(groupingCol) => {
          val df = dataset.repartition(new Column(groupingCol)).cache()
          //force materialization
          df.count
          df
        }
      }
      super.prepareDataframe(repartitionedDataset, trainingCols, numWorkers)
    } else {
      super.prepareDataframe(dataset, trainingCols, numWorkers)
    }
  }
```

we can see in LightGBMRanker.scala, the `prepareDataframe` method will call `super.prepareDataframe` after repartition by grouping column.

and in LightGBMBase.scala, this will repartition again when `getUseBarrierExecutionMode`:

```
    if (getUseBarrierExecutionMode) {
      val numPartitions = df.rdd.getNumPartitions
      if (numPartitions > numWorkers) {
        df.repartition(numWorkers)
      } else {
        df
      }
    } else {
      df.coalesce(numWorkers)
    }
```

this will shuffle the data again.

**Expected behavior**

LightGBMRanker should repartition by grouping column.

**Info (please complete the following information):**
 - SynapseML Version: v1.0.0-rc4

imatiach-msft:
@liyzcj  This does sound like a problem if you use barrier execution mode in combination with grouping column.  Perhaps the code could be changed to run repartitioning just once when using both, but somehow it needs to ensure that the same groups are in the same partition.  Using the example from this link:
https://stackoverflow.com/questions/58286502/spark-repartitioning-by-column-with-dynamic-number-of-partitions-per-column

"If our dataset contains 80 people from China, 15 people from France, and 5 people from Cuba, then we'll want 8 partitions for China, 2 partitions for France, and 1 partition for Cuba."

[...]

PR #1368, opened 2022-01-26
fix: repartition bug in LightGBMRanker

liyzcj:
#1336 fix repartition bug in LightGBMRanker


imatiach-msft (approved):

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMRanker.scala
```diff
@@ -93,13 +93,24 @@ class LightGBMRanker(override val uid: String)
     if (getRepartitionByGroupingColumn) {
       val repartitionedDataset = getOptGroupCol match {
         case None => dataset
-        case Some(groupingCol) => {
-          val df = dataset.repartition(new Column(groupingCol)).cache()
-          //force materialization
-          df.count
-          df
-        }
+        case Some(groupingCol) =>
```
imatiach-msft, line 96: I recall this logic was modified in this PR from another user for the ranker:
https://github.com/microsoft/SynapseML/pull/778
who had issues with this logic, so I would just be careful here

PR #1373, opened 2022-01-31
docs: Update Developer Readme

dciborow:
docs: Update Developer Readme to include Java and Winutils Setup directions.

### DIFF
--- a/website/docs/reference/developer-readme.md
+++ b/website/docs/reference/developer-readme.md
@@ -10,7 +10,7 @@ description: SynapseML Development Setup
 ## Windows
 1) Install JDK 11
     - https://docs.microsoft.com/en-us/java/openjdk/download#openjdk-11
-1)Configure Hadoop "winutils"
+1) Configure Hadoop "winutils"
     - Download from https://github.com/cdarlint/winutils/tree/master/hadoop-3.2.2/bin
     - Create new folder, "C:\Program Files\Hadoop\bin", and place winutils.exe and hadoop.dll inside
     - Add HADOOP_HOME="C:\Program Files\Hadoop" to Env Vars
--- a/website/docs/reference/developer-readme.md
+++ b/website/docs/reference/developer-readme.md
@@ -6,9 +6,11 @@ description: SynapseML Development Setup
 ---
 
 # SynapseML Development Setup
+
+## Windows
 1) Install JDK 11
     - https://docs.microsoft.com/en-us/java/openjdk/download#openjdk-11
-1) On Windows, Configure Hadoop "winutils"
+1)Configure Hadoop "winutils"
     - Download from https://github.com/cdarlint/winutils
     - Create new folder, for exaxmple "C:\Hadoop". Create "C:Hadoop\bin", and place winutils.exe inside
     - Set winutils.exe properties to "Allow Execution"
@@ -16,7 +18,7 @@ description: SynapseML Development Setup
     - Add HADOOP_HOME to Path
     - Restart Computer
 1) [Install SBT](https://www.scala-sbt.org/1.x/docs/Setup.html)
-1) Install Python via Anaconda
+1) Install Python 3.8 via Anaconda
     -  Full - https://docs.anaconda.com/anaconda/install/windows/
     -  Small - https://docs.conda.io/en/latest/miniconda.html#windows-installers
 1) Fork the repository on github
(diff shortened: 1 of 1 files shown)


dciborow:
Forgot to add steps for Python

website/docs/reference/developer-readme.md
```diff
@@ -6,20 +6,27 @@ description: SynapseML Development Setup
 ---
 
 # SynapseML Development Setup
-
+1) Install JDK 11
+    - https://docs.microsoft.com/en-us/java/openjdk/download#openjdk-11
+1) On Windows, Configure Hadoop "winutils"
+    - Download from https://github.com/cdarlint/winutils
```
mhamilton723, line 12: please link to a specific file for simplicity

website/docs/reference/developer-readme.md
```diff
@@ -6,20 +6,27 @@ description: SynapseML Development Setup
...
+    - Create new folder, for exaxmple "C:\Hadoop". Create "C:Hadoop\bin", and place winutils.exe inside
```
mhamilton723, line 13: Consider using `C:\Program Files\Hadoop\bin` explicitly here

website/docs/reference/developer-readme.md
```diff
@@ -6,20 +6,27 @@ description: SynapseML Development Setup
...
+    - Set winutils.exe properties to "Allow Execution"
```
mhamilton723, line 14: I didnt have to do this

website/docs/reference/developer-readme.md
```diff
@@ -6,20 +6,27 @@ description: SynapseML Development Setup
...
+    - Add HADOOP_HOME="C:\Hadoop" to Env Vars
+    - Add HADOOP_HOME to Path
```
mhamilton723, line 16: Dont need this step

website/docs/reference/developer-readme.md
```diff
@@ -6,20 +6,27 @@ description: SynapseML Development Setup
...
+    - Restart Computer
```
mhamilton723, line 17: Just need to restart intelliJ

website/docs/reference/developer-readme.md
```diff
@@ -6,22 +6,35 @@ description: SynapseML Development Setup
 ---
 
 # SynapseML Development Setup
-
+1) Install JDK 11
+    - https://docs.microsoft.com/en-us/java/openjdk/download#openjdk-11
+1) On Windows, Configure Hadoop "winutils"
```
imatiach-msft, line 11: I think this guide would be MUCH better if we had separate sections for Windows, Linux and MacOS.  Trying to follow these steps and keeping in mind what is windows vs linux vs macos is going to be hard.  Also, we should add some HTML to collapse the sections.  Take a look at the lightgbm guide for reference, it is much better than this:
https://github.com/microsoft/LightGBM/blob/master/docs/Installation-Guide.rst
https://lightgbm.readthedocs.io/en/latest/Installation-Guide.html

Especially the html - it has collapsed sections for different types of installations users may want to do.

imatiach-msft: Also, instead of putting this in developer-readme.md, it would be good if we just had a separate installations file.  But it might be out of scope for this PR.

dciborow: draft of linux steps stubbedout.

website/docs/reference/developer-readme.md
dciborow: image
let me investigate

website/docs/reference/developer-readme.md
```diff
@@ -7,21 +7,45 @@ description: SynapseML Development Setup
 
 # SynapseML Development Setup
 
+## Windows
+1) Install JDK 11
+    - https://docs.microsoft.com/en-us/java/openjdk/download#openjdk-11
+1) Configure Hadoop "winutils"
+    - Download from https://github.com/cdarlint/winutils/tree/master/hadoop-3.2.2/bin
+    - Create new folder, "C:\Program Files\Hadoop\bin", and place winutils.exe and hadoop.dll inside
+    - Add HADOOP_HOME="C:\Program Files\Hadoop" to Env Vars
```
KeerthiYandaOS, line 16: Is it a good idea to configure Hadoop along with winutils.exe and hadoop.dll: https://hadoop.apache.org/releases.html

dciborow: Previously, hadoop DDL was not required... but with the latest release Mark and I were having trouble getting this to work without it.

KeerthiYandaOS: True, Hadoop dll is required. but my thought was to see if it is a better idea to set up Hadoop itself instead of just winutils and dll file.

PR #1382, opened 2022-02-07
feat: merge SynapseE2E test into Spark3.1 branch and test

serena-ruan:


mhamilton723:
@serena-ruan i think these changes include some of Ric's. We might need to drop that commitand instead of rebasing on master just layer commits on top of 0.9.5 commit

serena-ruan:
> @serena-ruan i think these changes include some of Ric's is this intentional so that we get the synapse tests?

Yes, just want to make the synapse tests work again to make sure this branch is totally fine.

PR #1401, opened 2022-02-12
feat: Add SAR+ for PySpark

dciborow:
Instead of using Scala, use Python version when using SAR with PySpark.


build.sbt
```diff
@@ -161,6 +162,11 @@ packageSynapseML := {
          |    ],
          |    zip_safe=True,
          |    package_data={"synapseml": ["../LICENSE.txt", "../README.txt"]},
+         |    extras_require={
+         |      "required": [
+         |        "recommenders==1.0.0",
```
mhamilton723, line 167: do we need this dep if we also dep on sarplus by markus? i think recommenders I pretty large

dciborow: let me confirm... but i believe recommenders has the PySpark bindings.

Is the concern recommenders? or its deps? I can reach out to reco about having a "recommenders[sarplus]" extras so we only get the exact python deps that we need?

build.sbt
```diff
@@ -34,6 +34,7 @@ val extraDependencies = Seq(
   "org.apache.httpcomponents" % "httpclient" % "4.5.6",
   "org.apache.httpcomponents" % "httpmime" % "4.5.6",
   "com.linkedin.isolation-forest" %% "isolation-forest_3.2.0" % "2.0.8"
```
dciborow, line 36:
```suggestion
  "com.linkedin.isolation-forest" %% "isolation-forest_3.2.0" % "2.0.8",
```

build.sbt
```diff
@@ -160,7 +161,7 @@ packageSynapseML := {
          |        "Programming Language :: Python :: 3",
          |    ],
          |    zip_safe=True,
-         |    package_data={"synapseml": ["../LICENSE.txt", "../README.txt"]},
+         |    package_data={"synapseml": ["../LICENSE.txt", "../README.txt"]}
```
dciborow, line 164:
```suggestion
         |    package_data={"synapseml": ["../LICENSE.txt", "../README.txt"]},
```

PR #1406, opened 2022-02-18
build: Move from AzDO Pipelines to GitHub Workflows

riserrad:
AzDO Pipelines are not visible to the public, and the OSS Team at Microsoft now recommends using GitHub Workflows over Azure Pipelines.

This PR now introduces 11 new Workflows that represent the full internal microsoft.SynapseML Pipeline that we have.

Let's make sure we squash this while merging. There are 100+ commits in here :-)


mhamilton723 (approved):

mhamilton723:
This shit is unreal!!

PR #1426, opened 2022-02-24
feat: PDP-based feature importance

ezherdeva:
# Summary

Added PDP-based feature importance as an option for the ICEExplainer. If a user passes kind == feature, it returns dependencies for specified features calculated using formulas from https://christophm.github.io/interpretable-ml-book/pdp.html#pdp-based-feature-importance

# Tests

Added test verifies the expected shape of the output and view.

# Dependency chances
N/A


core/src/test/scala/com/microsoft/azure/synapse/ml/explainers/split1/ICEExplainerSuite.scala
```diff
@@ -53,6 +53,30 @@ class ICEExplainerSuite extends TestBase with TransformerFuzzing[ICETransformer]
...
+    .setNumericFeatures(Array(ICENumericFeature("col4", Some(5))))
+    .setTargetClasses(Array(1))
+    .setKind("feature")
+  lazy val outputFeat: DataFrame = iceFeat.transform(data).cache()
+  // Output schema: number_of_features rows * 2 cols (name of the feature + corresponding dependence)
+  // Output is sorted by dependence in descending order
+  // For this example you should expect:
+  //  outputFeat.show(false)
+ //  +------------------------+---------------+
+//  |dependence              |featureNames   |
```
mhamilton723, line 70: if featureNames is an output column perhaps we should provide a getter/setter and treat this and other columns like it as a parameter. featureNames can be the default value of this param if not set

mhamilton723 (changes requested):
Looks really nice, just one nit around featureNamesCol as a param

memoryz (dismissed):
Looks good!

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -55,16 +60,38 @@ trait ICEFeatureParams extends Params with HasNumSamples {
...
+      "Allowed values are \"average\" for PDP, \"individual\" for ICE and \"feature\" for PDP-based " +
+      "feature importance.",
+    ParamValidators.inArray(Array(averageKind, individualKind, featureKind))
   )
 
   def getKind: String = $(kind)
   def setKind(value: String): this.type = set(kind, value)
 
+  // Output column names for PDP-feature-importance case (kind == `feature`)
+  val outputColFeatureNames = new Param[String] (
```
mhamilton723, line 72: nit: can we just call this featureNameCol?

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -55,16 +60,38 @@ trait ICEFeatureParams extends Params with HasNumSamples {
...
+    this,
+    "outputColFeatureNames",
+    "Output column name which corresponds to names of the features used in calculation" +
+      " of PDP-based-feature-importance option (kind == `feature`)"
+  )
+  def getOutputColFeatureNames: String = $(outputColFeatureNames)
+  def setOutputColFeatureNames(value: String): this.type = set(outputColFeatureNames, value)
+
+  val outputColDependenceName = new Param[String] (
```
mhamilton723, line 81: nit: can we just call this dependenceNameCol

memoryz (approved):

PR #1432, opened 2022-03-02
fix: add retry to reduce flakiness

SudhindraKovalam:
# Summary
- The `GET` flood-line user-data in the `GeospatialServices-FloodingRisk` notebook can be flaky sometimes.
Adding simple retry
# Tests
N/A
# Dependency chances
N/A


mhamilton723 (changes requested):
Discussed over teams, i think the fix will be to make a safe_send type function and wrap all requests calls in it

mhamilton723:
I fixed this already!

PR #1440, opened 2022-03-14
feat: Upgrade the univariate anomaly detection version to v1.1-preview

yalaudah:
This PR upgrades the version of the univariate anomaly detection API endpoint to `v1.1-preview.1` to add support for an optional parameter `imputeMode` in the input, and expect severity scores in the output.


mhamilton723 (approved):
Thank you for this awesome contribution!

PR #1444, opened 2022-03-18
feat: Refactoring Param system and adding new ones

svotaw:
# Summary

Modifications to parameters for LightBGM and Vowpal Wabbit APIs
1. Create common unit testable utilities (core/utils/ParamsStringBuilder) for creating native library command-line argument strings
2. Refactor VowpalWabbit and LightGBM to use the utilities
3. Add more explicit setters for some requested LightGBM parameters
4. Add ability for LightBGM user to set custom args that aren't defined explicitly by SynapseML (as Vowpal Wabbit does now)

# Tests
Added unit tests for new arg string generation utilities


lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMClassifier.scala
```diff
@@ -36,28 +36,30 @@ class LightGBMClassifier(override val uid: String)
...
-    val modelStr = if (getModelString == null || getModelString.isEmpty) None else get(modelString)
-    ClassifierTrainParams(getParallelism, get(topK), getNumIterations, getLearningRate,
-      get(numLeaves), get(maxBin), get(binSampleCount), get(baggingFraction), get(posBaggingFraction),
-      get(negBaggingFraction), get(baggingFreq), get(baggingSeed), getEarlyStoppingRound, getImprovementTolerance,
-      get(featureFraction), get(maxDepth), get(minSumHessianInLeaf), numTasks, modelStr,
-      getIsUnbalance, getVerbosity, categoricalIndexes, actualNumClasses, getBoostFromAverage,
-      getBoostingType, get(lambdaL1), get(lambdaL2), get(isProvideTrainingMetric),
-      get(metric), get(minGainToSplit), get(maxDeltaStep), getMaxBinByFeature, get(minDataInLeaf), getSlotNames,
-      getDelegate, getDartParams, getExecutionParams(numTasksPerExec), getObjectiveParams)
+    ClassifierTrainParams(
```
mhamilton723, line 46: love how much cleaner this is here

out/production/SynapseML/com/microsoft/azure/synapse/ml/automl/FindBestModel.txt
```diff
@@ -0,0 +1 @@
+Evaluates and chooses the best model from a list of models.
```
mhamilton723, line 1: I think this out/production folder might have been checked in by accident

vw/src/main/scala/com/microsoft/azure/synapse/ml/vw/VowpalWabbitBase.scala
```diff
@@ -75,134 +73,81 @@ trait VowpalWabbitBase extends Wrappable
...
-  val args = new Param[String](this, "args", "VW command line arguments passed")
-  setDefault(args -> "")
+  // (update: this is a removed feature as of 3.0, so would have to use newer macro mechanisms)
+  val passThroughArgs = new Param[String](this, "args", "VW command line arguments passed")
+  setDefault(passThroughArgs -> "")
+  def getPassThroughArgs: String = $(passThroughArgs)
+  def setPassThroughArgs(value: String): this.type = set(passThroughArgs, value)
 
-  def getArgs: String = $(args)
+  @deprecated("Please use 'getPassThroughArgs'.", since="1.0.0-RC3")
```
mhamilton723, line 82: we should deprecate this in since 0.9.6 or something as they haven't been deprecated until now

mhamilton723 (changes requested):
This is awesome! Minor question on "out" folder and also we should send this by ilya and Markus C too if possible

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/BaseTrainParams.scala
```diff
@@ -0,0 +1,259 @@
...
+
+/** Defines the Dataset parameters passed to the LightGBM classifier.
+  *
+  * @param isEnableSparse Used to enable/disable sparse optimization.
+  * @param useMissing Set this to false to disable the special handle of missing value.
+  * @param zeroAsMissing Set to true to treat all zero as missing values
+  *                         (including the unshown values in LibSVM/sparse matrices)
+  *                      Set to false to use na for representing missing values.
+  */
+case class DatasetParams(isEnableSparse: Option[Boolean],
```
imatiach-msft, line 141: This is awesome!  Can you add tests for all of these new parameters?  It's not always as simple as adding them here, and sometimes the distributed version of lightgbm does not support them (like forcedsplits_filename, which apparently only works in single node case https://github.com/Microsoft/LightGBM/blob/master/docs/Parameters.rst#forcedsplits_filename and I only found out the hard way after adding it everywhere in scala but then realizing it just didn't do anything for distributed case).

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMClassifier.scala
imatiach-msft: this looks so much better!  nice work!

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/BaseTrainParams.scala
svotaw: I added some tests for the new sparse parameters.  I also did some code analysis of the LightGBM code, and I don't see any limitations of any of the new parameters.  When looking at the one you mentioned (forcedsplits_filename), it does indeed show in the validation code that you can't run it under the types of parallelism we use (data and voting).

I'm not sure I see the utility of keeping those tests.  They take time, and once we know it works it should work always (if they ever make a breaking change we would break anyways). Can I just remove them?  They weren't fancy, they just proved that LightGBM had the right value for the parameters and that the fit succeeded.

imatiach-msft: "I'm not sure I see the utility of keeping those tests."
It's useful to make sure that:
1.) The parameters actually do work right now, and with future code changes they will still work
2.) With new native lightgbm jar version updates in the future they still work, and names haven't changed, etc
3.) They serve as a good example for others trying to understand how the parameters work and what effect they have by looking at the test cases
At least that is my view anyway.  Without tests we don't really know if they work for sure.

mhamilton723 (approved):

PR #1456, opened 2022-03-26
fix: Fix reflective class loading in IntelliJ

svotaw:
# Summary

Some tests rely on the ability to reflect all classes in the synpaseml package (e.g. to filter to a set of all classes derived from a particular parent class).  The current implementation uses ClassPath.from() from a spark utility package.  However, this utility only works for ClassLoaders derived from UrlClassLoader (not well documented limitation).  In the context of sbt, this is true, so things work in pipelines and such. In IntelliJ, the ClassLoaders are the standard App/Ext/Boot trio, which are not, so any test using this pathway fails since ClassPath.from() returns an empty list.  

This PR adds a double check so that if the reflected class list from ClassPath comes back empty, it falls back to a manual lookup. This lookup is slower, so the choice was to not replace it, but instead add a path that should only run when a test is run from inside IntelliJ.  Caching was added so the lookup is only done once in either case.

# Tests

Some tests that were failing before in IntelliJ now work (used first FuzzingTest as a canary).  Due to the nature of the fix, there is no way to test this in the pipelines since they use sbt.

# Dependency chances

N/A

### DIFF
--- a/core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
+++ b/core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
@@ -8,9 +8,8 @@ import com.microsoft.azure.synapse.ml.codegen.Wrappable
 import java.lang.reflect.Modifier
 import org.sparkproject.guava.reflect.ClassPath
 
-import java.io.{File, InputStreamReader, IOException}
+import java.io.{File, IOException}
 import scala.collection.JavaConverters._
-import scala.collection.mutable.ListBuffer
 import scala.reflect.{ClassTag, classTag}
 
 /** Contains logic for loading classes. */
@@ -115,7 +114,7 @@ object JarLoadingUtils {
         case f if f.isDirectory => findClassesInDirectory(f, packageName + "." + file.getName)
         case f if f.getName.endsWith(".class") =>
           Seq[Class[_]](Class.forName(packageName + '.' + file.getName.substring(0, file.getName.length - 6)))
-        case _ => Seq[Class[_]]() // some other file, just ignore
+        case _ => Seq.empty // some other file, just ignore
       }
     }).flatten.toSeq
   }
--- a/core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
+++ b/core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
@@ -8,11 +8,15 @@ import com.microsoft.azure.synapse.ml.codegen.Wrappable
 import java.lang.reflect.Modifier
 import org.sparkproject.guava.reflect.ClassPath
 
+import java.io.{File, InputStreamReader, IOException}
 import scala.collection.JavaConverters._
+import scala.collection.mutable.ListBuffer
 import scala.reflect.{ClassTag, classTag}
 
 /** Contains logic for loading classes. */
 object JarLoadingUtils {
+
+  private var CachedAllClasses: List[Class[_]] = List[Class[_]]()
 
   def className(filename: String): String = {
     if (filename.endsWith(".class")) {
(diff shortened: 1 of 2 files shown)


lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/swig/SwigUtils.scala
```diff
@@ -31,8 +31,8 @@ abstract class ChunkedArray[T]() {
   def add(value: T): Unit
```
svotaw, line 31: Changes were made in this class because 1) we got build warnings about hidden classes due to casing, and 2) the new reflection code was confused somehow due to filenames that only differed in casing.  Corrected the casing to fix those and make a proper companion class as intended.

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
```diff
@@ -24,17 +28,8 @@ object JarLoadingUtils {
   }
 
   private[ml] val AllClasses = {
-    ClassPath.from(getClass.getClassLoader)
```
mhamilton723, line 27: I dont think you need this caching logic here because a val on an object is only computed once, you can even make it lazy if you don't want it always getting loaded

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
```diff
@@ -62,4 +57,85 @@ object JarLoadingUtils {
...
+  @throws[ClassNotFoundException]
+  @throws[IOException]
+  private def getClassesFromPackage(packageName: String): List[Class[_]] = {
+    val classLoader = Thread.currentThread.getContextClassLoader
+    assert(classLoader != null)
+
+    // ClassLoader does not expose a class list (except private vars), so scan the jar files directly
+    val path = packageName.replace('.', '/')
+    val resources = classLoader.getResources(path).asIterator()
+    val dirs = new ListBuffer[File]
```
mhamilton723, line 104: can this mutability be removed with a map on resources.asScala?

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
```diff
@@ -62,4 +57,85 @@ object JarLoadingUtils {
...
+    for (resource <- resources.asScala) {
+      dirs += new File(resource.getFile)
+    }
+
+    val classes = new ListBuffer[Class[_]]
+    for (directory <- dirs) {
```
mhamilton723, line 110: likewise could we convert this to a map?

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
```diff
@@ -62,4 +57,85 @@ object JarLoadingUtils {
...
+    * Recursive method used to find all classes in a given directory and subdirs.
+    *
+    * @param directory   The base directory
+    * @param packageName The package name for classes found inside the base directory
+    * @return The classes
+    * @throws ClassNotFoundException
+    */
+  @throws[ClassNotFoundException]
+  private def findClassesInDirectory(directory: File, packageName: String): ListBuffer[Class[_]] = {
+    val classes = new ListBuffer[Class[_]]
```
mhamilton723, line 126: can convert to map chaining or a for yield statement if you want to keep imperative paradigm without mutability

mhamilton723 (changes requested):
THis is such an difficult problem you fixed. Just a few things for mutability reduction

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/JarLoadingUtils.scala
mhamilton723: https://alvinalexander.com/scala/scala-for-loop-yield-examples-yield-tutorial/

mhamilton723 (approved):

PR #1458, opened 2022-03-30
fix: lightgbm getting stuck when empty partition is chosen as the main worker in singleDatasetMode

imatiach-msft:
# Summary

Lightgbm can get stuck when SingleDatasetMode is enabled and if an empty partition is, by random chance, chosen as the main worker in the executor.
The fix is to only allow non-empty partitions to be chosen as the main worker.

# Tests

Run the test "Verify LightGBM Classifier won't get stuck on empty partitions" in the file:
https://github.com/microsoft/SynapseML/blob/master/lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMClassifier.scala#L610
many times, by random chance it will often get stuck after several runs.

# Dependency chances

No dependency changes.

### DIFF
--- a/lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/TaskTrainingMethods.scala
+++ b/lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/TaskTrainingMethods.scala
@@ -14,16 +14,13 @@ object TaskTrainingMethods {
     * Otherwise, returns true for all tasks.
     * @param trainParams The training parameters.
     * @param log The logger.
+    * @param sharedState The shared state.
     * @return Whether the current task is enabled.
     */
   def isWorkerEnabled(trainParams: BaseTrainParams, log: Logger, sharedState: SharedState): Boolean = {
     if (trainParams.executionParams.useSingleDatasetMode) {
       // Find all workers in current JVM
-      val mainExecutorWorker = sharedState.mainExecutorWorker
-      val myTaskId = LightGBMUtils.getTaskId
-      val isMainWorker = mainExecutorWorker == myTaskId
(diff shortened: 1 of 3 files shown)


lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/TaskTrainingMethods.scala
```diff
@@ -14,16 +14,16 @@ object TaskTrainingMethods {
...
     * @param trainParams The training parameters.
     * @param log The logger.
+    * @param sharedState The shared state.
     * @return Whether the current task is enabled.
     */
   def isWorkerEnabled(trainParams: TrainParams, log: Logger, sharedState: SharedState): Boolean = {
     if (trainParams.executionParams.useSingleDatasetMode) {
+      // Note: the first valid worker with non-empty partitions sets
+      // the main executor worker, other workers read it
+      sharedState.linkMainExecutorWorker()
```
svotaw, line 24: IMHO, this is creating an undocumented side effect in what is basically a getter, and relies on the fact that this method is not called from LightGBMBase if the partition is empty. A more robust fix would call linkMainExecutionWorker in the trainLightGBM method before this is called, specifically conditioned on the partition being empty.
I realize that mainExecutionWorker needs to be initialized for this method to be called, but you could just throw if it's not.

imatiach-msft: good idea!  will update

imatiach-msft: done!

PR #1461, opened 2022-03-31
fix: remove quotation from logging 

serena-ruan:
# Summary

Remove quotation from BasicLogging so that our logs can be recorded in telemetry correctly with the existing QuotationScrubber

# Tests

Ran tests and verified loggings don't contain quotations anymore.

# Dependency changes

No dependency changes


mhamilton723 (changes requested):
I'm not sure removing quotes from this string is safe as i think this will break downstream json parsing. If we absolutely cannot use quotes in our messages due to Synapse logging rules perhaps we could consider a different encoding scheme for this info. One (not great) solution is to b64 encode messages, though keeping human readability is generally a nice thing

PR #1470, opened 2022-04-01
fix: GridSpace python API not working

memoryz:
# Summary

GridSpace python API not working:

```python
searchSpace = paramBuilder.build()
gridspace = GridSpace(searchSpace)
```
```
py4j.protocol.Py4JError: An error occurred while calling None.com.microsoft.azure.synapse.ml.automl.GridSpace. Trace:
py4j.Py4JException: Constructor com.microsoft.azure.synapse.ml.automl.GridSpace([class [Lscala.Tuple2;]) does not exist
                at py4j.reflection.ReflectionEngine.getConstructor(ReflectionEngine.java:179)
                at py4j.reflection.ReflectionEngine.getConstructor(ReflectionEngine.java:196)
                at py4j.Gateway.invoke(Gateway.java:237)
                at py4j.commands.ConstructorCommand.invokeConstructor(ConstructorCommand.java:80)
                at py4j.commands.ConstructorCommand.execute(ConstructorCommand.java:69)
                at py4j.GatewayConnection.run(GatewayConnection.java:238)
                at java.lang.Thread.run(Thread.java:748)
```

# Tests

Testing code:
```python
from synapse.ml.lightgbm import LightGBMClassifier
from synapse.ml.automl import *

model = LightGBMClassifier(objective="binary", featuresCol="features", labelCol="Bankrupt?", isUnbalance=True)

searchSpace = (
  HyperparamBuilder()
      .addHyperparam(model, model.learningRate, DiscreteHyperParam([0.001, 0.02]))
      .addHyperparam(model, model.lambdaL1, DiscreteHyperParam([0.01, 1.0]))
      .build()
)

gridSpace = GridSpace(searchSpace)
hypermap = gridSpace.space().paramMaps()
print(hypermap.next())
print(hypermap.next())
print(hypermap.next())
print(hypermap.next())
```

# Dependency chances

None.

AB#1730666

### DIFF
--- a/core/src/main/python/synapse/ml/automl/HyperparamBuilder.py
+++ b/core/src/main/python/synapse/ml/automl/HyperparamBuilder.py
@@ -3,7 +3,7 @@
 
 import sys
 
-if sys.version >= '3':
+if sys.version >= "3":
     basestring = str
 
 import pyspark
@@ -11,10 +11,12 @@ from pyspark import SparkContext
 from pyspark.ml.common import inherit_doc
 from pyspark.sql.types import *
 
+
 class HyperparamBuilder(object):
     """
     Specifies the search space for hyperparameters.
     """
+
     def __init__(self):
         ctx = SparkContext.getOrCreate()
         self.jvm = ctx.getOrCreate()._jvm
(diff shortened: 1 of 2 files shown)


core/src/main/python/synapse/ml/automl/HyperparamBuilder.py
```diff
@@ -39,50 +41,72 @@ def build(self):
...
 class GridSpace(object):
     """
     Specifies a predetermined grid of values to search through.
     """
+
     def __init__(self, paramValues):
         ctx = SparkContext.getOrCreate()
         self.jvm = ctx.getOrCreate()._jvm
-        hyperparamBuilder = self.jvm.com.microsoft.azure.synapse.ml.automl.HyperparamBuilder()
+        hyperparamBuilder = self.jvm.org.apache.spark.ml.tuning.ParamGridBuilder()
```
imatiach-msft, line 85: I'm a bit confused, I think we should still be using our HyperparamBuilder here and not spark's ParamGridBuilder?

memoryz: The reason is that GridSpace constructor accepts `Array[ParamMap]`, which is exactly what `ParamGridBuilder.build()` function returns. `HyperparamBuilder.build()` returns `Array[Param[_], Dist[_]]` instead. Technically I can still use HyperparamBuilder and convert the result, but it's only used as a local variable anyway, so I don't think it matters.

imatiach-msft: "Technically I can still use HyperparamBuilder and convert the result"

hmm I can't recall why I wrote this way in the original PR, it was 4 years ago:

https://github.com/microsoft/SynapseML/pull/168

Also I don't see any scala test for gridparam which is strange.  I do think that the intention was to always use HyperparamBuilder.  Let me check.

I also don't understand this:
```
if not isinstance(hyperparam, DiscreteHyperParam)
```
random param should be supported as well, and we get the value when adding it to the grid.  So I'm confused why it doesn't work.

imatiach-msft: hmm, ok, maybe this can be merged as is and I can try and look into whether it's possible to use HyperparamBuilder still, support RandomParam in addition to DiscreteHyperParam and also add some tests for this workflow, as I don't actually see this GridSpace tested anywhere.

imatiach-msft (approved):

imatiach-msft:
I merged the PR but I will follow-up and investigate the issues I pointed out

PR #1477, opened 2022-04-14
feat: MVAD enhancement

serena-ruan:
# Summary

1. add constraint on start/end time of MVAD to provide better debug info
2. As the training time would be quite long for large dataset, change the training step to return modelId immediately and check the model status if it's used for inference.

# Tests

Added unit tests

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
@@ -192,14 +192,34 @@ trait MADBase extends HasOutputCol
   val startTime = new Param[String](this, "startTime", "A required field, start time" +
     " of data to be used for detection/generating multivariate anomaly detection model, should be date-time.")
 
-  def setStartTime(v: String): this.type = set(startTime, v)
+  def setStartTime(v: String): this.type = {
+    try {
+      val formattedValue = DateTimeFormatter.ISO_INSTANT.format(DateTimeFormatter.ISO_INSTANT.parse(v))
+      set(startTime, formattedValue)
+    }
+    catch {
+      case e: java.time.format.DateTimeParseException =>
+        throw new IllegalArgumentException(
+          s"Start time should be ISO8601 format. e.g. 2021-01-01T00:00:00Z: ${e.toString}")
+    }
+  }
 
   def getStartTime: String = $(startTime)
 
   val endTime = new Param[String](this, "endTime", "A required field, end time of data" +
     " to be used for detection/generating multivariate anomaly detection model, should be date-time.")
 
-  def setEndTime(v: String): this.type = set(endTime, v)
+  def setEndTime(v: String): this.type = {
+    try {
+      val formattedValue = DateTimeFormatter.ISO_INSTANT.format(DateTimeFormatter.ISO_INSTANT.parse(v))
+      set(endTime, formattedValue)
+    }
+    catch {
+      case e: java.time.format.DateTimeParseException =>
+        throw new IllegalArgumentException(
+          s"End time should be ISO8601 format. e.g. 2021-01-01T00:00:00Z: ${e.toString}")
+    }
+  }
 
   def getEndTime: String = $(endTime)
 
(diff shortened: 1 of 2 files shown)


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -348,16 +373,24 @@ trait MADBase extends HasOutputCol
...
-      pw.print(DateTimeFormatter.ISO_INSTANT.format(timestamp))
-      pw.write(',')
+        // TODO: do we have to worry about locale?
+        // pw.format(Locale.US, "%f", row.get(featureIdx))
+        pw.println(row.get(featureIdx))
+      }
+      catch {
+        case e: java.time.format.DateTimeParseException =>
+          throw new IllegalArgumentException(
+            s"Timestamp column should be ISO8601 format. e.g. 2021-01-01T00:00:00Z: ${e.toString}")
```
mhamilton723, line 391:
```suggestion
            s"Timestamp column should be ISO8601 format. e.g. 2021-01-01T00:00:00Z, received: ${e.toString}")
```

pipeline.yaml
```diff
@@ -88,19 +89,6 @@ jobs:
     - template: templates/update_cli.yml
     - template: templates/conda.yml
     - template: templates/kv.yml
-    - bash: |
```
mhamilton723, line 91: We should keep this as-is, depending on publish takes too long (if everything is working correctly lol). The fix should be to publishLocal before python tests so that they can use local jars and not depend on cloud publishing

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -542,6 +631,29 @@ class DetectMultivariateAnomaly(override val uid: String) extends Model[DetectMu
   override def transform(dataset: Dataset[_]): DataFrame = {
     logTransform[DataFrame] {
 
+      // check model status first
```
mhamilton723, line 634: Good addition!

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -457,6 +490,62 @@ class FitMultivariateAnomaly(override val uid: String) extends Estimator[DetectM
...
+    val get = new HttpGet()
+    get.setURI(location)
+    key.foreach(get.setHeader("Ocp-Apim-Subscription-Key", _))
+    get.setHeader("User-Agent", s"synapseml/${BuildInfo.version}${HeaderValues.PlatformInfo}")
+    val resp = convertAndClose(sendWithRetries(client, get, getBackoffs))
+    get.releaseConnection()
+    Some(resp)
+  }
+
+  override protected def handlingFunc(client: CloseableHttpClient,
```
mhamilton723, line 505: https://github.com/microsoft/SynapseML/pull/1457/files

This PR added resiliency needed for MVAD. Could you please ensure that all the network calls flow through this function so that we can be a unified force against flaky network issues? This might already be the case but double checking this will help me sleep at night

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -192,14 +197,34 @@ trait MADBase extends HasOutputCol
   val startTime = new Param[String](this, "startTime", "A required field, start time" +
     " of data to be used for detection/generating multivariate anomaly detection model, should be date-time.")
 
-  def setStartTime(v: String): this.type = set(startTime, v)
+  def setStartTime(v: String): this.type = {
```
mhamilton723, line 200: nit: code duplication in both setters, (and elsewhere I think) consider a helper

mhamilton723 (changes requested):
Thank you for your work to make this code better!

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
serena-ruan: If you're meaning make sure FitMultivariateAnomaly's network goes inside this override function to query for result then yes, I've debugged the tests to see it go through this function. Or is there anything else you're thinking of?

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
serena-ruan: Thanks! Will refactor this part.

pipeline.yaml
serena-ruan: Ah yes of course then, I see python tests already depend on publishLocal though.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
```diff
@@ -457,6 +474,62 @@ class FitMultivariateAnomaly(override val uid: String) extends Estimator[DetectM
...
+                                      request: HTTPRequestData): HTTPResponseData = {
+    val response = HandlingUtils.advanced(getBackoffs: _*)(client, request)
+    if (response.statusLine.statusCode == 201) {
+      val location = new URI(response.headers.filter(_.name == "Location").head.value)
+      val maxTries = getMaxPollingRetries
+      val key = request.headers.find(_.name == "Ocp-Apim-Subscription-Key").map(_.value)
+      val it = (0 to maxTries).toIterator.flatMap { _ =>
+        val resp = queryForResult(key, client, location)
+        val fields = IOUtils.toString(resp.get.entity.get.content, "UTF-8").parseJson.asJsObject.fields
+        val status = if (fields.keySet.contains("modelInfo")) {
```
mhamilton723, line 499: I see that this code is duplicated twice (sorry for not catching this first time around). It might be a good idea to either pull this out into a separate method that parses this, or use spray to parse this so that its a quick 1-liner

pipeline.yaml
```diff
@@ -306,6 +306,7 @@ jobs:
       displayName: 'publish jar package to maven central'
 
 - job: PythonTests
+  dependsOn: Publish
```
mhamilton723, line 309:
```suggestion
```

mhamilton723 (changes requested):

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/MultivariateAnomalyDetection.scala
mhamilton723: Thank you!

mhamilton723 (approved):

PR #1488, opened 2022-04-22
fix: fix library checking while installing library on databricks cluster

serena-ruan:
# Summary

Should check library installation status before submitting jobs for databricksTest

# Tests

Ran Databricks test

# Dependency changes

None.

AB#1759003

### DIFF
--- a/core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/DatabricksUtilities.scala
+++ b/core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/DatabricksUtilities.scala
@@ -229,7 +229,7 @@ object DatabricksUtilities extends HasHttpClient {
     state == "RUNNING"
   }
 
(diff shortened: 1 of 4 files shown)


core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/DatabricksUtilities.scala
```diff
@@ -229,6 +229,17 @@ object DatabricksUtilities extends HasHttpClient {
     state == "RUNNING"
   }
 
+  def isLibrariesAllInstalled(clusterId: String): Boolean = {
```
mhamilton723, line 232: nit: isLibrariesAllInstalled -> areLibrariesInstalled

mhamilton723 (changes requested):
thank you! Just a nit on naming

PR #1510, opened 2022-05-12
feat: openai completions polymorphic prompts

niehaus59:
# Summary

Add support for polymorphic prompts in OpenAI/GPT3/s Completions endpoint as described here: https://github.com/Azure/OpenAI/blob/ed700c41469a799e30baedbbc13c62aa634b777f/Reference/swagger_2021_11_01.json#L64

# Tests

Added one test for each new prompt form, in OpenAISuite.scala

# Dependency changes

N/A

AB#1790206

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
@@ -1,0 +1,312 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.azure.synapse.ml.cognitive
+
+import com.microsoft.azure.synapse.ml.codegen.{GenerationUtils, Wrappable}
+import com.microsoft.azure.synapse.ml.io.http._
+import com.microsoft.azure.synapse.ml.logging.BasicLogging
+import org.apache.http.entity.{AbstractHttpEntity, ContentType, StringEntity}
+import org.apache.spark.ml.ComplexParamsReadable
+import org.apache.spark.ml.param._
+import org.apache.spark.ml.util._
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.types._
+import spray.json.DefaultJsonProtocol._
+import spray.json._
+import AnyJsonFormat.anyFormat
+import scala.language.existentials
+
+
+trait HasSetServiceName extends Wrappable with HasURL {
+  override def pyAdditionalMethods: String = super.pyAdditionalMethods + {
+    """
+      |def setDeploymentName(self, value):
+      |    self._java_obj = self._java_obj.setDeploymentName(value)
+      |    return self
+      |""".stripMargin
+  }
+
+  def setServiceName(v: String): this.type = {
+    setUrl(s"https://${v}.openai.azure.com/")
+  }
+}
+
+trait HasPrompt extends HasServiceParams {
+  val prompt: ServiceParam[String] = new ServiceParam[String](
+    this, "prompt", "The text to complete", isRequired = true)
+
+  def getPrompt: String = getScalarParam(prompt)
+
+  def setPrompt(v: String): this.type = setScalarParam(prompt, v)
+
+  def getPromptCol: String = getVectorParam(prompt)
+
+  def setPromptCol(v: String): this.type = setVectorParam(prompt, v)
+}
+
+trait HasAPIVersion extends HasServiceParams {
+  val apiVersion: ServiceParam[String] = new ServiceParam[String](
+    this, "apiVersion", "version of the api", isRequired = true, isURLParam = true) {
+    override val payloadName: String = "api-version"
+  }
+
+  def getApiVersion: String = getScalarParam(apiVersion)
+
+  def setApiVersion(v: String): this.type = setScalarParam(apiVersion, v)
+
+  def getApiVersionCol: String = getVectorParam(apiVersion)
+
+  def setApiVersionCol(v: String): this.type = setVectorParam(apiVersion, v)
+
+  setDefault(apiVersion -> Left("2022-03-01-preview"))
+}
+
+trait HasDeploymentName extends HasServiceParams {
+  val deploymentName = new ServiceParam[String](
+    this, "deploymentName", "The name of the deployment", isRequired = true)
+
+  def getDeploymentName: String = getScalarParam(deploymentName)
+
+  def setDeploymentName(v: String): this.type = setScalarParam(deploymentName, v)
+
+  def getDeploymentNameCol: String = getVectorParam(deploymentName)
+
+  def setDeploymentNameCol(v: String): this.type = setVectorParam(deploymentName, v)
+}
+
+trait HasMaxTokens extends HasServiceParams {
+
+  val maxTokens: ServiceParam[Int] = new ServiceParam[Int](
+    this, "maxTokens",
+    "The maximum number of tokens to generate. Has minimum of 0.",
+    isRequired = false)
+
+  def getMaxTokens: Int = getScalarParam(maxTokens)
+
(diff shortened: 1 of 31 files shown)


niehaus59:
>   def setPrompt(v: String): this.type = setScalarParam(prompt, v)

I tried a much simpler change where I just overloaded setPrompt with Seq[String], Seq[Int] and Seq[Seq[Int]], but I couldn't get the new forms to work. Let me know if that's something I should look further into.

---
Refers to: cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala:41 in a981b38. [](commit_id = a981b3891b963687960003c93359326dd4baa104, deletion_comment = False)

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
```diff
@@ -34,7 +34,7 @@ trait HasSetServiceName extends Wrappable with HasURL {
 
 trait HasPrompt extends HasServiceParams {
   val prompt: ServiceParam[String] = new ServiceParam[String](
-    this, "prompt", "The text to complete", isRequired = true)
+    this, "prompt", "The text to complete", isRequired = false)
```
niehaus59, line 37: >false [](http://example.com/codeflow?start=57&length=5)

I'm assuming the orElse(throw...) near the bottom of this file will handle the case where none of the prompt formats are specified.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
```diff
@@ -299,14 +339,46 @@ class OpenAICompletion(override val uid: String) extends CognitiveServicesBase(u
...
+      getValueOpt(r, prompt)
+        .map(prompt => getStringEntity(prompt, optionalParams))
+        .orElse(getValueOpt(r, bulkPrompt)
+          .map(bulkPrompt => getArrayStringEntity(bulkPrompt, optionalParams)))
+        .orElse(getValueOpt(r, indexPrompt)
+          .map(indexPrompt => getIndexStringEntity(indexPrompt, optionalParams)))
+        .orElse(getValueOpt(r, bulkIndexPrompt)
+          .map(bulkIndexPrompt => getIndexArrayStringEntity(bulkIndexPrompt, optionalParams)))
+        .orElse(throw new IllegalArgumentException(
+          "Payload needs to contain String, Seq[String], Seq[Int] or Seq[Seq[Int]]. This code should not run"))
```
niehaus59, line 351: >Payload needs to contain String, Seq[String], Seq[Int] or Seq[Seq[Int]]. [](http://example.com/codeflow?start=11&length=72)

This probably needs better wording. Suggestions?

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
```diff
@@ -45,6 +45,45 @@ trait HasPrompt extends HasServiceParams {
   def setPromptCol(v: String): this.type = setVectorParam(prompt, v)
 }
 
+trait HasBulkPrompt extends HasServiceParams {
```
mhamilton723, line 48: Is this our naming or theirs? Perhaps something like `Batch` instead of `Bulk`

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
```diff
@@ -45,6 +45,45 @@ trait HasPrompt extends HasServiceParams {
...
+  def setIndexPrompt(v: Seq[Int]): this.type = setScalarParam(indexPrompt, v)
+
+  def getIndexPromptCol: String = getVectorParam(indexPrompt)
+
+  def setIndexPromptCol(v: String): this.type = setVectorParam(indexPrompt, v)
+}
+
+trait HasBulkIndexPrompt extends HasServiceParams {
+  val bulkIndexPrompt: ServiceParam[Seq[Seq[Int]]] = new ServiceParam[Seq[Seq[Int]]](
+    this, "bulkIndex", "Sequence of index sequences to complete", isRequired = false)
```
mhamilton723, line 76: The second arg of the params should match the name of the param so this would be 
```suggestion
    this, "bulkIndexPrompt", "Sequence of index sequences to complete", isRequired = false)
```

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
```diff
@@ -45,6 +45,45 @@ trait HasPrompt extends HasServiceParams {
...
+  def setBulkPrompt(v: Seq[String]): this.type = setScalarParam(bulkPrompt, v)
+
+  def getBulkPromptCol: String = getVectorParam(bulkPrompt)
+
+  def setBulkPromptCol(v: String): this.type = setVectorParam(bulkPrompt, v)
+}
+
+trait HasIndexPrompt extends HasServiceParams {
+  val indexPrompt: ServiceParam[Seq[Int]] = new ServiceParam[Seq[Int]](
+    this, "index", "Sequence of indexes to complete", isRequired = false)
```
mhamilton723, line 63:
```suggestion
    this, "indexPrompt", "Sequence of indexes to complete", isRequired = false)
```

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
mhamilton723: The prompt field should be set with a String, Seq[String], Seq[Int] or Seq[Seq[Int]], received ${.....}.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
niehaus59: They don't really have a set name. In the swagger they call it a list and elsewhere they call it an array.  I'll make the change to Batch.

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/cognitive/split1/OpenAISuite.scala
```diff
@@ -23,34 +20,77 @@ class OpenAICompletionSuite extends TransformerFuzzing[OpenAICompletion] with Op
...
     super.assertDFEq(df1.drop("out"), df2.drop("out"))(eq)
   }
 
-  override def testObjects(): Seq[TestObject[OpenAICompletion]] =
-    Seq(new TestObject(completion, df))
+  override def testObjects(): Seq[TestObject[OpenAICompletion]] = Seq(
+    new TestObject(promptCompletion, promptDF),
+    new TestObject(bulkPromptCompletion, bulkPromptDF),
+    new TestObject(indexCompletion, indexDF),
+    new TestObject(bulkIndexCompletion, bulkIndexDF))
```
mhamilton723, line 93: Dont necessarily need all these in test objects if you don't want! This will cause all instances to generate python tests and serialization tests

mhamilton723 (changes requested):
Love it!!! Only a few minor things

core/src/main/scala/org/apache/spark/ml/param/IntArrayArrayParam.scala
```diff
@@ -0,0 +1,90 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package org.apache.spark.ml.param
+
+import scala.collection.JavaConverters._
+import org.json4s._
+import org.json4s.jackson.JsonMethods._
+
+/*object IntArrayArrayJsonProtocol extends DefaultJsonProtocol {
```
mhamilton723, line 10: This looks like debugging code

niehaus59:
Switched to a different branch and checked in there.

PR #1518, opened 2022-06-02
feat: Support deep vision

serena-ruan:
# Summary

Support common DNN models for deep vision classification.

# Tests

Added unit test.

# Dependency changes

Added requirements.txt file to explain dependencies for python package.


mhamilton723:
Giddy with excitement about this :) 

Also one QQ i see all the python style has changed, is it because you ran it through something like black?

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/DatabricksUtilities.scala
```diff
@@ -56,15 +58,26 @@ object DatabricksUtilities {
     Map("pypi" -> Map("package" -> "mlflow"))
   ).toJson.compactPrint
 
+  val DLLibraries: String = List(
```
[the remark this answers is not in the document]
serena-ruan: Sure.

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/GPUTests.scala
```diff
@@ -0,0 +1,63 @@
...
+class GPUTests extends TestBase {
+
+  val clusterId: String = createClusterInPool(GPUClusterName, AdbGpuRuntime, PoolId)
+  val jobIdsToCancel: mutable.ListBuffer[Int] = mutable.ListBuffer[Int]()
+
+  println("Checking if cluster is active")
+  tryWithRetries(Seq.fill(60 * 15)(1000).toArray) { () =>
+    assert(isClusterActive(clusterId))
+  }
+  println("Installing libraries")
```
mhamilton723, line 23: might want to abstract this duplicate code in a function and use for both regular and GPU submission

python/setup.py
```diff
@@ -0,0 +1,80 @@
...
+    ],
+    url="https://github.com/Microsoft/SynapseML",
+    python_requires=f">={MINIMUM_SUPPORTED_PYTHON_VERSION}",
+    project_urls={
+        "Website": "https://microsoft.github.io/SynapseML/",
+        "Documentation": "https://mmlspark.blob.core.windows.net/docs/{}/pyspark/index.html".format(
+            version
+        ),
+        "Source Code": "https://github.com/Microsoft/SynapseML",
+    },
```
mhamilton723, line 79: We should have a deeper discussion about packaging next week because im not quite sure how this file was needed

python/src/synapse/ml/dl/DeepVisionClassifier.py
```diff
@@ -0,0 +1,181 @@
...
+        "string representation of optimizer function for the underlying pytorch_lightning model",
+    )
+
+    pretrained = Param(
+        Params._dummy(),
+        "pretrained",
+        "True/False, whether to use pretrained weights for the backbone model or not",
+    )
+
+    feature_extracting = Param(
```
mhamilton723, line 42: might want to name this something active like extract_features, also im not sure what this does

python/src/synapse/ml/dl/DeepVisionClassifier.py
```diff
@@ -0,0 +1,181 @@
...
+        "string representation of torch.nn.functional loss function for the underlying pytorch_lightning model, e.g. binary_cross_entropy",
+    )
+
+    optimizer_name = Param(
+        Params._dummy(),
+        "optimizer_name",
+        "string representation of optimizer function for the underlying pytorch_lightning model",
+    )
+
+    pretrained = Param(
```
mhamilton723, line 36: i think we'll always have to used pretrained weights if we arent training full net

mhamilton723 (changes requested):
Looks awesome!! Can we have a call to walk through this so i can be better at reviewing?

notebooks/deepLearning/DeepLearning - Deep Vision Classifier.ipynb
```diff
@@ -0,0 +1,350 @@
+{
```
mhamilton723, line 1: Some comments on this file (hard to leave at the right place so ill stick em in here)

1. You have some complex path and mounting stuff, consider using spark.readBinaryFiles, filtering out the content to just get the paths, and then modifying the name to point to mounted storage instead.

2. Some functions you define are camelcase, consider making these snake case because of python land

3. Is the _transform row function used?
4. in the definition of the udf you dont need to use a lambda, can justt pass func in directly
5. The comment "This is an imoportant parameter doesent tell enough to users, consider a brief explanation"
6. Is the dummy callback required? 
7. Might want to make the deep vision classifier have a few more defaults so that we can keep init simple and clean for this demo
8. you do this thing where you have to set the estimator and the model to have different names for "features" any way tto make it so that the code here is re-used and works without setting on the model?
9. It looks like we need a argmax udf before we send it into the evaluator. Would other spark multiclass stuff need this too? If not we should move our API to align with other spark multiclass stuff if possible
10. your top of the cell looks alot like horovod install script, any way to just use this script directly?

python/src/synapse/ml/dl/DeepVisionClassifier.py
```diff
@@ -0,0 +1,181 @@
...
+        Params._dummy(), "backbone", "backbone of the deep vision classifier"
+    )
+
+    num_layers_to_fine_tune = Param(
+        Params._dummy(),
+        "num_layers_to_fine_tune",
+        "number of last layers to fine tune for the model, should be smaller or equal to 3",
+    )
+
+    num_classes = Param(Params._dummy(), "num_classes", "number of target classes")
```
mhamilton723, line 22: Do these align with sparkML names? Not saying they dont right now, but just a good thing to double check

python/src/tests/dl/test_deep_vision_classifier.py
```diff
@@ -0,0 +1,143 @@
...
+            num_classes=17,
+            feature_cols=["image"],
+            label_cols=["label"],
+            optimizer_name="adam",
+            loss_name="cross_entropy",
+            batch_size=16,
+            epochs=epochs,
+            validation=0.1,
+            verbose=1,
+            profiler=None,
```
mhamilton723, line 117: You might want to apply the same similplifications from the notebook to this test, perhaps minus the removal of callbacks if you find those useful for debugging

python/version.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in project root for information.
+
+import re
+
```
mhamilton723, line 5: I think we'll want to just add this code as extra files in the deep learning python code so that the codegen system naturally fits in with this. If the setup.py that gets generated is wrong in some way for this project, perhaps we should use this as a first example in codegen of a custom or overriding setup.py for a project so that future developers can benefit from this exploration.

mhamilton723 (changes requested):

python/src/synapse/ml/dl/DeepVisionClassifier.py
serena-ruan: You're right, they use camelCase

python/src/synapse/ml/dl/DeepVisionClassifier.py
serena-ruan: So you mean we just remove this parameter and fix it to True internally and do not let users train the full net?

python/src/synapse/ml/dl/DeepVisionClassifier.py
serena-ruan: In LitDeepVisionModel: 
# Freeze those weights
if feature_extracting:
    for p in self.model.parameters():
        p.requires_grad = False

Should I also remove this param and always freeze all weights at the begining. And only free those layers users want to train explicitly?

python/src/synapse/ml/dl/DeepVisionClassifier.py
serena-ruan: But there're conflicts of this format as horovod uses snake_case... And we're extending their parameters, so should we just keep snake_case?

python/setup.py
serena-ruan: Yes, we should discuss about how we should release this. I create this setup.py file for this python folder as it doesn't have dependency on jars so we can release it directly, and it also has different environment requirements than our other pieces as it relies on horovod, pytorch, etc. So if this is an independent package, I didn't utilize our codegen system or the scala-side version generation, and we can just pack it and install with pip.

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/GPUTests.scala
serena-ruan: Done.

notebooks/deepLearning/DeepLearning - Deep Vision Classifier.ipynb
serena-ruan: 1.2.4.6.7.10. Done.
3. Yes, it's used in DeepVisionClassifier's transformation_fn parameter, so that it transforms the image internally.
5. I'll remove this parameter and I haven't figured out how this param should be set tbh.. If I left it as default value in databricks it triggers errors.
8. That's because only estimator accepts transformation_fn so we can use path as input, and model reads the features directly so I use another udf to transform the path into features before inference.
9. I think it's needed as our model output contains an array of label probability instead of the real label prediction.

mhamilton723 (changes requested):

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/DatabricksUtilities.scala
```diff
@@ -56,15 +61,27 @@ object DatabricksUtilities {
...
   val NotebookFiles: Array[File] = FileUtilities.recursiveListFiles(
     FileUtilities.join(
-      BuildInfo.baseDirectory.getParent, "notebooks").getCanonicalFile)
+      BuildInfo.baseDirectory.getParent, "notebooks", "features").getCanonicalFile)
 
   val ParallelizableNotebooks: Seq[File] = NotebookFiles.filterNot(_.isDirectory)
 
+  val GPUNotebooks: Seq[File] = FileUtilities.recursiveListFiles(
+    FileUtilities.join(
+      BuildInfo.baseDirectory.getParent, "notebooks", "deepLearning"
```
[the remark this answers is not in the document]
mhamilton723: This comment still holds

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/DatabricksUtilities.scala
```diff
@@ -334,6 +370,58 @@ object DatabricksUtilities {
   }
 }
 
+abstract class DatabricksTestProcess extends TestBase {
+  import DatabricksUtilities._
+
+  def databricksTestProcess(clusterId: String, libraries: String, notebooks: Seq[File]): mutable.ListBuffer[Int] = {
```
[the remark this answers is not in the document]
mhamilton723: still stands

python/src/synapse/ml/dl/DeepVisionClassifier.py
```diff
@@ -0,0 +1,320 @@
...
+        "string representation of optimizer function for the underlying pytorch_lightning model",
+    )
+
+    dropout_aux = Param(
+        Params._dummy(),
+        "dropout_aux",
+        "numeric value that's applied to googlenet InceptionAux module's dropout layer only: probability of an element to be zeroed",
+    )
+
+    transform_fn = Param(
```
mhamilton723, line 62: Is this expecting a pytorch transformation, or a func that transforms rows? The former would be a much nicer API for this

python/src/tests/dl/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+            transforms.RandomHorizontalFlip(),
+            transforms.ToTensor(),
+            transforms.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
+        ]
+    )
+    return transform
+
+
+@pytest.fixture(scope="module")
+def transform_row_func(transform):
```
mhamilton723, line 88: is this stuff still needed with your latest simplifications?

python/setup.py
```diff
@@ -0,0 +1,80 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
```
mhamilton723, line 1: Is this still needed if we go the route of asking people to install horovod and pytorch prior to installing SynapseMl so that this can be done in regular packaging workflow?

python/src/synapse/ml/dl/DeepVisionModel.py
```diff
@@ -0,0 +1,120 @@
...
+from pyspark import keyword_only
+import torchvision.transforms as transforms
+from PIL import Image
+import torch
+from pyspark.sql.functions import udf, col
+from pyspark.sql.types import DoubleType
+import numpy as np
+
+
+class DeepVisionModel(TorchModel):
```
mhamilton723, line 15: Perhaps we should also extend SparkML Probablistic Classifier here and in the estimator

python/src/tests/dl/test_deep_vision_classifier.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def test_mobilenet_v2(
+    transform_row_func,
+    get_data_path,
+    read_image_and_transform_udf,
+):
+    spark = SparkSession.builder.master("local[*]").getOrCreate()
+
+    # backend = SparkBackend(
```
mhamilton723, line 84: still need this?

python/src/synapse/ml/dl/DeepVisionClassifier.py
serena-ruan: Yes this is expecting a pytorch transformation, the func that transforms rows is TorchEstimator's transformation_fn parameter.

python/src/tests/dl/test_deep_vision_classifier.py
serena-ruan: Ah nope, will clean up the tests.

python/src/tests/dl/conftest.py
serena-ruan: Will clean up

python/src/synapse/ml/dl/DeepVisionModel.py
serena-ruan: Seems we can't inherit Probablistic Classification model here as it complains about `python cannot create consistent method ordering` , and this pobablistic Classifier also contains a complicated logic of different columns, so I'm going to extend the `_PredictorParams` directly and set the PredictionCol to the max value in model's outputCol.

serena-ruan: Ideally, we should persuade hovorod team to making HorovodModel extend PredictionModel.

deep-learning/src/main/python/synapse/ml/dl/DeepVisionModel.py
```diff
@@ -0,0 +1,141 @@
...
+        optimizer=None,
+        run_id=None,
+        _metadata=None,
+        loss=None,
+        loss_constructors=None,
+        # diff from horovod
+        transform_fn=None,
+        labelCol="label",
+        featuresCol="features",
+        predictionCol="prediction",
```
mhamilton723, line 39: I think we should either have all the params be settable with camel or none. Having half and half seems like it will confuse folks

notebooks/features/simple_deep_learning/DeepLearning - Deep Vision Classifier.ipynb
```diff
@@ -0,0 +1,182 @@
...
+    "    backbone=\"resnet50\",\n",
+    "    store=store,\n",
+    "    callbacks=callbacks,\n",
+    "    num_classes=17,\n",
+    "    batch_size=16,\n",
+    "    epochs=epochs,\n",
+    "    validation=0.1,\n",
+    ")\n",
+    "\n",
+    "deep_vision_model = deep_vision_classifier.fit(train_df).setOutputCols([\"label_prob\"])"
```
mhamilton723, line 139: is setOutputCols the right api here? I think for a predictor there are explicit cols for probabilities, rawProbabilities, and predictions

deep-learning/src/main/python/synapse/ml/dl/DeepVisionModel.py
serena-ruan: I'll create params instead of extending _PredictorParams.

notebooks/features/simple_deep_learning/DeepLearning - Deep Vision Classifier.ipynb
serena-ruan: Yes, TorchModel extends HasOutputCol, but we can actually delete this setOutputCols as it internally gives it default values, and I also added predictionCol for our model. We couldn't follow the exact same behavior as predictor because the definition and rawProbabilities don't really apply here, but I changed the _transform function to add predictionCol so the output dataframe is more clear.

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/CodeGen.scala
```diff
@@ -134,11 +135,27 @@ object CodeGen {
...
+         |    "pytorch_lightning>=1.5.0,<1.5.10",
+         |]
+         |
+         |dl_extra_list = ["torchvision>=0.12.0", "horovod==0.25.0"]""".stripMargin
+    } else ""
+    val extraRequirements = if (conf.name.contains("deep-learning")) {
+      s"""    install_requires=dl_require_list,
+         |    extras_require={"extras": dl_extra_list},
+         |    python_requires=f">={MINIMUM_SUPPORTED_PYTHON_VERSION}",""".stripMargin
+    } else ""
```
mhamilton723, line 158: nit: can this just be extraRequirements as opposed to also the dependency list? seems like its just additional stuff to override

mhamilton723: If its a matter of indentation theres a utility methoid for this

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/GPUTests.scala
```diff
@@ -0,0 +1,41 @@
...
+    file => {
+      val tmpFile = new File("tmp.txt")
+      val w = new PrintWriter(tmpFile)
+      Source.fromFile(file).getLines().map {
+        _.replaceAll("CUR_VERSION", BuildInfo.version)
+      }.foreach(x => w.println(x))
+      w.close()
+      tmpFile.renameTo(file)
+    }
+  )
```
mhamilton723, line 32: What is this code doing exactly?

deep-learning/src/main/python/synapse/ml/dl/DeepVisionModel.py
```diff
@@ -0,0 +1,124 @@
...
+        optimizer=None,
+        run_id=None,
+        _metadata=None,
+        loss=None,
+        loss_constructors=None,
+        # diff from horovod
+        transform_fn=None,
+        label_col="label",
+        image_col="image",
+        prediction_col="prediction",
```
mhamilton723, line 39: If we dont use the actual predictor API in SparkML will we work OOTB with the Evaluators?

deep-learning/src/test/python/synapsemltest/dl/test_deep_vision_classifier.py
```diff
@@ -0,0 +1,113 @@
...
+    ]
+
+    train_folder, test_folder = get_data_path
+
+    with local_store() as store:
+
+        deep_vision_classifier = DeepVisionClassifier(
+            backbone="mobilenet_v2",
+            store=store,
+            backend=ctx,
```
mhamilton723, line 94: Is this test skipped for a reason?

deep-learning/src/test/python/synapsemltest/dl/test_deep_vision_model.py
```diff
@@ -0,0 +1,117 @@
...
+        additional_layers_to_train=1,
+        num_classes=17,
+        input_shape=[-1, 3, 224, 224],
+        optimizer_name="adam",
+        loss_name="cross_entropy",
+        label_col="label",
+        image_col="image",
+    )
+
+    callbacks = [MyDummyCallback(epochs), ModelCheckpoint(dirpath="target/resnet50/")]
```
mhamilton723, line 114: nit: can probabbly drop checkpoint callback here to avoid polluting

notebooks/features/simple_deep_learning/DeepLearning - Deep Vision Classifier.ipynb
```diff
@@ -0,0 +1,186 @@
...
+    "### Environment Setup -- reinstall horovod based on new version of pytorch"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install https://mmlspark.blob.core.windows.net/pip/CUR_VERSION/synapseml_deep_learning-0.10.0.dev1-py2.py3-none-any.whl --force-reinstall --no-deps"
```
mhamilton723, line 23: Is this necessary with the latest stuff to make it packaged along with python jar?

notebooks/features/simple_deep_learning/DeepLearning - Deep Vision Classifier.ipynb
```diff
@@ -0,0 +1,186 @@
...
+    "from horovod.spark.common.store import DBFSLocalStore\n",
+    "from pytorch_lightning.callbacks import ModelCheckpoint\n",
+    "from synapse.ml.dl import *\n",
+    "\n",
+    "run_output_dir = \"/dbfs/FileStore/test/resnet50\"\n",
+    "store = DBFSLocalStore(run_output_dir)\n",
+    "\n",
+    "epochs = 10\n",
+    "\n",
+    "callbacks = [ModelCheckpoint(filename=\"{epoch}-{train_loss:.2f}\")]"
```
mhamilton723, line 124: nit: You can probabbly drop ModelCheckpoint if you want

mhamilton723 (changes requested):
A few questions (sorryyyyyy!)

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/GPUTests.scala
serena-ruan: In the notebook, the first cell is this `%pip install https://mmlspark.blob.core.windows.net/pip/CUR_VERSION/synapseml_deep_learning-0.10.0.dev1-py2.py3-none-any.whl --force-reinstall --no-deps` so I add this extra piece to replace CUR_VERSION to the correct build version, then it will install the wheel package on all nodes in databricks. FYI I have tried excluding this pip install and just install our jar package, unfortunately that doesn't work with horovod as it seems can't find the correct module where dl code lives. I'm not sure why though, but installing the wheel package should be enough for deep vision usage.

deep-learning/src/test/python/synapsemltest/dl/test_deep_vision_classifier.py
serena-ruan: emmm yes, perhaps I could also delete this test. I wrote this test for local tests, but I don't have gpus locally so can't enabling the gpu test, same thing for pipeline I guess. Another reason is that this test looks exactly the same as the notebook. So I skip this test in case others want to try with local machines containing gpus.

notebooks/features/simple_deep_learning/DeepLearning - Deep Vision Classifier.ipynb
serena-ruan: Yes, see my comment on GPUtests :D

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/CodeGen.scala
serena-ruan: Done.

deep-learning/src/test/python/synapsemltest/dl/test_deep_vision_classifier.py
mhamilton723: ok so this is kept due to the simplicity of workign with it, makes sense

core/src/test/scala/com/microsoft/azure/synapse/ml/nbtest/GPUTests.scala
mhamilton723: you might be able to use synapse.ml.core.__scala_version__ here instead of code generation/modification that is hard for the user to do

mhamilton723 (approved):

PR #1520, opened 2022-06-06
feat: add black hook

serena-ruan:
# Summary

Add black as a hook to auto-format python files. This is used to auto-format python codes before submitting a PR.
To enable it please set `git config core.hooksPath hooks` in your local dev environment.

# Tests

Tested locally. You could only commit a message after passing the black check

# Dependency changes

Add black as requirement for local development.


mhamilton723 (changes requested):
Just realized we might want to enforce this with the --check flag in our build. 

https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#exit-code

serena-ruan:
> Just realized we might want to enforce this with the --check flag in our build.
> 
> https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#exit-code

In the pre-commit file, it runs `black --check` against the git diff result.

PR #1526, opened 2022-06-12
fix: Fix multiclass training with initial scores

svotaw:
# Summary

The use of initial scores with multiclass training was not working.  There was 1 test, but it was marked as "ignore" due to being flaky.  In actuality, it only passed sometimes out of pure luck since there were several bugs.

1. The test did not use the "multiclass" objective, so LightGBM was actually using binary under the covers
2. We were only sending 1 column of initial score data for multiclass, so when marked as true multiclass, the test always failed due to LightGBM reporting the error.
3. LightGBM represents multiclass initial scores in column format, and we were sending in row format.  The data needed to be transposed to be correct.

# Tests

I turned the test back on, although it is still flaky for some reason.  Could use some advice on this.  However, having traced through all the code underneath, we are now appropriately sending multiclass initial scores to LightGBM.  One column is sent for every class, and it is transposed from row to column format.

# Dependency changes

_If you needed to make any changes to dependencies of this project, please describe them here._

AB#1827802


lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/DatasetAggregator.scala
```diff
@@ -40,6 +40,46 @@ private[lightgbm] object ChunkedArrayUtils {
         chunkedArray.getItem(chunkCount, lastChunkIdx, defaultVal))
     }
   }
+
+  def insertTransposedChunkedArray[T: Numeric](chunkedArray: ChunkedArray[T],
```
svotaw, line 44: Note that I tried to put the transpose in several places, and there were some that were more efficient and easier to read, but they would require changes in the LightGBM library to add functionality.  I figured it was better to just fix it with SynapseML changes only since we are moving to streaming, where this is not a problem.

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMClassifier.scala
```diff
@@ -185,13 +185,16 @@ class VerifyLightGBMClassifier extends Benchmarks with EstimatorFuzzing[LightGBM
...
   }
 
-  ignore("Verify LightGBM Multiclass Classifier with vector initial score") {
-    val scoredDF1 = baseModel.fit(breastTissueDF).transform(breastTissueDF)
-    val df2 = scoredDF1.withColumn(initScoreCol, col(rawPredCol))
-      .drop(predCol, rawPredCol, probCol, leafPredCol, featuresShapCol)
-    val scoredDF2 = baseModel.setInitScoreCol(initScoreCol).fit(df2).transform(df2)
-
-    assertMulticlassImprovement(scoredDF1, scoredDF2)
+  test("Verify LightGBM Multiclass Classifier with vector initial score") {
```
imatiach-msft, line 188: might be a good idea to add a test just for the insertTransposedChunkedArray to validate it, but not necessary for this PR

imatiach-msft (dismissed):

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMCommon.scala
```diff
@@ -0,0 +1,77 @@
...
+        // Assert column order in source (should be sequential numbers)
+        val array = SwigUtils.nativeDoubleArrayToArray(transposedArray.array, rows * cols)
+        assert(array.zipWithIndex.forall(pair => pair._1 == pair._2))
+      } finally {
+        transposedArray.delete()
+      }
+    })
+  }
+
+  test("Verify chunked array transpose complex") {
```
imatiach-msft, line 46: nice tests!

imatiach-msft (approved):

PR #1546, opened 2022-06-24
fix: remove debugging code

niehaus59:
# Summary

Remove some code I used earlier for debugging that got checked in inadvertently

# Tests

N/A

# Dependency changes

No dependency changes

AB#1845700


mhamilton723 (approved):
Thanks for this cleanup Mark :)

PR #1557, opened 2022-07-08
fix: Refactor of LightGBM wrapper to prepare for streaming

svotaw:
# Summary

In preparation for adding a "streaming" mode to LightGBM functionality (mostly to improve memory usage), this PR refactors the existing wrapper code.  Followup PRs will actually add the streaming components, but this initial PR makes the following major interrelated changes to improve the organization of LightGBM code:
1. Splits the driver and worker methods into distinct classes.  `LightGBMBase` is now driver-only methods, and all worker code is in `BasePartitionTask` and derived classes (only `BulkPartitionTask `for now). 
2. Splits Networking-specific methods into a new `NetworkManager`, a class that encapsulates all Networking functionality, both driver/worker and Spark/LightGBM.
3. Creates `TrainingContext` object to help pass around configuration and initialization information easily, without passing lots of method parameters.  This applies at the driver level for all of training, but there are also similar new `PartitionTaskContext` and `PartitionTaskTrainingState` objects that encapsulate read-only and variable properties respectively for individual tasks.
4. Adds `InstrumentationMeasures`, a class that holds timing perf for various stages, both for all of training and individual tasks.  This class is returned along with the Booster, so is even exposed to the user.

I also cleaned up all warnings about file/method length, use of "while" loops, excessive code branching, etc.  There's more official tail recursion and less use of vars.

Notable new functionality:
1. The networking phase now gathers information about the partition size and distribution across executors, and passes that to all Tasks (previously it was just limited info, and just "main" workers).
2. Timing perf can now be reported to the user and can also help us with benchmarking and diagnostics

**NOTE**: The actual streaming components are not huge, but were omitted for now to help shorten the long PR, and also since this PR can be checked in without any new native LightGBM library components.

# Tests

Since this is a refactor, the existing LightGBM test suite should be sufficient, and for now runs in "bulk" mode only.  The next PR will add "streaming" mode and reconcile the tests. 

# Dependency changes

_If you needed to make any changes to dependencies of this project, please describe them here._

AB#1866292


lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BulkPartitionTask.scala
```diff
@@ -0,0 +1,130 @@
...
+                                             referenceDataset: Option[LightGBMDataset]): LightGBMDataset = {
+    val ac = if (forValidation) dataState.aggregatedValidationData.get
+             else dataState.aggregatedTrainingData.get
+    try {
+      val datasetInner: LightGBMDataset = ac.generateDataset(ctx, referenceDataset)
+      ctx.trainingCtx.columnParams.groupColumn.foreach(_ => datasetInner.addGroupColumn(ac.getGroups))
+      datasetInner.setFeatureNames(ctx.trainingCtx.featureNames, ac.getNumCols)
+      datasetInner
+    } finally {
+      ac.cleanup()
```
mhamilton723, line 65: likewise here

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
```diff
@@ -374,251 +376,288 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
...
+                                       trainingParams: BaseTrainParams,
+                                       numCols: Int,
+                                       measures: InstrumentationMeasures): (Array[Row], Array[Long]) = {
+    measures.markRowStatisticsStart()
+
+    // Get the row counts per partition
+    measures.markRowCountsStart()
+    val indexedRowCounts: Array[(Int, Long)] = dataframe
+      .select(getLabelCol)
+      .rdd
```
mhamilton723, line 509: nit: you might be able to avoid rdd api by using some api on the task context or something like this to get the partition number

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
```diff
@@ -374,251 +376,288 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
...
+      .mapPartitionsWithIndex({case (i,rows) => Iterator((i,rows.size.toLong))}, true)
+      .collect()
+    val rowCounts: Array[Long] = new Array[Long](indexedRowCounts.length)
+    indexedRowCounts.foreach(pair => rowCounts(pair._1) = pair._2)
```
mhamilton723, line 513: Can you just make this array explicityly instead of doing this mutation

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
```diff
@@ -374,251 +376,288 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
...
-    val mapPartitionsFunc = trainLightGBM(batchIndex, networkParams,
-      validationData, trainParams, numTasksPerExec, schema, sharedState.get)(_)
+  protected def executePartitionTasks(ctx: TrainingContext,
+                                      dataframe: DataFrame,
+                                      measures: InstrumentationMeasures): LightGBMBooster = {
+    // Create the object that will manage the mapPartitions function
+    // TODO next PR, add in StreamingPartitionTask
+    val workerTaskHandler: BasePartitionTask = if (ctx.isStreaming) new BulkPartitionTask()
+    else new BulkPartitionTask()
+    val mapPartitionsFunc = workerTaskHandler.mapPartitionTask(ctx)(_)
```
mhamilton723, line 597: Looks like an if that returns same thing in both branches

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
```diff
@@ -636,18 +675,22 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
...
+
+  protected def calculateCustomTrainParams(params: BaseTrainParams, dataset: Dataset[_]): Unit = { }
 
   protected def stringFromTrainedModel(model: TrainedModel): String
 
   /** Allow algorithm specific preprocessing of dataset.
     *
-    * @param dataset The dataset to preprocess prior to training.
+    * @param df The data frame to preprocess prior to training.
     * @return The preprocessed data.
```
mhamilton723, line 693: nit: dataframe

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/ClassifierTrainParams.scala
```diff
@@ -21,9 +20,13 @@ case class ClassifierTrainParams(passThroughArgs: Option[String],
                                  objectiveParams: ObjectiveParams,
                                  seedParams: SeedParams,
                                  categoricalParams: CategoricalParams) extends BaseTrainParams {
+  var numClass: Int = 1
```
mhamilton723, line 23: Case classes shouldnt have mutable state, you can get around this by adding a setNumClass method that returns another case class

mhamilton723 (changes requested):
Left a few nits and comments, awesome work :)

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
svotaw: not sure what you mean. this is getting a count of the number of rows in each partition by partition id.  there is no information about partition row count in task context.  And if there is, PLEASE show me. :)

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BasePartitionTask.scala
```diff
@@ -0,0 +1,415 @@
...
+
+          // Create the final Dataset for training and execute training iterations
+          finalizeDatasetAndTrain(taskCtx, dataIntermediateState)
+        } else {
+          log.info(s"Helper task ${taskCtx.taskId}, partition ${taskCtx.partitionId} finished processing rows")
+          ctx.sharedState().dataPreparationDoneSignal.countDown()
+          Array { PartitionResult(None, taskCtx.measures) }.toIterator
+        }
+      } finally {
+        cleanup(taskCtx)
```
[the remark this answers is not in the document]
svotaw: I did not change this logic, just moved it and extracted the cleanup code to it's own method.  But I believe it is correct.  If you look at what cleanup() does, it frees the network and any other internal needs for bulk or streaming.  It does not operate on the iterator (which is not even passed to the cleanup method).  If this throws, the whole Task fails.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BulkPartitionTask.scala
svotaw: I just moved this logic and did not modify it.  But I believe it does the correct thing.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/ClassifierTrainParams.scala
svotaw: can you give an example of what you mean by setNumClass() method that returns an object?  (setters in my experience do not return values).  Do you mean recreate the whole object after changing 1 property?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMPerformance.scala
```diff
@@ -0,0 +1,223 @@
...
+      - initializationTime
+      - dataPreparationTime
+      - waitTime
+      - datasetCreationTime
+      - validationDatasetCreationTime
+      - trainingIterationsTime)
+  }
+}
+
+class InstrumentationMeasures() extends Serializable {
```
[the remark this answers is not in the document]
svotaw: yeah, this class was created ad hoc over time and I didn't go back and clean it up.  will do

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
eisber: I'll also need that information for VW bindings. can we move it into a util... and given anybody finds a better way of doing it, update it there?

svotaw: > I'll also need that information for VW bindings. can we move it into a util... and given anybody finds a better way of doing it, update it there?

You need what information exactly?  Just so I'm clear on what I'm turning into a util. :)

eisber: I need to know the number of rows per partition to trigger all reduce.

mhamilton723 (dismissed):
Approved to the best of my knowledge, @imatiach-msft @eisber please feel free to take a look and provide any comments!

mhamilton723 (approved):

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
svotaw: done, added a util in core ClusterUtil

eisber (commented):
Reviewed VW changes. Nice work!

PR #1580, opened 2022-07-21
feat: Add LightGBM streaming execution mode

svotaw:
# Summary

Add the streaming execution mode to LightGBM wrapper.  This mode uses almost no memory on top of what LightGBM needs to execute.

# Tests

Tests will be modified to run in both bulk and streaming mode before checkin.  Before this PR was pushed, LightGBMClassifier tests were all passing for streaming mode (the bulk of the tests).
There are also some new tests just for streaming components and instrumentation in Common.

# Dependency changes

This PR cannot be checked in without corresponding changes in native LightGBM library, or pointing to custom upload.  https://github.com/microsoft/LightGBM/pull/5299

### DIFF
--- a/lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BasePartitionTask.scala
+++ b/lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BasePartitionTask.scala
@@ -33,7 +33,7 @@ case class PartitionDataState(aggregatedTrainingData: Option[BaseAggregatedColum
   */
 case class PartitionTaskTrainingState(ctx: PartitionTaskContext,
                                       booster: LightGBMBooster) {
-  lazy val evalNames: Array[String] = booster.getEvalNames()
+  lazy val evalNames: Array[String] = booster.getEvalNames
   lazy val evalCounts: Int = evalNames.length
   lazy val bestScore = new Array[Double](evalCounts)
   lazy val bestScores = new Array[Array[Double]](evalCounts)
@@ -79,13 +79,25 @@ case class PartitionTaskContext(trainingCtx: TrainingContext,
                               value)
   }
 
-  lazy val streamingExecutorRowCount: Int = {
+  /* The relative index of this Task's thread relative to other executor threads
+   */
+  lazy val threadIndex: Int = networkTopologyInfo.executorPartitionIdList.sortBy(id => id).indexOf(partitionId)
+
+  /* The count of partitions in this executor
+   */
+  lazy val executorPartitionCount: Int = trainingCtx.partitionCounts.get.length
+
+  /* The total count of partition rows in this executor
+   */
+  lazy val executorRowCount: Int = {
     // Use the map of all partition counts and the map of partitions on this executor to get local count
     val allPartitionRowCounts = trainingCtx.partitionCounts.get
     networkTopologyInfo.executorPartitionIdList
       .map(partitionId => allPartitionRowCounts(partitionId)).sum.toInt
   }
 
+  /* The starting offset of where this partition should place rows in the shared executor Dataset
+   */
   lazy val streamingPartitionOffset: Int = {
     // Use the map of all partition counts and the map of partitions on this executor to get this partition offset
     val allPartitionRowCounts = trainingCtx.partitionCounts.get
@@ -225,11 +237,11 @@ abstract class BasePartitionTask extends Serializable with Logging {
   private def finalizeDatasetAndTrain(ctx: PartitionTaskContext,
                                       dataState: PartitionDataState): Iterator[PartitionResult] = {
     ctx.measures.isActiveTrainingTask = true
-    val trainDataset: LightGBMDataset = generateFinalDataset(ctx, dataState, forValidation = false, None)
+    val trainDataset: LightGBMDataset = getFinalTrainingDataset(ctx, dataState)
     try {
       val validDatasetOpt: Option[LightGBMDataset] =
         if (!ctx.trainingCtx.hasValidationData) None
-        else Option(generateFinalDataset(ctx, dataState, forValidation = true, Some(trainDataset)))
+        else Option(getFinalValidationDataset(ctx, dataState, trainDataset))
 
       try {
         log.info(s"Creating LightGBM Booster for partition ${ctx.partitionId}, task ${ctx.taskId}")
@@ -304,6 +316,9 @@ abstract class BasePartitionTask extends Serializable with Logging {
     if (trainingCtx.useSingleDatasetMode)
       log.info(s"Using singleDatasetMode. Task: $taskId, PartId: $partitionId. Main task: $mainExecutorWorkerId" +
         s" shouldExecuteTraining: $shouldExecuteTraining, isEmptyPartition: $isEmptyPartition")
+
+    val shouldCalcValidationDataset = trainingCtx.sharedState.validationDatasetWorker.getOrElse(-1) == taskId
+
     PartitionTaskContext(trainingCtx,
                          partitionId,
                          taskId,
@@ -312,7 +327,7 @@ abstract class BasePartitionTask extends Serializable with Logging {
                          shouldExecuteTraining,
                          isEmptyPartition,
                          shouldReturnBooster,
-                         shouldCalcValidationDataset = false)
+                         shouldCalcValidationDataset)
   }
 
   /**
@@ -324,17 +339,24 @@ abstract class BasePartitionTask extends Serializable with Logging {
   protected def preparePartitionDataInternal(ctx: PartitionTaskContext, inputRows: Iterator[Row]): PartitionDataState
 
   /**
-    * Generate the final dataset for this task.  Internal implementation for specific execution modes.
+    * Generate the final training dataset for this task.  Internal implementation for specific execution modes.
     * @param ctx The training context.
     * @param dataState Any intermediate data state (used mainly by bulk execution mode).
-    * @param forValidation Whether to generate the final training dataset or the validation dataset.
-    * @param referenceDataset A reference dataset to start with (used mainly for validation dataset).
     * @return LightGBM dataset Java wrapper.
     */
-  protected def generateFinalDatasetInternal(ctx: PartitionTaskContext,
+  protected def getTrainingDatasetInternal(ctx: PartitionTaskContext,
+                                           dataState: PartitionDataState): LightGBMDataset
+
+  /**
+    * Generate the final opt validation dataset for this task. Internal implementation for specific execution modes.
+    * @param ctx The training context.
+    * @param dataState Any intermediate data state (used mainly by bulk execution mode).
+    * @param referenceDataset A reference dataset to start with.
+    * @return LightGBM dataset Java wrapper.
+    */
+  protected def getValidationDatasetInternal(ctx: PartitionTaskContext,
                                              dataState: PartitionDataState,
-                                             forValidation: Boolean,
-                                             referenceDataset: Option[LightGBMDataset]): LightGBMDataset
+                                             referenceDataset: LightGBMDataset): LightGBMDataset
 
   /** Cleanup the task
     *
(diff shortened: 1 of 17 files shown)


lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BasePartitionTask.scala
```diff
@@ -348,35 +370,46 @@ abstract class BasePartitionTask extends Serializable with Logging {
...
+  /**
+    * Generate the final validation dataset for this task.  This should only be run be tasks that will participate in
+    * the training rounds, i.e. in useSingleDataset mode it will only be 1 task/executor.
+    * @param ctx The training context.
+    * @param dataState Any intermediate data state (used mainly by bulk execution mode).
+    * @param referenceDataset A reference dataset to start with (used mainly for validation dataset).
+    * @return LightGBM dataset Java wrapper.
+    */
+  private def getFinalValidationDataset(ctx: PartitionTaskContext,
+                                             dataState: PartitionDataState,
```
mhamilton723, line 401: nit: some weird indentation here

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BulkPartitionTask.scala
```diff
@@ -50,12 +50,20 @@ class BulkPartitionTask extends BasePartitionTask {
...
-                                             referenceDataset: Option[LightGBMDataset]): LightGBMDataset = {
-    val ac = if (forValidation) dataState.aggregatedValidationData.get
-             else dataState.aggregatedTrainingData.get
+                                             referenceDataset: LightGBMDataset): LightGBMDataset = {
+    generateDataset(ctx, dataState.aggregatedValidationData.get, Option(referenceDataset))
+  }
+
+  protected def generateDataset(ctx: PartitionTaskContext,
+                                ac: BaseAggregatedColumns,
+                                referenceDataset: Option[LightGBMDataset]): LightGBMDataset = {
```
mhamilton723, line 66: These look like rather thin methods, are they used in multiple ways or overriden elsewhere?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
```diff
@@ -626,10 +627,6 @@ trait LightGBMBase[TrainedModel <: Model[TrainedModel]] extends Estimator[Traine
                                    batchIndex: Int,
                                    numTasksPerExecutor: Int,
                                    networkManager: NetworkManager): TrainingContext = {
-    if (trainParams.executionParams.executionMode != LightGBMConstants.BulkExecutionMode) {
-      throw new Exception("Only bulk execution mode supported for now")
```
mhamilton723, line 630: Maybe make this an Illegal argument exception?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/TrainingContext.scala
```diff
@@ -45,19 +47,27 @@ case class TrainingContext(batchIndex: Int,
...
   val hasInitialScores: Boolean = { columnParams.initScoreColumn.isDefined &&
                                     columnParams.initScoreColumn.get.nonEmpty }
   val hasGroups: Boolean = { columnParams.groupColumn.isDefined && columnParams.groupColumn.get.nonEmpty }
 
+  val hasWeightsAsInt: Int = toInt(hasWeights)
+  val hasInitialScoresAsInt: Int = toInt(hasInitialScores)
+  val hasGroupsAsInt: Int = toInt(hasGroups)
+
+  if (!isStreaming && !isBulk) {
+    throw new Exception("Execution mode must be streaming or bulk")
```
mhamilton723, line 68: might want to throw IllegalArgumentException here

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/BaseTrainParams.scala
```diff
@@ -20,6 +20,8 @@ abstract class BaseTrainParams extends Serializable {
   def seedParams: SeedParams
   def categoricalParams: CategoricalParams
 
+  def getNumClass: Int = 1
```
mhamilton723, line 23: Is this hard-coding intentional?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/LightGBMParams.scala
```diff
@@ -66,7 +66,7 @@ trait LightGBMExecutionParams extends Wrappable {
 
   val microBatchSize = new IntParam(this, "microBatchSize",
     "Specify how many elements are sent in a streaming micro-batch.")
-  setDefault(microBatchSize -> 1)
+  setDefault(microBatchSize -> 100)
```
mhamilton723, line 69: we should make sure that if a partition contains 70 rows or something than it doesent hang

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/swig/SwigUtils.scala
```diff
@@ -169,3 +169,79 @@ class IntSwigArray(val array: SWIGTYPE_p_int) extends BaseSwigArray[Int] {
...
+  def getItem(index: Long): SWIGTYPE_p_void = lightgbmlib.voidPtrArray_getitem(array, index)
+
+  def setItem(index: Long, item: SWIGTYPE_p_void): Unit = {
+    lightgbmlib.voidPtrArray_setitem(array, index, item) // set native pointer value
+  }
+
+  def delete(): Unit =  {
+    lightgbmlib.delete_voidPtrArray(array)
+  }
+}
```
mhamilton723, line 197: Is there any way to push some of this stuff back into the bindings themselves? This isnt something thats absolutely necessary right now but might be good to think about once we get this working

lightgbm/src/test/resources/log4j.properties
```diff
@@ -5,4 +5,6 @@ log4j.appender.stdout.layout.ConversionPattern=%d{HH:mm:ss} %-5p %c{1}:%L - %m%n
 
 log4j.rootLogger=WARN, stdout
 log4j.logger.org.apache.spark=WARN, stdout
+log4j.additivity.org.apache.spark=false
 log4j.logger.com.microsoft=INFO, stdout
+log4j.additivity.com.microsoft=false
```
mhamilton723, line 10: What do these lines and file do?

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMClassifier.scala
```diff
@@ -366,43 +364,43 @@ class VerifyLightGBMClassifier extends Benchmarks with EstimatorFuzzing[LightGBM
...
+
+             // model2 should terminate early
+            val model2 = baseModel
+              .setNumLeaves(100)
+              .setNumIterations(100)
+              .setLearningRate(0.9)
+              .setMinDataInLeaf(2)
+              .setMetric("auc")
+              .setValidationIndicatorCol(validationCol)
+              .setEarlyStoppingRound(5)
```
mhamilton723, line 386: Some coded repetition here, might be able to copy and set just the param thats diff

mhamilton723: also looks like weird indent

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMCommon.scala
```diff
@@ -74,4 +91,213 @@ class VerifyLightGBMCommon extends TestBase {
...
+      datasetVoidPtr = lightgbmlib.voidpp_handle()
+      val resultCode = lightgbmlib.LGBM_DatasetCreateFromSampledColumn(
+        sampledData.getSampleData,
+        sampledData.getSampleIndices,
+        numCols,
+        sampledData.getRowCounts,
+        numRows,
+        numRows,
+        s"max_bin=255 bin_construct_sample_cnt=$numRows min_data_in_leaf=1 num_threads=3",
+        datasetVoidPtr)
```
mhamilton723, line 133: This looks a little low level, is this what is expected of a user or more just for testing internal bits

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMCommon.scala
```diff
@@ -74,4 +91,213 @@ class VerifyLightGBMCommon extends TestBase {
...
+    medianMax = getMedian(measurements.map(m => m.taskOverheadTimes().max))
+    println(s"Median-max Task overhead times, $medianMax")
+    medianMax = getMedian(measurements.map(m => m.taskInitializationTimes().max))
+    println(s"Median-max Task initialization times, $medianMax")
+    medianMax = getMedian(measurements.map(m => m.taskDataPreparationTimes.max))
+    println(s"Median-max Task data preparation times, $medianMax")
+    medianMax = getMedian(measurements.map(m => m.taskDatasetCreationTimes.max))
+    println(s"Median-max Task dataset creation times, $medianMax")
+    medianMax = getMedian(measurements.map(m => m.taskTrainingIterationTimes.max))
+    println(s"Median-max Task training iteration times, $medianMax")
```
mhamilton723, line 292: This looks like alot of duplicated code, and in previous PRs there is a ton of methods like this. Perhaps we should make these measurements just entries in a Map so that its more "dynamic" and we can iterate through it better

mhamilton723 (changes requested):
Mostly just minor things, thanks so much for your work on this and super excited to help get this in

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMCommon.scala
svotaw: not sure what you mean.  this is a unit test of the class.  what does a user have to do with it?

svotaw: IMHO, there aren't enough "low level" tests that are really testing the code :)  We should really have more true unit tests, and not just functional API tests

lightgbm/src/test/resources/log4j.properties
svotaw: Without these lines, every log entry gets printed twice in the output.  Without the file, there are no logs for tests at all.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/LightGBMParams.scala
svotaw: I know it doesn't (since there are no test datasets with an even 100 number of rows), but you are right there should be a test for it explicitly

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/BaseTrainParams.scala
svotaw: yes, it gets overridden by a multiclass parameters

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
mhamilton723: Still applies i think

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMCommon.scala
```diff
@@ -74,4 +92,206 @@ class VerifyLightGBMCommon extends TestBase {
...
+    printMedianMeasure("Median Row statistics time", measures, m => m.rowStatisticsTime())
+    printMedianMeasure("Median Training time", measures, m => m.trainingTime())
+    printMedianMeasure("Median Overhead time", measures, m => m.overheadTime)
+    printMedianMeasure("Median-max Task total times", measures, m => m.taskTotalTimes().max)
+    printMedianMeasure("Median-max Task overhead times", measures, m => m.taskOverheadTimes().max)
+    printMedianMeasure("Median-max Task initialization times", measures, m => m.taskInitializationTimes().max)
+    printMedianMeasure("Median-max Task data preparation times", measures, m => m.taskDataPreparationTimes().max)
+    printMedianMeasure("Median-max Task dataset creation times", measures, m => m.taskDatasetCreationTimes().max)
+    printMedianMeasure("Median-max Task training iteration times", measures, m => m.taskTrainingIterationTimes().max)
+  }
```
mhamilton723, line 280: this perf metrics code looks a little repetitive, perhaps it could be refactored to do this without a types API so you can just loop over metrics and print out these different variants

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMCommon.scala
```diff
@@ -74,4 +92,206 @@ class VerifyLightGBMCommon extends TestBase {
...
+
+    (0 until measurementCount).foreach(i => {
+      val measuredModel = baseModel
+        .setUseSingleDatasetMode(useSingleDataset)
+        .setExecutionMode(executionMode)
+        .setMatrixType(matrixType)
+        .setMicroBatchSize(microBatchSize)
+      println(s"** Start Measurement $i")
+      val _ = measuredModel.fit(train)
+      measures(i) = measuredModel.getPerformanceMeasures.get
```
mhamilton723, line 246: measures should probabblky be accessible on the fit model not the estimator

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split2/VerifyLightGBMRankerBulk.scala
```diff
@@ -0,0 +1,13 @@
...
+package com.microsoft.azure.synapse.ml.lightgbm.split2
+
+import com.microsoft.azure.synapse.ml.lightgbm.LightGBMConstants
+
+/** Tests to validate the functionality of LightGBM Ranker module in bulk mode. */
+class VerifyLightGBMRankerBulk extends VerifyLightGBMRankerStream {
+  override lazy val executionMode: String = LightGBMConstants.BulkExecutionMode
+  override def ignoreSerializationFuzzing: Boolean = true
+  override def ignoreExperimentFuzzing: Boolean = true
+}
```
mhamilton723, line 13: This looks a little odd, if you are sharing an implementation and just changing this execution mode, perhaps make the base class an actual base class instead of having bulk inherit stream. It will add 1 extra class, but it should be more in-line with the actual abstraction

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMCommon.scala
svotaw: This is really just a temporary util that I made an ignored test for now, which is why I didn't try and optimize it. It was some quick and dirty code. :) It was for some basic benchmarking, and I will refactor it later as part of doing more official perf analysis/benchmarking on the streaming stuff. I'll either make it a real util somewhere or refactor it as you said to be less repetitive.

svotaw: in some future PR  :)

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/VerifyLightGBMCommon.scala
svotaw: ok, I can think about this as part of the future benchmarking work. The performance measures are already in main, so this PR doesn't really touch that.

I think the decision to put it on the estimator had to do with some interaction between batching, error handling, and the fact that the model doesn't exist until the end. But looking at the code, it looks like we could redesign it that way, and I think you are right.

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split2/VerifyLightGBMRankerBulk.scala
svotaw: fair enough, I think I did it this way in anticipation that one day we'd just remove bulk. :)

Oh I know why I did it this way... if you make an abstract base class, you can't click on individual tests and run them in IJ. It was for convenience of being able to run individual tests.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/LightGBMBase.scala
mhamilton723: im dumb lol

mhamilton723 (approved):
Amazing work!

PR #1586, opened 2022-07-26
feat: R test gen

niehaus59:
This PR introduces test generation for the R interface to SynapseML, similar to that for Python and Dotnet.

We generate tests for models in the R language that call the generated R functions that in turn call Scala code. 

## How is this patch tested?
We generate tests and run them.

- [x] Yes. Make sure the dependencies are resolved correctly, and list changes here.
We have new dependences on R packages jsonlite and mlflow. Dependences on sparklyr and testthat were already present.

AB#1898553


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
```diff
@@ -299,7 +299,7 @@ trait HasSetLocation extends Wrappable with HasURL with HasUrlPath {
   }
 
   def setLocation(v: String): this.type = {
-    setUrl(s"https://$v.api.cognitive.microsoft.com/" + urlPath)
+    setUrl(s"https://$v.api.cognitive.microsoft.com/" + urlPath.stripPrefix("/"))
```
niehaus59, line 302: >stripPrefix("/") [](http://example.com/codeflow?start=64&length=16)

driveby

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/TextAnalytics.scala
```diff
@@ -609,4 +633,10 @@ class TextAnalyze(override val uid: String) extends TextAnalyticsBase(uid)
     )
   }
 
+  // total kluge for R
+  private def convertJsonToTasks(jsonString: String): Seq[TextAnalyzeTask] = {
+    implicit val formats = DefaultFormats
+    val p = parse(jsonString).extract[Seq[Map[String, Map[String,String]]]]
+    Seq[Map[String, String]](p.head.head._2).map(m => TextAnalyzeTask(m))
+  }
```
niehaus59, line 641: This one was particularly awful. I couldn't find a lower level way to pass the argument.

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/cognitive/split3/SpeechToTextSDKSuite.scala
```diff
@@ -29,7 +29,7 @@ trait SpeechToTextSDKSuiteBase extends TestBase with CognitiveKey with CustomSpe
   import spark.implicits._
 
   val region = "eastus"
-  lazy val resourcesDir = new File(new File(getClass.getResource("/").toURI).toString)
```
niehaus59, line 32: >new File(new File(getClass.getResource("/").toURI).toString) [](http://example.com/codeflow?start=26&length=60)

This nested File structure was causing a test failure. Not sure why it didn't cause problems in python or dotnet.

core/src/main/R/synapse/ml/core/serialize/java_params_patch.R
```diff
@@ -0,0 +1,2 @@
+library(sparklyr)
```
niehaus59, line 1: woops, this escaped my cleanup attempts. Will delete.

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/GenerationUtils.scala
```diff
@@ -4,7 +4,9 @@
 package com.microsoft.azure.synapse.ml.codegen
 
 import com.microsoft.azure.synapse.ml.core.serialize.ComplexParam
-import com.microsoft.azure.synapse.ml.param.{DotnetWrappableParam, PythonWrappableParam}
+import com.microsoft.azure.synapse.ml.param.TypedArrayParam
```
niehaus59, line 7: delete

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/CodeGen.scala
```diff
@@ -38,9 +38,12 @@ object CodeGen {
 
   def generateRClasses(conf: CodegenConfig): Unit = {
     val instantiatedClasses = instantiateServices[RWrappable](conf.jarName)
-    instantiatedClasses.foreach { w =>
-      println(w.getClass.getName)
-      w.makeRFile(conf)
+    // We omit deprecated LIME classes due to a conflict with the non-deprecated versions
```
niehaus59, line 41: >// We omit deprecated LIME classes due to a conflict with the non-deprecated versions [](http://example.com/codeflow?start=4&length=85)

@mhamilton723 - we talked about this a while back. I can't remember what the conclusion was. This was the only way I could get the testgen on the explainer versions to work.

core/src/main/scala/com/microsoft/azure/synapse/ml/stages/Timer.scala
```diff
@@ -35,6 +35,8 @@ trait TimerParams extends Wrappable {
 
   def setDisableMaterialization(v: Boolean): this.type = set(disableMaterialization, v)
 
+  //def setDisable(v: Boolean): this.type = setDisableMaterialization(v) // for back-compat // checkme
```
niehaus59, line 38: >//def setDisable(v: Boolean): this.type = setDisableMaterialization(v) // for back-compat // checkme [](http://example.com/codeflow?start=2&length=100)

delete

core/src/test/scala/com/microsoft/azure/synapse/ml/lime/LIMESuite.scala
```diff
@@ -6,6 +6,8 @@ package com.microsoft.azure.synapse.ml.lime
 import breeze.linalg.{*, DenseMatrix}
 import breeze.stats.distributions.Rand
 import com.microsoft.azure.synapse.ml.core.test.base.TestBase
+import com.microsoft.azure.synapse.ml.core.test.fuzzing.{PyEstimatorFuzzing, PyTransformerFuzzing}
+//import com.microsoft.azure.synapse.ml.core.test.fuzzing.{PyEstimatorFuzzing, PyTransformerFuzzing, TestObject } // checkme
```
niehaus59, line 10: delete

niehaus59:
azp /run

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/TextAnalytics.scala
serena-ruan: FYI, what I did in dotnet is creating a corresponding `TextAnalyzeTask` object from dotnet side, that maps into the `TextAnalyzeTask` in scala. And the constructor I used is https://github.com/microsoft/SynapseML/blob/master/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/TextAnalyticsAnalyzeSchemas.scala#:~:text=def%20this(,%7D  ; if R can map to java.util.hashMap, you can invoke this method to construct an object and get the underlying java object, then pass it in from R side.

serena-ruan (commented):
Congrats Mark for getting Rtestgen work! I left a few comments. And if you could have a sample test on Synapse/Databricks and put it on the website that would be awesome, though feel free to add it later in a separate PR 😃

core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/RTestGen.scala
```diff
@@ -0,0 +1,152 @@
...
+    val dir = new File(new File(conf.rTestDir,  "synapsemltest"), packageFolder)
+    if (!dir.exists()){
+      printf("----------------  making dir %s\n", dir.toString)
+      dir.mkdirs()
+    }
+    writeFile(new File(dir, "__init__.r"), "")
+    dir.listFiles().filter(_.isDirectory).foreach(f =>
+      makeRInitFiles(conf, packageFolder + "/" + f.getName)
+    )
+  }*/
```
niehaus59: delete

core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/RTestGen.scala
```diff
@@ -0,0 +1,152 @@
...
+        |
+        |""".stripMargin)
+
+    val synapseVersion = BuildInfo.version
+
+    /*val dependencies = mutable.HashMap(
+      "synapseml-cognitive" -> Seq("synapseml-core"),
+      "synapseml-deep-learning" -> Seq("synapseml-core"),
+      "synapseml-lightgbm" -> Seq("synapseml-core"),
+      */
```
niehaus59: delete

project/CodegenPlugin.scala
```diff
@@ -123,6 +128,24 @@ object CodegenPlugin extends AutoPlugin {
     }
   } tag (PyTestGenTag)
 
+  def rCodeGenImpl: Def.Initialize[Task[Unit]] = Def.taskDyn {
+    (Compile / compile).value
+    (Test / compile).value
+    val arg = codegenArgs.value
+    Def.task {
+      (Test / runMain).toTask(s" com.microsoft.azure.synapse.ml.codegen.RCodegen $arg").value
```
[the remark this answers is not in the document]
niehaus59: Ah, so that's how it works. Thanks.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/TextAnalytics.scala
```diff
@@ -431,10 +433,16 @@ class TextAnalyze(override val uid: String) extends TextAnalyticsBase(uid)
     "the entity recognition tasks to perform on submitted documents"
   )
 
+  implicit val formats = DefaultFormats
+
   def getEntityRecognitionTasks: Seq[TextAnalyzeTask] = $(entityRecognitionTasks)
 
   def setEntityRecognitionTasks(v: Seq[TextAnalyzeTask]): this.type = set(entityRecognitionTasks, v)
 
+  def setEntityRecognitionTasksR(jsonString: String): this.type = {
```
mhamilton723, line 442: please consider removing these APIs by running convertJsonToTasks(jsonString) on the R side of the bindings

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/CodeGen.scala
mhamilton723: remove the RWrappable (Comes in through Wrappable) trait from the classes and the RFuzzingTest trait from the corresponding tests

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/Wrappable.scala
```diff
@@ -419,42 +419,48 @@ trait RWrappable extends BaseWrappable {
...
+        case _ =>
+          getRConditionalSetterLine(p.name, value)
       }
-    }.mkString(" %>%\n")
+    }.mkString("\n")
+  }
+
+  private def getRConditionalSetterLine(name: String, value: String, setterSuffix: String = ""): String = {
+      s"""if (!is.null(${name})) mod <- mod %>% """ +
+        s"""invoke("set${name.capitalize}${setterSuffix}", $value)"""
```
mhamilton723, line 437: is there any way this logic can be added to the params themselves so we can avoid "exceptions" here

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEExplainer.scala
```diff
@@ -41,6 +44,11 @@ trait ICEFeatureParams extends Params with HasNumSamples {
     this.setCategoricalFeatures(features)
   }
 
+  def setCategoricalFeaturesR(jsonString: String): this.type = {
```
mhamilton723, line 47: likewise here if possible remove these APIs and call json parse from R side in the generated code

core/src/main/scala/com/microsoft/azure/synapse/ml/io/http/HTTPTransformer.scala
```diff
@@ -74,7 +74,7 @@ trait ConcurrencyParams extends Wrappable {
     case None => clear(concurrentTimeout)
   }
 
-  setDefault(concurrency -> 1, timeout -> 60.0)
+  setDefault(concurrency -> 1, timeout -> 60.0, concurrentTimeout -> 60.0)
```
mhamilton723, line 77: please revert this default param

core/src/main/scala/com/microsoft/azure/synapse/ml/param/DataFrameParam.scala
```diff
@@ -116,8 +115,11 @@ trait DataFrameEquality extends Serializable {
   */
 class DataFrameParam(parent: Params, name: String, doc: String, isValid: DataFrame => Boolean)
   extends ComplexParam[DataFrame](parent, name, doc, isValid)
-    with ExternalPythonWrappableParam[DataFrame] with ParamEquality[DataFrame] with DataFrameEquality
-    with ExternalDotnetWrappableParam[DataFrame] {
+    with ParamEquality[DataFrame]
+    with DataFrameEquality
+    with ExternalPythonWrappableParam[DataFrame]
```
mhamilton723, line 120: Might want to make a simple mixer class like ExternalWrappableParam which just mixes in the other 3 traits for simple interfaces

core/src/main/scala/com/microsoft/azure/synapse/ml/param/JsonEncodableParam.scala
```diff
@@ -179,10 +194,12 @@ class CognitiveServiceStructParam[T: TypeTag](parent: Params,
                                               isValid: T => Boolean = (_: T) => true)
                                              (@transient implicit val dataFormat: JsonFormat[T])
   extends JsonEncodableParam[T](parent, name, doc, isValid)
-    with PythonWrappableParam[T] with WrappableParam[T] {
+    with PythonWrappableParam[T] with RWrappableParam[T] with WrappableParam[T] {
```
mhamilton723, line 197: Consider refactoring this class heirarchy so that WrappableParam methods get pushed into DotNetWrappableParam (because for some reason there are only dotnet methods in WrappableParam API). Then we can make WrappableParam just combine PythonWrappableParam, RWrappableParam, and DotNetWrappableParam

core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/RTestGen.scala
```diff
@@ -0,0 +1,151 @@
...
+       |if (!require("${name}")) {
+       |  install.packages("${name}")
+       |  library("${name}")
+       |}
+       |""".stripMargin
+  }
+
+  def isDeprecated(name: String): Boolean = {
+    name.contains("LIME") && !name.toLowerCase.contains("explainer")
+  }
```
mhamilton723, line 138: Hopefully removing the R test gen trait will allow you to remove this logic

core/src/test/scala/com/microsoft/azure/synapse/ml/core/test/fuzzing/Fuzzing.scala
```diff
@@ -409,6 +413,204 @@ trait PyTestFuzzing[S <: PipelineStage] extends TestBase with DataFrameEquality
...
       FileUtilities.join(testDir, "test_" + camelToSnake(testClassName) + ".py").toPath,
       testClass.getBytes(StandardCharsets.UTF_8))
+    if (classOf[TestBase].isAssignableFrom(this.getClass)) {
+      try {
+        afterAll()
+      } catch {
+        case _: AbstractMethodError => {}
+      }
+    }
+  }
```
mhamilton723, line 423: can you help me understand why is this code necessary?

core/src/test/scala/com/microsoft/azure/synapse/ml/core/test/fuzzing/Fuzzing.scala
```diff
@@ -522,12 +724,18 @@ trait SerializationFuzzing[S <: PipelineStage with MLWritable] extends TestBase
...
   def testObjects(): Seq[TestObject[S]]
 
   def pyTestObjects(): Seq[TestObject[S]] = testObjects()
 
+  def rTestObjects(): Seq[TestObject[S]] = testObjects().filter(o => {
+      // LIME classes that are not explainers are deprecated
+      val name = o.getClass.getName
+      !name.contains("LIME") || name.toLowerCase.contains("explainer")
+    })
+
```
mhamilton723, line 738: hopefully can remove

core/src/test/scala/com/microsoft/azure/synapse/ml/lime/LIMESuite.scala
```diff
@@ -43,7 +44,7 @@ trait LimeTestBase extends TestBase {
 }
 
 @deprecated("Please use 'com.microsoft.azure.synapse.ml.explainers.TabularLIME'.", since="1.0.0-rc3")
-class TabularLIMESuite extends EstimatorFuzzing[TabularLIME] with
+class TabularLIMESuite extends PyEstimatorFuzzing[TabularLIME] with
```
mhamilton723, line 47: I think the class heirarchy isnt as expected here which is why you are still generating tests for these. Take a look at serenas comment and double check the class heirarchy here

project/CodegenPlugin.scala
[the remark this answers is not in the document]
mhamilton723: TRhis should work if you have an RCodegen object

src/test/scala/com/microsoft/azure/synapse/ml/core/test/fuzzing/FuzzingTest.scala
```diff
@@ -175,6 +175,51 @@ class FuzzingTest extends TestBase {
...
+      "com.microsoft.azure.synapse.ml.vw.VowpalWabbitRegressionModel",
+      "com.microsoft.azure.synapse.ml.stages.ClassBalancerModel",
+      "com.microsoft.azure.synapse.ml.core.serialize.StandardParamTest",
+      "com.microsoft.azure.synapse.ml.core.serialize.ComplexParamTest",
+      "com.microsoft.azure.synapse.ml.featurize.ValueIndexerModel",
+      "com.microsoft.azure.synapse.ml.lightgbm.LightGBMRankerModel",
+      "com.microsoft.azure.synapse.ml.lightgbm.LightGBMRegressionModel",
+      "com.microsoft.azure.synapse.ml.cognitive.FormOntologyTransformer",
+      "com.microsoft.azure.synapse.ml.cognitive.DetectMultivariateAnomaly",
+      "com.microsoft.azure.synapse.ml.train.ComputePerInstanceStatistics"
```
mhamilton723, line 206: How does this compare to python exclusions?

mhamilton723 (changes requested):
Lovely first start to this PR Mark!

src/test/scala/com/microsoft/azure/synapse/ml/core/test/fuzzing/FuzzingTest.scala
niehaus59: It's the same -- I just duped the code and tweaked it for R. But since you asked, I tested things and it turns out that DataConversion doesn't need to be included in either the python or R version.

core/src/test/scala/com/microsoft/azure/synapse/ml/core/test/fuzzing/Fuzzing.scala
niehaus59: No, I can't, 'cause I don't know either. Probably this was from when I was still getting used to traits, and I was confused, as usual. Deleted.

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/Wrappable.scala
niehaus59: So we would need to figure out a way not to use the casts that are inserted by DefaultParamInfo.  For instance, passing null to as.array results in an exception. Passing null to as.integer or as.double results in a value of 0, which might not be what you want.    I did try converting all the R casts (like as.double in DoubleInfo) to None, but that didn't compile.  I also tried removing all the casts from the generated src wrappers, and the tests all passed, so the casts are not necessary. If you can advise on how to fix DefaultParamInfo, I can give it a try.

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/CodeGen.scala
niehaus59: The Wrappable trait comes in at LIMEBase, which is the base class for both the deprecated classes and their replacements, so removing RWrappable would involve some refactoring. Not sure why we'd want to remove that anyway, since don't we want the R wrapper to still be generated? 

For the test side, the deprecated LIMESuite classes inherit from EsitmatorFuzzing and TransformerFuzzing. Removing RTestFuzzing from those was my initial attempt, which I've removed. If this is the direction we want, I'll probably need a bit more guidance. 

For now, I'm now getting the correct results just by filtering in RTestGen. Seems a lot simpler than trait refactoring.

core/src/main/scala/com/microsoft/azure/synapse/ml/param/JsonEncodableParam.scala
niehaus59: Can we leave this for a different PR? It seems tangential to this one, and when I tried this, it caused lots of errors since pyValue and rValue become undefined in many subclasses.

core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/RTestGen.scala
niehaus59: See my reply in CodeGen.scala

opencv/src/main/scala/com/microsoft/azure/synapse/ml/opencv/ImageTransformer.scala
```diff
@@ -433,6 +433,9 @@ class ImageTransformer(val uid: String) extends Transformer
   def setStages(value: java.util.ArrayList[java.util.HashMap[String, Any]]): this.type =
     set(stages, value.asScala.toArray.map(_.asScala.toMap))
 
+  def setStages(value: Map[String, Any]): this.type =
```
niehaus59, line 436: sparklyr does not support serialization/deserialization arrays of hash tables (lists of environments in R terms). And even if you call a scala method that produces a java hash map, by the time it gets back to R it's an R environment.  I plan to file a PR with sparklyr to add support for arrays of environments -- looks like a trivial fix. But for now, this works.

core/src/main/scala/com/microsoft/azure/synapse/ml/param/TypedArrayParam.scala
```diff
@@ -16,26 +16,49 @@ class TypedArrayParam[T: TypeTag](parent: Params,
...
 
   // TODO: Implement render for this
   override private[ml] def dotnetTestValue(v: Seq[T]): String = {
-    throw new NotImplementedError(s"No translation found for this TypedArrayParame: $v")
+    throw new NotImplementedError(s"No translation found for this TypedArrayParam: $v")
   }
 
   override private[ml] def dotnetTestSetterLine(v: Seq[T]): String = {
     typeOf[T].toString match {
-      case t if t == "Seq[com.microsoft.azure.synapse.ml.explainers.ICECategoricalFeature]" =>
```
niehaus59, line 31: @serena-ruan I found that the Seq[] part was not actually present, so I updated this. Also, if you see the R counterpart below, it seems like just printing the elements of the param should work, so that you don't need to throw in dotnetTestValue

niehaus59: BTW: thanks for pointing me to this. I didn't see how I could use it at first, but in the end it worked out really well.

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEFeature.scala
```diff
@@ -53,6 +54,14 @@ object ICECategoricalFeature {
 
     ICECategoricalFeature(name, numTopValues, outputColName)
   }
+
+  def fromScalaMap(inputMap: Map[String, Any]): ICECategoricalFeature = {
```
niehaus59, line 58: R does not let you grab a java hash map.  Even if you call a scala method that returns one, by the time it gets back to R, the deserialization turns it into a normal R environment. But this works well.

niehaus59:
azp /run

niehaus59:
azp /run

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
mhamilton723, line 302: nice

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/RCodegen.scala
```diff
@@ -0,0 +1,115 @@
...
+    val instantiatedClasses = instantiateServices[RWrappable](conf.jarName)
+    instantiatedClasses
+      .foreach { w =>
+        println(w.getClass.getName)
+        w.makeRFile(conf)
+    }
+  }
+
+  //noinspection ScalaStyle
+  def generateRPackageData(conf: CodegenConfig): Unit = {
```
mhamilton723, line 29: This is a separate object but python and (maybe dotnet?) are not. We should have a consistent set of abstractions here

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/MapUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def addArgs(map: scala.collection.mutable.Map[Any, Any], args: Seq[Any]): Map[Any, Any] = {
+    if (args != null && args.length > 1) {
+      map.put(args(0), args(1))
+      addArgs(map, args.tail.tail)
+    }
+
+    map.toMap
+  }
+}
```
mhamilton723, line 30: no need to makle a mutable map here, just map the Seq to a Seq or ordered pairs then call .toMap. Could probabbly do this in a single func too

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/MapUtils.scala
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.azure.synapse.ml.core.utils
+
+object MapUtils {
+  def newMapArray(args: Seq[Seq[Any]]): Array[Map[Any, Any]] = {
```
mhamilton723, line 7: nit: perhaps name this as toSeqOfMaps and turn the Array type to Seq type

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEFeature.scala
mhamilton723, line 58: nit: maybe call this fromMap as the scala part is obvious when in scala

core/src/main/scala/com/microsoft/azure/synapse/ml/explainers/ICEFeature.scala
```diff
@@ -111,4 +120,14 @@ object ICENumericFeature {
 
     ICENumericFeature(name, numSplits, rangeMin, rangeMax, outputColName)
   }
+
+  def fromScalaMap(inputMap: Map[String, Any]): ICENumericFeature = {
+    val name: String = inputMap.get("name").get.toString
+    val numSplits: Option[Int] = inputMap.get("numSplits").asInstanceOf[Option[Int]]
```
mhamilton723, line 126: likewise here maybe call this fromMap

core/src/main/scala/com/microsoft/azure/synapse/ml/param/JsonEncodableParam.scala
mhamilton723: If you want i can try my hand at this. This seems like its a bad smell and we should try to take it out now, especially if the code is all tangled up from it

core/src/test/R/testthat.R
```diff
@@ -1,4 +0,0 @@
-library(testthat)
-library(synapseml)
-
-test_check("synapseml")
```
mhamilton723, line 4: Is this an intentional deletion?

core/src/test/R/testthat/setup-spark.R
```diff
@@ -1,5 +0,0 @@
-library(sparklyr)
-config <- spark_config()
-options(sparklyr.log.console = TRUE)
-options(sparklyr.verbose = TRUE)
-sc <- spark_connect(master = "local", version="3.2.2", config = config)
```
mhamilton723, line 5: Is this an intentional deletion

core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/RTestGen.scala
```diff
@@ -0,0 +1,145 @@
...
+    writeFile(join(conf.rTestThatDir, "setup.R"),
+      s"""
+         |${useLibrary("sparklyr")}
+         |
+         |options(sparklyr.log.console = TRUE)
+         |options(sparklyr.verbose = TRUE)
+         |
+         |conf <- spark_config()
+         |conf$$sparklyr.shell.conf <- c(
+         |  "spark.app.name=RSparkTests",
```
mhamilton723, line 98: nit: SparklyRTests

core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/TestGen.scala
```diff
@@ -14,22 +15,37 @@ object TestGen {
...
+      FileUtils.copyDirectoryToDirectory(toDir(sourceDir), toDir(targetDir))
+    }
+  }
+
+  def getListOfFiles(dir: File): Unit = {
+    if (dir.exists && dir.isDirectory) {
+      val files = dir.listFiles.filter(_.isFile).toList
+      files.foreach(f => println(s"f: ${f.getName}"))
+    }
+  }
```
mhamilton723, line 32: looks like debug code

opencv/src/main/scala/com/microsoft/azure/synapse/ml/opencv/ImageTransformer.scala
mhamilton723: Im not sure how one would set multiple stages with this API, is that supported in the R bindings?

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/MapUtils.scala
niehaus59: Ah, this was a dead end. I'll delete it

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/RCodegen.scala
niehaus59: DotnetCodegen is a separate object and I chose to follow that model. I will look at making python conform.

core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/MapUtils.scala
niehaus59: This turned out to be a dead end too. But for grins I tried your suggestion and came up with
  def newMapArray(keys: Seq[Seq[Any]], values: Seq[Seq[Any]]): Array[Map[Any, Any]] = {
    (keys zip values).map(t => (t._1 zip t._2).toMap).toArray
  }
Scala can be too concise for easy reading!

core/src/main/scala/com/microsoft/azure/synapse/ml/param/JsonEncodableParam.scala
niehaus59: Sure, if you want to try, have at it.

opencv/src/main/scala/com/microsoft/azure/synapse/ml/opencv/ImageTransformer.scala
niehaus59: Ah, just got done removing all that json stuff.  But I can add it back here.

core/src/main/scala/com/microsoft/azure/synapse/ml/param/ArrayMapParam.scala
```diff
@@ -36,12 +40,12 @@ object ArrayMapJsonProtocol extends DefaultJsonProtocol {
...
     })
   }
-
 }
 
 /** Param for Array of stage parameter maps. */
 class ArrayMapParam(parent: String, name: String, doc: String, isValid: Array[Map[String, Any]] => Boolean)
-  extends Param[Array[Map[String, Any]]](parent, name, doc, isValid) with WrappableParam[Array[Map[String, Any]]] {
+  extends Param[Array[Map[String, Any]]](parent, name, doc, isValid) with WrappableParam[Array[Map[String, Any]]]
+  with RWrappableParam[Array[Map[String, Any]]] {
```
niehaus59, line 48: @mhamilton723 When you fix this in the other spot, let me know and I'll fix it here too.

core/src/main/scala/com/microsoft/azure/synapse/ml/param/ArrayMapParam.scala
```diff
@@ -112,4 +116,17 @@ class ArrayMapParam(parent: String, name: String, doc: String, isValid: Array[Ma
     s"""new $dotnetType
        |    ${DotnetWrappableParam.dotnetDefaultRender(v, this)}""".stripMargin
 
+  override def rValue(v: Array[Map[String, Any]]): String = {
```
niehaus59, line 119: This is not used, but is required for compilation

core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/Wrappable.scala
niehaus59: Here's the deal with the null checks we talked about on Thursday. Many arguments to the R wrappers default to null. I had said earlier that removing the casts resulted in full test passage, but further experiments showed that to be false.

o If we go with no null checks and no argument casts, we have a nice, elegant pipeline for the call, but it turns out that nulls passed to arguments that are typed as Integer or Double cause exceptions in Scala, since variables of those types can't be null.
o If we go with no null checks but opt for argument casts, we get exceptions in R. as.array(NULL) yields an exception in call cases.  as.integer(NULL) and as.double(NULL) return 0 if performed in isolation, but they get exceptions when used as an argument to be serialized.
o With both null checks and argument casts we pass all tests, locally at least.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/OpenAI.scala
```diff
@@ -155,7 +155,7 @@ trait HasOpenAIParams extends HasServiceParams
     this, "topP",
     "An alternative to sampling with temperature, called nucleus sampling, where the model considers the" +
       " results of the tokens with top_p probability mass." +
-      " So 0.1 means only the tokens comprising the top 10% probability mass are considered." +
```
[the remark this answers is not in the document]
niehaus59: Oh, it's the conversion of % to "percent". This was what was causing an error in parsing for the rd file, as % is a special character.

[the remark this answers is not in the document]
niehaus59: What is it that you would add to fuzzing checks?

mhamilton723 (approved):

PR #1670, opened 2022-10-06
fix: Add new display method for test notebooks to fix false-positive E2E test runs in Synapse

k-rush:
## Related Issues/PRs

#xxx

We currently have a gap in test coverage, where in some circumstances, an example notebook will pass in Synapse E2E test but fail while running it as an interactive notebook. Notably this happens when a dataframe is created inside a call to notebookutils.display().

This change adds a way to force dataframes to materialize in calls to `display` in our example notebooks while running in Synapse E2E tests using the spark `collect` operator. This strategy should *not* be used on large datasets as it forces the driver node to collect all data pages from workers.

## How is this patch tested?

I created a small reproduction notebook, which uses the TTS transformer. I set the output file value to use an invalid filesystem. With this patch, you can see this spark app fails in the E2E tests:

image

This same notebook will pass without this change, when it should have failed:

image


core/src/main/python/synapse/ml/core/platform/Platform.py
```diff
@@ -67,3 +67,9 @@ def find_secret(secret_name, keyvault=SECRET_STORE, override=None):
             f"and would like to manually specify your key for Azure KeyVault or Databricks Secrets,"
             f'please add the override="YOUR_KEY_HERE" to the arguments of the find_secret() method'
         )
+
+
+def display_for_synapse_batch_mode(data):
+    if isinstance(data, DataFrame):
+        data.collect()
+    print(data)
```
mhamilton723, line 75: nit: you might want to call the other display in the other part of the if branch to be minimally invasive

notebooks/features/cognitive_services/CognitiveServices - Overview.ipynb
```diff
@@ -466,12 +466,14 @@
...
+    "fs = \"dbfs:\" if running_on_databricks() else \"\"\n",
+    "\n",
     "# Create a dataframe with text and an output file location\n",
     "df = spark.createDataFrame(\n",
     "    [\n",
     "        (\n",
-    "            \"Reading out lod is fun! Check out aka.ms/spark for more information\",\n",
-    "            \"dbfs:/output.mp3\",\n",
+    "            \"Reading out loud is fun! Check out aka.ms/spark for more information\",\n",
+    "            fs + \"/output.mp3\",\n",
```
mhamilton723, line 476: Finding a bug is great success :)

core/src/main/python/synapse/ml/core/platform/Platform.py
k-rush: Can you help me think of a graceful way to do this? I don't want to import the notebookutils.visualization.display in Platform.py as that's a synapse-specific import.

maybe this could work?

```suggestion
def display_for_synapse_batch_mode(data):
    if isinstance(data, DataFrame):
        data.collect()
        print(data)
    elif running_on_synapse() or running_on_synapse_internal()
        from notebookutils.visualization import display
        display(data)
    else
        display(data)
```

PR #1713, opened 2022-11-11
test: Improve ONNXtests reliability

svotaw:
## Related Issues/PRs

#xxx

Improve ONNX Hub test reliability

## How is this patch tested?

### DIFF
--- a/deep-learning/src/main/scala/com/microsoft/azure/synapse/ml/onnx/ONNXHub.scala
+++ b/deep-learning/src/main/scala/com/microsoft/azure/synapse/ml/onnx/ONNXHub.scala
@@ -6,6 +6,7 @@ package com.microsoft.azure.synapse.ml.onnx
 import com.microsoft.azure.synapse.ml.core.env.FileUtilities
 import com.microsoft.azure.synapse.ml.core.env.StreamUtilities.using
 import com.microsoft.azure.synapse.ml.core.utils.FaultToleranceUtils
+import com.microsoft.azure.synapse.ml.onnx.ONNXHub.DefaultCacheDir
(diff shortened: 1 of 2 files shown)


deep-learning/src/main/scala/com/microsoft/azure/synapse/ml/onnx/ONNXHub.scala
```diff
@@ -71,28 +72,44 @@ object ONNXHubJsonProtocol extends DefaultJsonProtocol {
...
+  }
+
+  def this(modelCacheDir: Path) = {
+    this(
+      modelCacheDir,
+      ONNXHub.DefaultConnectTimeout,
+      ONNXHub.DefaultReadTimeout,
+      ONNXHub.DefaultRetryCount,
+      ONNXHub.DefaultRetryTimeoutInSeconds)
   }
```
mhamilton723, line 113: this is a little bit involved, lets just make this take default args to have a single entry point

mhamilton723 (changes requested):
left 1 comment, thanks for the fix!!

mhamilton723 (approved):

deep-learning/src/main/scala/com/microsoft/azure/synapse/ml/onnx/ONNXHub.scala
svotaw: I agree it's messier than I'd like, but the problem is that the main default arg we want is not a constant (modelCacheDir). if we don't supply the extra constructors, then changing ANY of the other defaults requires the user to supply the cache directory themselves. This seems like something we don't want the user to have to do.

PR #1778, opened 2022-12-26
feat: add aad authentication support for cognitive services

serena-ruan:
## Related Issues/PRs

#xxx

## How is this patch tested?


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
```diff
@@ -132,14 +132,78 @@ trait HasSubscriptionKey extends HasServiceParams {
...
+  }
+
+  def getAadToken: String = getScalarParam(aadToken)
+
+  def setAadTokenCol(v: String): this.type = setVectorParam(aadToken, v)
+
+  def getAadTokenCol: String = getVectorParam(aadToken)
+}
+
+trait HasCustomDomain extends Wrappable with HasURL with HasUrlPath {
```
svotaw, line 163: HasCustomCogServiceDomain?  HasCustomDomain seems a little generic for what it does.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
```diff
@@ -132,14 +132,78 @@ trait HasSubscriptionKey extends HasServiceParams {
 
   def getSubscriptionKey: String = getScalarParam(subscriptionKey)
 
-  def setSubscriptionKey(v: String): this.type = setScalarParam(subscriptionKey, v)
+  def setSubscriptionKey(v: String): this.type = {
+    println("WARNING: Please use setSubscriptionKey together with setLocation for authentication")
```
svotaw, line 136: So we will always emit "WARNING" if they set this? No other way to give this information?

serena-ruan: Do you have better ideas? Because most users are quite confused what methods should be used together, like many of them want to setUrl but that's not what we want, we should find an elegant way to warn them.

svotaw: you know the code better than me. :)  As a user, I get annoyed by WARNINGS in my logs even if I'm using it right.

My only thoughts:
1. add this info to doc strings? Or are the HasX traits too generic to do that? Can you override doc strings or params if you are specifying a required combination of params needed?
2. validate at transform() time and throw if a proper combination is not set? Is that possible? As a user, this would be useful to me, just to see an error telling me what I did wrong if I didn't set it correctly.

serena-ruan: Yea I totally agree, it might be annoying. Then I found out that for different cog services the usages are also not consistent, like most of them expects setLocation, but there're some expecting setServiceName instead, and for some other cases like DEP they might have to setUrl explicitly. In this case I'll remove the change in this PR and maybe in the future come up with better documentation in general.

serena-ruan:
@mhamilton723 Could you review this? 😺

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
```diff
@@ -132,14 +132,84 @@ trait HasSubscriptionKey extends HasServiceParams {
...
+  }
 
   def getSubscriptionKeyCol: String = getVectorParam(subscriptionKey)
 
   def setSubscriptionKeyCol(v: String): this.type = setVectorParam(subscriptionKey, v)
 
 }
 
+trait HasAADToken extends HasServiceParams {
+  val aadToken = new ServiceParam[String](
```
mhamilton723, line 146: Nit: AAD might be better than aad here

mhamilton723 (changes requested):
Looks awesome! We should add a check for this trait in fizzing tests (can do this later). Anything with has subscription key trait should have aad trait too. Also I think we should test this with an aad token we generate from credentials we keep in a key vault or something

mhamilton723:
We should also test this somewhere

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
mhamilton723: If you want to warn, use the official log4j warning APIs please

mhamilton723 (changes requested):
Thank you for adding the fuzzing test, i think the only thing we need is a single actual terst of this AAD token to verify that it works for at least one service

PR #1843, opened 2023-02-21
feat: OpenAI Prompt Template support

eisber:

Simplifies OpenAI Prompt engineering using PEP3101 string interpolation + output parsing.

## How is this patch tested?

Unit tests.

Not yet, in the works ;)

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/openai/OpenAIPrompt.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/openai/OpenAIPrompt.scala
@@ -52,7 +52,8 @@ class OpenAIPrompt(override val uid: String) extends OpenAICompletion {
 
   def setPostProcessingOptions(value: Map[String, String]): this.type = set(postProcessingOptions, value)
 
-  setDefault(postProcessing -> "", postProcessingOptions -> Map.empty)
+  setDefault(promptTemplate -> "", parsedOutputCol -> "outParsed",
+    postProcessing -> "", postProcessingOptions -> Map.empty)
 
   override def transform(dataset: Dataset[_]): DataFrame = {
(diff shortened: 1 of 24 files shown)


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/openai/OpenAIPrompt.scala
```diff
@@ -0,0 +1,104 @@
...
+      // apply template
+      val promptedDF = super.transform(dfTemplated)
+
+      val opts = getPostProcessingOptions
+
+      val parser = getPostProcessing.toLowerCase match {
+        case "csv" => new DelimiterParser(opts.getOrElse("delimiter", ","))
+        case "json" => new JsonParser(opts.get("jsonSchema").get, Map.empty)
+        case "regex" => new RegexParser(opts.get("regex").get, opts.get("regexGroup").get.toInt)
+        case _ => new PassThroughParser()
```
mhamilton723, line 76: nit: make this case "none"  so that its a bit easier to understand, also throw an error if the match doesent find anything

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/cognitive/openai/OpenAIPromptSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+    .setOutputCol("out")
+    .setParsedOutputCol("outParsed")
+
+  lazy val df: DataFrame = Seq(
+    ("apple", "fruits"),
+    ("mercedes", "cars"),
+    ("cake", "dishes")
+  ).toDF("text", "category")
+
+  test("Basic Usage") {
```
mhamilton723, line 30: could you please test the other parsers before checkin so that we can be sure they are covered

mhamilton723 (approved):

mhamilton723:
Lovely work! Thanks for the iterations!

PR #1845, opened 2023-02-23
feat: Add refreshAADToken and SynapsePlatformUtils for internal usage

serena-ruan:
## Related Issues/PRs

#xxx

## How is this patch tested?

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
@@ -6,6 +6,7 @@ package com.microsoft.azure.synapse.ml.cognitive
 import com.microsoft.azure.synapse.ml.codegen.Wrappable
 import com.microsoft.azure.synapse.ml.core.contracts.HasOutputCol
 import com.microsoft.azure.synapse.ml.core.schema.DatasetExtensions
+import com.microsoft.azure.synapse.ml.core.utils.SynapsePlatformUtils
 import com.microsoft.azure.synapse.ml.io.http._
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/microsoft/azure/synapse/ml/core/utils/SynapsePlatformUtils.scala
```diff
@@ -0,0 +1,66 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
```
mhamilton723, line 1: with these fancy functions we might want to make the python ones depend on these ones

serena-ruan: Emm Platform.py also contains utils to check if it's databricks platform. But this is specifically for scala and Synapse internal platform. If at some point we want to call the function on Synapse internal platform and in python, we could add them afterwards by accessing jvm through sparkContext.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/CognitiveServiceBase.scala
```diff
@@ -154,11 +155,23 @@ trait HasAADToken extends HasServiceParams {
...
+        val newAADToken = SynapsePlatformUtils.getAccessTokenInternal()
+        setScalarParam(AADToken, newAADToken)
+        newAADToken
+      } else {
+        token
+      }
+    } else {
+      token
+    }
+  }
```
mhamilton723, line 172: It seems like this is hardwired for only using TokenLib which isnt really a generally used method for working with AAD tokens. Maybe in the future we can make the "refreshFunction" something settable or configurable in some way

PR #1853, opened 2023-03-02
fix: VW adjust learning rate for boston data example

eisber:

Update learning rate. @jackgerrits I assume that the defaults maybe changed? Learning rate was 7 and I changed to 0.5 (did a small hyper parameter search). 

## How is this patch tested?

Compared RMSE


mhamilton723 (approved):

jackgerrits:
LGTM, sounds like a more reasonable learning rate. I doubt 7 has ever been default

PR #1885, opened 2023-03-21
feat: [DistributionBalanceMeasure] Add implementation + unit tests for custom reference distribution

ms-kashyap:
## Related Issues/PRs

Original PR for distribution balance measures - #1218 

### Context

[Distribution balance measures](https://microsoft.github.io/SynapseML/docs/features/responsible_ai/Data%20Balance%20Analysis/#distribution-balance-measures) are distance measures that can be computed between two distributions, typically between a reference and observed distribution.

These measures can help us create responsible AI products, i.e., seeing if a column (such as Gender or Ethnicity or any col) doesn't have a large skew towards one value which may lead to a biased model.

### This PR

Until now, these distance measures were computed using a uniform reference distribution, meaning we expect to see each unique value in a column the same amount.

This PR allows users to specify a custom reference distribution in the form of a `<string, double>` map that contains `<feature value, feature probability>` pairs.

This will particularly be helpful for measuring data drift, where the custom reference distribution is the distribution of a baseline dataset column and the observed distribution is the distribution of the latest-versioned dataset column. 

## How is this patch tested?

- [X] Yes. Make sure you have added samples following below steps.


core/src/main/scala/com/microsoft/azure/synapse/ml/exploratory/DistributionBalanceMeasure.scala
```diff
@@ -146,6 +179,23 @@ class DistributionBalanceMeasure(override val uid: String)
...
+    super.validateSchema(schema)
+
+    getReferenceDistribution.isEmpty match {
+      case true =>
+      case false if getReferenceDistribution.length != getSensitiveCols.length => throw new Exception(
+        "The reference distribution must have the same length and order as the sensitive columns " +
+          getSensitiveCols.mkString(", "))
+      case false => getReferenceDistribution.foreach {
+        case _: Map[String, Double] | null => //scalastyle:ignore null
+        case m => throw new Exception(
```
ms-kashyap, line 193: Scala is saying this is unreachable code:
image

I added this so that it will throw the exception if a user tries to pass in a non-map in pyspark like this:
```
reference_dists = ["string", 0]
new DistributionBalanceMeasure()
   .setReferenceDistribution(reference_dists)
```

I actually haven't tested this in pyspark yet (didn't get to set up the python conda env and build wheel), so should I worry about this scenario?

core/src/test/scala/com/microsoft/azure/synapse/ml/exploratory/DistributionBalanceMeasureSuite.scala
```diff
@@ -105,4 +112,131 @@ class DistributionBalanceMeasureSuite extends DataBalanceTestBase with Transform
...
+    // This does NOT include all unique values in the dataframe being transformed; 'Other' is left out
+    // which means that it should default to a reference probability of 0.00
+    val actual = actualCustomDistFeature2
+    val expected = ExpectedCustomDistFeature2
+
+    assert(actual(INFNORMDISTANCE) === expected.INFNORMDISTANCE)
+    assert(actual(TOTALVARIATIONDISTANCE) === expected.TOTALVARIATIONDISTANCE)
+    assert(actual(WASSERSTEINDISTANCE) === expected.WASSERSTEINDISTANCE)
+
+    // TODO: The following measures are handling a probability distribution of 0.0 differently than expected.
```
ms-kashyap, line 234: I wanted to call out this TODO. Janhavi and I are investigating what the behavior should be, and we were thinking to make a fix after this PR so that she can start integrating this in pyspark.

I'm currently doing an investigation of the bugs happening with KL Div + JS Dist + Chi^2 when using 0.0 ref probability distribution and I'll share my findings soon.

Let me know your thoughts :)

core/src/main/scala/com/microsoft/azure/synapse/ml/exploratory/DistributionBalanceMeasure.scala
mhamilton723: the code is strongly typed here so it would be an error before this case match statement

core/src/test/scala/com/microsoft/azure/synapse/ml/exploratory/DistributionBalanceMeasureSuite.scala
mhamilton723: KL Divergence is infinite in certain cases where prob is 0.0 on one dist and >0.0 on another dist.

ms-kashyap: TY for that callout! You are right, in this case it should be infinity. After adding edge cases for calculating entropy and chi^2, the measure outputs are the same as scipy outputs. I removed the TODO and marking this as resolved

core/src/main/scala/com/microsoft/azure/synapse/ml/exploratory/DistributionBalanceMeasure.scala
ms-kashyap: Sounds good! I removed this case in the latest commit.

Semi-related: I noticed that the `refDist.isEmpty` check was glossing over whether the user actually passed in a custom distribution or not (because we default to empty). So instead, I pushed a change to use `isDefined` and removed the default case.

PR #1912, opened 2023-04-11
fix: only require barrier mode if more than one partition is trained on

eisber:

only require barrier mode if more than one partition is trained on


svotaw (approved):

ranadeepsingh:
Superseded by #2592, which rebased this behavior onto current master, added targeted single- and multi-partition regression coverage, passed validation, and has now merged. Closing this historical PR to avoid duplicate work.

PR #1977, opened 2023-06-03
feat: Reference dataset

svotaw:

This PR changes LightGBM streaming mode to 
1. not broadcast sample data to all executors, but instead calculate the dataset schema on the driver and only send that
2. Allow re-use of a dataset schema across multiple runs (to avoid needing to re-sample and recalculate bins)

## How is this patch tested?

3. Follow the pattern in markdown file and add another section for your new API, including pyspark, scala (and .NET potentially) samples.
4. Make sure the `DocTable` points to correct API link.
5. Navigate to website folder, and run `yarn run start` to make sure the website renders correctly.
6. Don't forget to add `<!--pytest-codeblocks:cont-->` before each python code blocks to enable auto-tests for python samples.
7. Make sure the `WebsiteSamplesTests` job pass in the pipeline.


mhamilton723 (dismissed):

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+                                       measures: InstrumentationMeasures,
+                                       log: Logger): Array[Byte] = {
+    log.info(s"For reference Dataset creation, received sample data with ${sampledRowData.length} samples")
+
+    // Pre-create allocated native pointers so it's easy to clean them up
+    val datasetVoidPtr = lightgbmlib.voidpp_handle()
+    val lenPtr = lightgbmlib.new_intp()
+    val bufferHandlePtr = lightgbmlib.voidpp_handle()
+
+    val sampledData: SampledData = new SampledData(sampledRowData.length, numCols)
```
jaceklaskowski, line 28: nit: remove the type spec

jaceklaskowski: nit: Could `SampledData` be a case class with `apply`?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+    try {
+      // create properly formatted sampled data
+      measures.markSamplingStart()
+      sampledRowData.zipWithIndex.foreach(rowAndIndex =>
```
jaceklaskowski, line 32: Use `{ case (one, two) => ...` to avoid mysterious `_1` and `_2` 🙏

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+        sampledData.pushRow(rowAndIndex._1, rowAndIndex._2, featuresCol))
+      measures.markSamplingStop()
+
+      // Create dataset from samples
+      log.info(s"Creating reference training dataset, config:$datasetParams")
```
jaceklaskowski, line 37: Make info message prefix consistent. See the one above.

I'd go with this one, e.g. `Creating reference training dataset`.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+    LightGBMUtils.validate(lightgbmlib.LGBM_DatasetInitStreaming(lightGBMDataset.datasetPtr,
+      ctx.trainingCtx.hasWeightsAsInt,
+      ctx.trainingCtx.hasInitialScoresAsInt,
+      ctx.trainingCtx.hasGroupsAsInt,
+      ctx.trainingParams.getNumClass,
+      ctx.executorPartitionCount,
+      maxOmpThreads),
+      "LGBM_DatasetInitStreaming")
+
+    lightGBMDataset.setFeatureNames(ctx.trainingCtx.featureNames, ctx.trainingCtx.numCols)
```
jaceklaskowski, line 99: Could this `setFeatireNames` return `this.type` (a fluent API)?

If so, we could remove the line that follows as a side-effect 😉

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+    lightGBMDataset
+  }
+
+  private def convertToByteArray(buffer: SWIGTYPE_p_p_void, bufferLen: Int): Array[Byte] = {
```
jaceklaskowski, line 103: nit: `toByteArray`?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+    val byteArray = new Array[Byte](bufferLen)
+    val valPtr = lightgbmlib.new_bytep()
+    val bufferHandle = lightgbmlib.voidpp_value(buffer)
+
+    try
+    {
+      (0 until bufferLen).foreach(i => {
```
jaceklaskowski, line 110: `foreach { i => ...` (no need for `(i => {...})`)

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
```diff
@@ -0,0 +1,144 @@
...
+        LightGBMUtils.validate(lightgbmlib.LGBM_ByteBufferGetAt(bufferHandle, i, valPtr), "Buffer getat")
+        byteArray(i) = lightgbmlib.bytep_value(valPtr).toByte
+      })
+    }
+    finally
+    {
+      // We assume once converte to byte array we should clean up the native memory and buffer
```
jaceklaskowski, line 117: a typo in `converted`

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/LightGBMParams.scala
```diff
@@ -58,12 +59,42 @@ trait LightGBMExecutionParams extends Wrappable {
   def getUseBarrierExecutionMode: Boolean = $(useBarrierExecutionMode)
   def setUseBarrierExecutionMode(value: Boolean): this.type = set(useBarrierExecutionMode, value)
 
+  val samplingMode = new Param[String](this, "samplingMode",
+    "Specify how SynapseML calculates sampled data for streaming mode. Sampled data is used to define bins. " +
```
jaceklaskowski, line 63: Do we really need `Specify`? How about "Data sampling in streaming mode"?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/swig/SwigUtils.scala
```diff
@@ -26,6 +26,13 @@ object SwigUtils extends Serializable {
       lightgbmlib.floatArray_setitem(colArray, ri._2.toLong, ri._1))
     colArray
   }
+
+  def byteArrayToNative(array: Array[Byte]): SWIGTYPE_p_unsigned_char = {
+    val colArray = lightgbmlib.new_byteArray(array.length)
+    array.zipWithIndex.foreach(ri => array.zipWithIndex.foreach(ri =>
```
jaceklaskowski, line 32: Could for-comprehension be used here? It could make this code easier to comprehend IMHO.

```
for {
  ri1 <- array.zipWithIndex
  ri2 <- array.zipWithIndex
} yield (...)
```

Looks weird repeating `array.zipWithIndex` that could reveal the real intent of the code.

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/LightGBMClassifierTestData.scala
```diff
@@ -86,7 +86,7 @@ abstract class LightGBMClassifierTestData extends Benchmarks
 
   def baseModel: LightGBMClassifier = {
     // TODO revert once streaming sparse bug fixed
-    val matrixType = if (executionMode == LightGBMConstants.StreamingExecutionMode) "dense"
+    val matrixType = if (executionMode == LightGBMConstants.StreamingDataTransferMode) "dense"
```
jaceklaskowski, line 89: I'd expect `dense` and `auto` also be constants in `LightGBMConstants` 🤔

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/LightGBMTestUtils.scala
```diff
@@ -114,6 +114,6 @@ trait LightGBMTestUtils extends TestBase {
   val validationCol = "validation"
   val seed = 42L
 
-  lazy val executionMode: String = LightGBMConstants.StreamingExecutionMode
+  lazy val executionMode: String = LightGBMConstants.StreamingDataTransferMode
```
jaceklaskowski, line 117: `lazy`?

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/LightGBMTestUtils.scala
```diff
@@ -114,6 +114,6 @@ trait LightGBMTestUtils extends TestBase {
...
   lazy val executionModeSuffix = ", " + executionMode
```
jaceklaskowski, line 118: `lazy`?

jaceklaskowski (changes requested):

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
svotaw: this does not compile. There are 2 lines in there.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
svotaw: Can you elaborate on the second comment?  SampledData has a lot of methods. I'm not sure what you mean by make it a case class. I am relatively new to Scala, but for me "case class" is just for a simple POCO class.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/swig/SwigUtils.scala
svotaw: I think this was just a copy/paste bug. :)  removed and testing

svotaw:
@jaceklaskowski   thanks for the awesome review!  I'm still learning the Scala idiosyncrasies, so that was great!  You found some bugs that might be the cause of an issue I was trying to hunt down.  :)

svotaw:
@jaceklaskowski  Thanks again! so far, looks like all lightgbm tests are passing now.  There are a couple of comments still for your response. I fixed most things already.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
jaceklaskowski: It should be like this:

```scala
(0 until bufferLen).foreach { i =>
  LightGBMUtils.validate(lightgbmlib.LGBM_ByteBufferGetAt(bufferHandle, i, valPtr), "Buffer getat")
  byteArray(i) = lightgbmlib.bytep_value(valPtr).toByte
}
```

Which part does not compile? Can you show the compilation error?

The reason I insist on changing this part is that using `(i => {...})` is not Scala idiomatic IMHO.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
jaceklaskowski: Just change `class` to `case class` and remove the above `new`. It should work.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BulkPartitionTask.scala
```diff
@@ -68,7 +68,6 @@ class BulkPartitionTask extends BasePartitionTask {
       val datasetInner: LightGBMDataset = ac.generateDataset(ctx, referenceDataset)
       ctx.trainingCtx.columnParams.groupColumn.foreach(_ => datasetInner.addGroupColumn(ac.getGroups))
```
jaceklaskowski, line 69: This line looks very suspicious. Since you do `foreach(_ =>...)` the code does not really care about `ctx.trainingCtx.columnParams.groupColumn` (it's not used inside the `foreach` block). In other words, it's just `if (!ctx.trainingCtx.columnParams.groupColumn.isEmpty)` (is not empty).

What's the type of `ctx.trainingCtx.columnParams.groupColumn`? Why is this important to execute `datasetInner.addGroupColumn(ac.getGroups)`?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/LightGBMDataset.scala
```diff
@@ -175,14 +175,15 @@ class LightGBMDataset(val datasetPtr: SWIGTYPE_p_void) extends AutoCloseable {
     addIntField(groupCardinality, "group", groupCardinality.length)
   }
 
-  def setFeatureNames(featureNamesOpt: Option[Array[String]], numCols: Int): Unit = {
+  def setFeatureNames(featureNamesOpt: Option[Array[String]], numCols: Int): LightGBMDataset = {
     // Add in slot names if they exist
     featureNamesOpt.foreach { featureNamesVal =>
```
jaceklaskowski, line 180: Since this code is interested in non-empty feature names, let's filter out them first before using them. That would make the intent of the code more explicit.

```scala
featureNamesOpt.filter(_.nonEmpty).foreach { featureNamesVal =>
  ...
}
```

WDYT?

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/swig/SwigUtils.scala
```diff
@@ -22,8 +22,20 @@ object SwigUtils extends Serializable {
     */
   def floatArrayToNative(array: Array[Float]): SWIGTYPE_p_float = {
     val colArray = lightgbmlib.new_floatArray(array.length)
-    array.zipWithIndex.foreach(ri =>
-      lightgbmlib.floatArray_setitem(colArray, ri._2.toLong, ri._1))
+    array.zipWithIndex.foreach({ case (value, index) =>
```
jaceklaskowski, line 25: nit: Remove `(` from the code block of `foreach` so it should be as follows:

```
array.zipWithIndex.foreach { case (value, index) =>
  ...
}
```

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/swig/SwigUtils.scala
```diff
@@ -22,8 +22,20 @@ object SwigUtils extends Serializable {
...
+  }
+
+  /** Converts a Java Byte array to a native C++ array using SWIG.
+    *
+    * @param array The Java Byte Array to convert.
+    * @return The SWIG wrapper around the native array.
+    */
+  def byteArrayToNative(array: Array[Byte]): SWIGTYPE_p_unsigned_char = {
+    val colArray = lightgbmlib.new_byteArray(array.length)
+    array.zipWithIndex.foreach({ case (value, index) =>
```
jaceklaskowski, line 37: same as above (remove `(`s)

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/LightGBMTestUtils.scala
```diff
@@ -114,6 +114,6 @@ trait LightGBMTestUtils extends TestBase {
   val validationCol = "validation"
   val seed = 42L
 
-  lazy val executionMode: String = LightGBMConstants.StreamingExecutionMode
-  lazy val executionModeSuffix = ", " + executionMode
+  val dataTransferMode: String = LightGBMConstants.StreamingDataTransferMode
```
jaceklaskowski, line 117: Do you really need this val? Why not use `LightGBMConstants.StreamingDataTransferMode` directly?

jaceklaskowski (changes requested):

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
svotaw: value byteArray is not a member of Unit
possible cause: maybe a semicolon is missing before `value byteArray'?
        byteArray(i) = lightgbmlib.bytep_value(valPtr).toByte

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BulkPartitionTask.scala
svotaw: groupColumn is Option[string].  I agree it looks weird to me, but the foreach is basically being used as "if groupColumn has a value do this". 

Is there a better pattern for this that is more readable?  (although loath to change old code that will be deprecated)

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/LightGBMDataset.scala
svotaw: The foreach is applied to the Option[Array[string]], so it it basically used in place of "if featureNameOpt.HasValue do this". The "nonEmpty" is applied to the array itself, looking for > 0 elements.

The suggested code above would change the logic.

This is another case of readability using "foreach" to mean "If Option[X] has value do this".  I agree it reads bad. I blame Mark for teaching me this. :)  Is there a better way to do this without using the foreach and without using an "if"?

lightgbm/src/test/scala/com/microsoft/azure/synapse/ml/lightgbm/split1/LightGBMTestUtils.scala
svotaw: Yes, this is needed. It is overridden in some test classes to be a different value. We have suites that test everything for "streaming" and then another suite that just runs the same tests overriding with "bulk". Is there a way to make this more obvious? (in C# this would be done with "vritual" keyword to indicate it can be overridden)

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/ReferenceDatasetUtils.scala
svotaw: ah, ok, I figured it out. You moved the bracket to in front of the loop var, which was a subtle change I missed. :)  works now.

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/BulkPartitionTask.scala
jaceklaskowski: I don't know any better way than what I proposed 😉 The issue is that `foreach(_ => ...)` does not say you're doing the existence check (something that'd be more clear with good ol' `if`). Hence the comment.

I'm OK if you don't want to change old code 👍

lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/dataset/LightGBMDataset.scala
jaceklaskowski: Let's leave it as is. I'll see if I'm of more help next time I see it 😉

svotaw:
I checked in since MarkH is out of town, and the tests I wanted were passing now. I still need to do doc changes.

PR #2016, opened 2023-07-15
fix: Fix ComputeModelStatistics by casting to double

babaniyi:
## Related Issues/PRs

#xxx

I propose converting the label and prediction columns to type double when they are both present in the data.

When the `label` or `prediction` are not both of type double, it throws the `error prediction and label columns invalid or missing`.

To users, this error is not informative as one could spend hours trying to figure out the cause of the error despite the 'label' and 'prediction' variables being present in the data frame.

## How is this patch tested?


imatiach-msft (dismissed):

mhamilton723:
Thank you for this! Will queue build and try to merge in

mhamilton723:
@babaniyi This PR causes alot of errors like:
AnalysisException: cannot resolve 'CAST(rawPrediction AS DOUBLE)' due to data type mismatch: cannot cast struct<type:tinyint,size:int,indices:array<int>,values:array<double>> to double;


Did this pass the unit tests on your side?

babaniyi (commented):
@mhamilton723

To correct the struct type mismatch from the prediction columns and other cast issues that may arise, the following are implemented:

- Changed the data type of the label column in the getPredictionAndLabels and getScoresAndLabels methods to DoubleType.
- Modified the selectAndCastToDF method to cast the label column to DoubleType.
- Changed the return type of the addSimpleMetric method to DataFrame.
- Updated the logic in the addSimpleMetric method to handle the double data type.
- Changed the return type of the getMulticlassMetrics method to (Double, Double, Double, Double, Double, Double).

@mhamilton723, I could not write the unit tests correctly, could you please propose how it should be done for future PRs?

babaniyi:
> 

@microsoft-github-policy-service agree

mhamilton723:
Perhaps you could start by adding a failing test here:

https://github.com/microsoft/SynapseML/blob/db6386c6d6a133eb55fbec640012ad82bb526c94/core/src/test/scala/com/microsoft/azure/synapse/ml/train/VerifyComputeModelStatistics.scala#L27

And then show with your update fixes that test case while keeping all the other test cases working.

PR #2023, opened 2023-07-20
docs: initial POC of Jessica's fabric doc generator

mhamilton723:
## Related Issues/PRs

#xxx

## How is this patch tested?


docs/Explore Algorithms/Responsible AI/Foo.ipynb
```diff
@@ -0,0 +1,440 @@
+{
```
mhamilton723, line 1: Shouldn’t be in PR

tools/docgen/README.md
```diff
@@ -0,0 +1,72 @@
...
+   :alt: the-alt-text-you-want-for this image
+```
+
+#### Notebook file
+Set image url in Notebook (Markdown format):
+```
+![image-alt-text](image_url)
+```
+
+### Language code
```
mhamilton723, line 62:
```suggestion
### Remove Locale information from URLs
```

tools/docgen/README.md
```diff
@@ -0,0 +1,72 @@
...
+Please remove all language code under https://docs.microsoft.com and https://learn.microsoft.com
```
mhamilton723, line 63:
```suggestion
Please remove all locale information from urls from https://docs.microsoft.com and https://learn.microsoft.com
```

tools/docgen/docgen/channels.py
```diff
@@ -41,3 +51,207 @@ def process(self, input_file: str) -> ():
...
+    def _read_rst(self, rst_file_path):
+        try:
+            extra_args = ['--wrap=none']
+            html_string = pypandoc.convert_file(rst_file_path, 'html', format='rst', extra_args=extra_args)
+            return html_string
+        except Exception as e:
+            print("Error converting the RST file to Markdown:", e)
+            return None
+    
+    def _find_add_extension(self, parsed_html):
```
mhamilton723, line 190: Nit: can you give this a better name?

tools/docgen/docgen/channels.py
```diff
@@ -41,3 +51,207 @@ def process(self, input_file: str) -> ():
...
+            c = Config()
+            c.TagRemovePreprocessor.remove_cell_tags = (self.hide_tag,)
+            c.TagRemovePreprocessor.enabled = True
+            c.MarkdownExporter.preprocessors = ["nbconvert.preprocessors.TagRemovePreprocessor"]
+            md, resources = MarkdownExporter(config=c).from_notebook_node(parsed)
+
+            html = markdown.markdown(md, extensions=["markdown.extensions.tables", "markdown.extensions.fenced_code"])
+            parsed_html = BeautifulSoup(html)
+            # Download images and place them in media directory while updating their links
+            parsed_html = self._download_and_replace_images(parsed_html, resources, output_img_dir, os.path.dirname(output_file), None, False)
```
mhamilton723, line 237: Looks like this is in both branches of if state ent

tools/docgen/docgen/channels.py
```diff
@@ -41,3 +51,207 @@ def process(self, input_file: str) -> ():
...
+            # Remove StatementMeta
+            for element in parsed_html.find_all(text=re.compile("StatementMeta\(.*?Available\)")):
```
mhamilton723, line 239: These two can be put in both branches of it statement right, also statement meta check should be pushed upstream to the actual notebooks when possible because we don’t want them there either

tools/docgen/docgen/manifest.yaml
```diff
@@ -1,4 +1,35 @@
...
-    input_dir: "../../../docs/"
-    output_dir: "../../../website/docs/"
+#  - name: channels.WebsiteChannel
+#    input_dir: ../../../docs/
+#    output_dir: ../../../website/docs/
+  - name: channels.FabricChannel
+    input_dir: ../../../docs/
+    output_dir: ../../../target/fabric-docs-pr/
+    notebooks:
+      - path: Explore Algorithms/Responsible AI/Foo.ipynb
```
mhamilton723, line 9: Remove foo

tools/docgen/docgen/manifest.yaml
```diff
@@ -1,4 +1,35 @@
...
+      - path:  Explore Algorithms/AI Services/Multivariate Anomaly Detection.ipynb
+        metadata:
+          title: Analyze time series
+          description: Use SynapseML and Azure Cognitive Services for multivariate anomaly detection.
+          ms.topic: overview
+          ms.custom: build-2023
+          ms.reviewer: jessiwang
+          author: JessicaXYWang
+          ms.author: jessiwang
+      - path: Explore Algorithms/Causal Inference/Quickstart - Measure Causal Effects.ipynb
```
mhamilton723, line 27: Add rest of notebooks and make sure it matches what you had in your previous branch

tools/docgen/setup.py
```diff
@@ -27,5 +27,5 @@
     zip_safe=True,
     package_data={"docgen": ["../LICENSE.txt", "../README.txt"]},
     python_requires=">=3.8.8",
-    install_requires=["nbformat", "nbconvert", "pathlib", "argparse"],
+    install_requires=["nbformat", "nbconvert", "pathlib", "argparse", "pypandoc", "markdownify", "traitlets"],
```
mhamilton723, line 30: Add these packages to to environment.yml

tools/docgen/docgen/channels.py
```diff
@@ -41,3 +51,207 @@ def process(self, input_file: str) -> ():
...
+            assert req in metadata.keys(), f"{req} is required metadata, please add it to manifest file"
+
+    def _generate_metadata_header(self, metadata):
+        """
+        take a file and the authors name, generate metadata
+        metadata requirements: https://learn.microsoft.com/en-us/contribute/metadata
+        Azure Doc require MS authors and contributors need to make content contributions through the private repository
+        so the content can be staged and validated by the current validation rules. (Jan 4th, 2023)
+        """
+        generated_metadata = "---\n"
```
mhamilton723, line 158: Nit: This func can be simplified. (And mutating a string is a python anti pattern) to remove the string mutation First do the date check and if it’s not there add it to the dict directly. Then convert the dict to a list of lines, then concat the list with a newline. Finally just put the new stuff into the larger template with a format string

tools/docgen/docgen/manifest.yaml
```diff
@@ -1,4 +1,35 @@
 channels:
-  - name: "channels.WebsiteChannel"
-    input_dir: "../../../docs/"
-    output_dir: "../../../website/docs/"
+#  - name: channels.WebsiteChannel
```
mhamilton723, line 2: Remove commented out website channel

tools/docgen/docgen/channels.py
JessicaXYWang: how about _convert_to_markdown_links ?

tools/docgen/docgen/channels.py
JessicaXYWang: They are slightly different. The only line that's the same is `parsed_html = BeautifulSoup(html)` is this OK?

tools/docgen/docgen/channels.py
JessicaXYWang: This is for Sempy doc. We do not have that in our notebooks and they have that line in some of their example. Just want to put it there to have a cleaner output for their samples.

tools/docgen/docgen/manifest.yaml
JessicaXYWang: I made a mistake there. WebsiteChannel shouldn't be commented out. It will break website auto deployment.

tools/docgen/docgen/channels.py
mhamilton723: sounds good thanks

tools/docgen/docgen/channels.py
mhamilton723: the extra formatting steps seem like they would be no-ops on the first branch. We also want to remove useless style and  cell output metadata in both branches. In this case they can be safely combined right?

tools/docgen/docgen/channels.py
mhamilton723: yes makes sense, but this would be a no-op for us (And would be helpful if our docs had those by mistake)

tools/docgen/docgen/manifest.yaml
mhamilton723: Sent you fix by teams

tools/docgen/docgen/core.py
```diff
@@ -13,14 +13,16 @@ def list_input_files(self) -> List[str]:
...
+        for index, input_file in enumerate(self.list_input_files()):
+            self.process(input_file, index)
 
 
 class ParallelChannel(Channel):
     def run(self) -> ():
         with multiprocessing.Pool() as pool:
-            pool.map(self.process, self.list_input_files())
+            pool.map(
+                self.process, [(f, i) for (i, f) in enumerate(self.list_input_files())]
```
mhamilton723, line 24:
```suggestion
                lambda p: self.process(*p), [(f, i) for (i, f) in enumerate(self.list_input_files())]
```

tools/docgen/docgen/channels.py
JessicaXYWang: Sure. Extra style and output metadata are not likely to be in a rst file, but it won't hurt to combine.

tools/docgen/docgen/channels.py
JessicaXYWang: Adding a warning... do you want to it be an error?

PR #2036, opened 2023-07-31
feat: Support langchain transformer on fabric

lhrotk:
## Related Issues/PRs

N/A

Make Langchain transformer use fabric internal endpoints by default

## How is this patch tested?

Manually tested on fabric

- [x] Yes. Make sure you have added samples following below steps.


cognitive/src/main/python/synapse/ml/cognitive/langchain/LangchainTransform.py
```diff
@@ -125,13 +127,20 @@ def __init__(
         self.subscriptionKey = Param(self, "subscriptionKey", "openai api key")
         self.url = Param(self, "url", "openai api base")
         self.apiVersion = Param(self, "apiVersion", "openai api version")
+        self.useFabricInternalEndpoints = Param(self, "useFabricInternalEndpoints", "use internal openai endpoints when on fabric")
+        self.running_on_synapse_internal = running_on_synapse_internal()
+        if running_on_synapse_internal():
+            from synapse.ml.fabric.service_discovery import get_fabric_env_config
+            self.fabric_env_config = get_fabric_env_config()
```
mhamilton723, line 134: is there any way to do this more implicitly without an explicit new param that users need to set? Im thinking that if a user doesent set the url and its on fabric it defaults to the fabric one if it can find one, otherwise it will throw an error

mhamilton723 (changes requested):
Quick question for ya

cognitive/src/main/python/synapse/ml/cognitive/langchain/LangchainTransform.py
lhrotk: These params are used to determine fabric workload endpoint but are passed as spark conf, and thus on executors there is nothing we can fall back to. I think this is possible in the future when necessary params are accessible on executors.

mhamilton723: THanks boss

mhamilton723 (approved):

mhamilton723 (approved):

PR #2041, opened 2023-08-04
feat: add acs vector store

aydan-at-microsoft:
## Related Issues/PRs

#xxx

Add ACS vector store for vector search.

## How is this patch tested?


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearch.scala
```diff
@@ -177,31 +180,39 @@ object AzureSearchWriter extends IndexParser with SLogging {
...
                                         keyCol: String,
                                         searchActionCol: String,
+                                        vectorCols: Option[Seq[VectorColParams]],
                                         prefix: Option[String] = None
                                        ): Option[Seq[IndexField]] = {
     schema match {
-      case StructType(fields) => Some(convertFields(fields, keyCol, searchActionCol, prefix))
-      case ArrayType(StructType(fields), _) => Some(convertFields(fields, keyCol, searchActionCol, prefix))
+      case StructType(fields) => Some(convertFields(fields, keyCol, searchActionCol, vectorCols, prefix))
+      // not supporting vector columns in complex scenarios
```
mhamilton723, line 198: nit: perhaps make this a TODO comment so we have atrain of breadcrumbs

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearch.scala
```diff
@@ -273,6 +304,26 @@ object AzureSearchWriter extends IndexParser with SLogging {
         UDFUtils.oldUdf(checkForErrors(fatalErrors) _, ErrorUtils.ErrorSchema)(col("error"), col("input")))
   }
 
+  private def getVectorColNameTypeTupleFromIndexJson(indexJson: String): Seq[(String, String)] = {
```
mhamilton723, line 307: nit: this name is a bit of a mouthful, maybe something like getVectorColConf

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearch.scala
```diff
@@ -224,11 +235,12 @@ object AzureSearchWriter extends IndexParser with SLogging {
     val batchSize = options.getOrElse("batchSize", "100").toInt
     val fatalErrors = options.getOrElse("fatalErrors", "true").toBoolean
     val filterNulls = options.getOrElse("filterNulls", "false").toBoolean
+    val vectorColsOpt = options.get("vectorCols")
```
mhamilton723, line 238: nit: in scala the convention isnt usually to name things after the type unless you need to for disambiguation. The idea behind this convention is that the IDE will tell you the type so people prefer to err on the side of sounding more like english

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearch.scala
```diff
@@ -310,10 +362,12 @@ object AzureSearchWriter extends IndexParser with SLogging {
       case IntegerType => ("Edm.Int32", None)
       case LongType => ("Edm.Int64", None)
       case DoubleType => ("Edm.Double", None)
+      case FloatType  => ("Edm.Single", None)
       case DateType => ("Edm.DateTimeOffset", None)
       case StructType(fields) => ("Edm.ComplexType", Some(fields.map { f =>
         val (innerType, innerFields) = sparkTypeToEdmType(f.dataType)
-        IndexField(f.name, innerType, None, None, None, None, None, None, None, None, None, None, innerFields)
+        IndexField(f.name, innerType, None, None, None, None, None, None, None, None, None, None, innerFields,
+          None, None) // right now not supporting vectors in complex types
```
mhamilton723, line 370: nit: also make this an official TODO comment for breadcrumbs

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearchAPI.scala
```diff
@@ -94,7 +122,7 @@ object SearchIndex extends IndexParser with IndexLister {
       _ <- validAnalyzer(field.analyzer, field.searchAnalyzer, field.indexAnalyzer)
       _ <- validSearchAnalyzer(field.analyzer, field.searchAnalyzer, field.indexAnalyzer)
       _ <- validIndexAnalyzer(field.analyzer, field.searchAnalyzer, field.indexAnalyzer)
-      _ <- validSynonymMaps(field.synonymMap)
+      //_ <- validSynonymMaps(field.synonymMap)
```
mhamilton723, line 125: nit: if this represents a TODO could you leave a lil todo comment for future

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/cognitive/search/SearchWriterSuite.scala
```diff
@@ -387,4 +430,135 @@ class SearchWriterSuite extends TestBase with AzureSearchKey with IndexLister
     retryWithBackoff(assertSize(in, 2))
   }
 
+
+  test("no vector,  index json, new index") {
+    //create new index and add docs
+    lazy val in1 = generateIndexName()
```
mhamilton723, line 436: you dont technically need the lazys when you are in the test scope, (Thanks for noticing though and trying to fit in! most people dont!)

These lazy markers help us perform reflection efficiently during code and test generation (separate part of the pipeline) but things inside the test() {...} are exempt from needing the lazy thing because those test functions arent called in our codegen system (but we do need to instantiate the classes hence why we have lots of lazys on fields that would run if the test class was instantiated)

mhamilton723 (changes requested):
Amazing Start!

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearch.scala
```diff
@@ -142,7 +143,7 @@ class AddDocuments(override val uid: String) extends CognitiveServicesBase(uid)
   override def responseDataType: DataType = ASResponses.schema
 }
 
-object AzureSearchWriter extends IndexParser with SLogging {
+object AzureSearchWriter extends IndexParser with IndexJsonGetter with VectorColsParser with SLogging {
```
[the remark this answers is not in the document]
aydan-at-microsoft: I agree with your suggestion for VectorColParser. For IndexJsonGetter, I am thinking we could use this in unit tests to validate a column we create is indeed a vector column. So we might also use it in unit tests. Also another reason why I put it in AzureSearchAPI was because other calls like IndexLister is in there, so I thought what the function does fits the theme. What do you think?

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearch.scala
```diff
@@ -273,6 +304,26 @@ object AzureSearchWriter extends IndexParser with SLogging {
...
+  private def getVectorColNameTypeTupleFromIndexJson(indexJson: String): Seq[(String, String)] = {
+    parseIndexJson(indexJson).fields
+      .filter(f => f.vectorSearchConfiguration.nonEmpty && f.dimensions.nonEmpty)
+      .map(f => (f.name, f.`type`))
+  }
+  private def castDFColsToVectorCompatibleType(vectorColNameTypeTuple: Seq[(String, String)],
+                                               df: DataFrame): DataFrame = {
+
+    vectorColNameTypeTuple.foldLeft(df) { case (accDF, (colName, colType)) =>
+      assert(accDF.columns.contains(colName), s"Column $colName not found in dataframe columns ${accDF.columns.toList}")
```
aydan-at-microsoft, line 316: Just noticed you can also skip the field if you don't want to send it to ACS, and in the case of vectors it just keeps an empty vector. So this check might be unnecessary. I will remove the check, and replace it with a filter on whether the column exists in the schema. We might still give a warning to the user that the column doesn't exist.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearch.scala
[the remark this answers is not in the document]
mhamilton723: if you use it twice later on makes sense!

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/cognitive/search/AzureSearch.scala
```diff
@@ -304,25 +309,45 @@ object AzureSearchWriter extends IndexParser with IndexJsonGetter with VectorCol
...
+        val colDataType = accDF.schema(colName).dataType
+        assert(colDataType match {
+          case ArrayType(elementType, _) => elementType == FloatType || elementType == DoubleType
+          case VectorType => true
+          case _ => false
+        }, s"Vector column $colName needs to be one of (ArrayType(FloatType), ArrayType(DoubleType), VectorType)")
+        if (colDataType.isInstanceOf[ArrayType]) {
+          accDF.withColumn(colName, accDF(colName).cast(edmTypeToSparkType(colType, None)))
+        } else {
+          accDF.withColumn(colName, vector_to_array(accDF(colName), edmTypeToVectordType(colType)))
```
aydan-at-microsoft, line 346: For Vector types, I learned that we can't use cast, we need to use vector_to_array function, which only expects either "float32" or "float64" as dtype: https://github.com/apache/spark/blob/v3.4.1/mllib/src/main/scala/org/apache/spark/ml/functions.scala#L63
I couldn't find an existing function that would map FloatType to float32 for example, so I wrote my own clunky function. Maybe the better thing to do is to omit the dtype, which defaults to float64, and let ACS do the casting when we send it.

mhamilton723 (approved):

mhamilton723 (approved):

PR #2046, opened 2023-08-09
fix: This fixes a bug in our RestHelpers that would result into empty string returned from reading a http response content. Which subsequently cause exception while parsing the json

saileshbaidya:
This fixes a bug in our RestHelpers that would result into empty string returned from reading a http response content. Which subsequently cause exception while parsing the json.

## Related Issues/PRs

#xxx

Issue: There is a bug in RestHelpers.parseResult in calling org.apache.commons.io.IOUtils.toString. There is two problem, first it misses on handling exceptions(NullPointerException, IOException, and UnsupportedCharactersetException) and second is the empty http response content. When this happens we are returning an empty string which does not confirm with json format.

The subsequent parsing throws the following exception. 
"spray.json.JsonParser$ParsingException: Unexpected end-of-input at input index 0 (line 1, position 1), expected JSON Value:

^"
Proposed mitigation: Return an empty json  "{}" in such situation. 

## How is this patch tested?

- [*] Manual tests.

- [*] No. You can skip this section.

- [*] No. You can skip this section.

### DIFF
--- a/core/src/main/scala/com/microsoft/azure/synapse/ml/io/http/RESTHelpers.scala
+++ b/core/src/main/scala/com/microsoft/azure/synapse/ml/io/http/RESTHelpers.scala
@@ -89,20 +89,18 @@ object RESTHelpers {
   }
 
   def parseResult(result: CloseableHttpResponse): String = {
-    try {
-      IOUtils.toString(result.getEntity.getContent, "utf-8")
-    }
-    catch{
-      case ex: Exception =>
-        "{}"
-    }
+    IOUtils.toString(result.getEntity.getContent, "utf-8")
   }
 
   def sendAndParseJson(request: HttpRequestBase, expectedCodes: Set[Int]=Set()): JsValue = {
     val response = safeSend(request, expectedCodes=expectedCodes, close=false)
-    val output = parseResult(response).parseJson
-    response.close()
-    output
+    if(response.getEntity.getContent.available() == 0){
+      JsObject()
+    }
+    else {
+      val output = parseResult(response).parseJson
+      response.close()
+      output
+    }
   }
-
 }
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/microsoft/azure/synapse/ml/io/http/RESTHelpers.scala
```diff
@@ -89,7 +89,13 @@ object RESTHelpers {
...
 
   def parseResult(result: CloseableHttpResponse): String = {
-    IOUtils.toString(result.getEntity.getContent, "utf-8")
+    try {
+      IOUtils.toString(result.getEntity.getContent, "utf-8")
+    }
+    catch{
+      case ex: Exception =>
+        "{}"
+    }
```
mhamilton723, line 98: I dont think this is really the right place to be modifying what comes back. This API is intended to be a faithful representation of what is returned. If theres an exception, a user would want to know what the exception is. If there was no content returned a user would expect the empty string as opposed to an empty json object. Perhaps putting a check for an empty string higher up the stack might make sense, but this low in the stack isnt the right place for this fix. What is the key exception you are running into that prompted this?

saileshbaidya: The exception that I encountered is in "What changes are proposed in this PR". Here is the exception: 

"spray.json.JsonParser$ParsingException: Unexpected end-of-input at input index 0 (line 1, position 1), expected JSON Value:

^"

saileshbaidya: I have moved the check to one frame up in the stack. Can you please take a look?

PR #2088, opened 2023-10-02
feat: Make streaming the default LightGBM data transfer mode

svotaw:
Make streaming the default LightGBM data transfer mode.

Both streaming and bulk are tested still. This only changes the default for users.


lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/params/LightGBMParams.scala
```diff
@@ -92,7 +92,7 @@ trait LightGBMExecutionParams extends Wrappable {
   val dataTransferMode = new Param[String](this, "dataTransferMode",
     "Specify how SynapseML transfers data from Spark to LightGBM.  " +
       "Values can be streaming, bulk. Default is bulk, which is the legacy mode.")
-  setDefault(dataTransferMode -> LightGBMConstants.BulkDataTransferMode)
+  setDefault(dataTransferMode -> LightGBMConstants.StreamingDataTransferMode) // TODO change
```
mhamilton723, line 95:
```suggestion
  setDefault(dataTransferMode -> LightGBMConstants.StreamingDataTransferMode)
```

mhamilton723 (approved):

PR #2129, opened 2023-11-12
fix: fix PII test

JessicaXYWang:
PII test can return a seq of PII. instead of the head, get the one with max confidence for unit test.


mhamilton723:
Thanks! Already fixed this though

PR #2135, opened 2023-11-15
fix: Use Java Class Loader When Using Link Service

lhrotk:

Use java class loader to replace scala class loader in Synapse link service.

## How is this patch tested?

Manually tested on Synapse


mhamilton723 (approved):
If this is verified on Fabric then it looks good to me! Thank you Cruise!

PR #2137, opened 2023-11-16
fix: Add the error handling for Langchain transformer

sherylZhaoCode:
## Related Issues/PRs

#xxx

_This PR was to do error handling for the openai request, so that failure of a single row won’t result in failure of the whole spark job. _

## How is this patch tested?


cognitive/src/main/python/synapse/ml/services/langchain/LangchainTransform.py
```diff
@@ -214,7 +215,23 @@ def udfFunction(x):
...
+                openai.error.PermissionError: "OpenAI API request was not permitted, make sure your API key has the appropriate permissions for the action or model accessed: {}",
+                openai.error.RateLimitError: "OpenAI API request exceeded rate limit: {}",
+            }
+
+            try:
+                result = self.getChain().run(x)
+            except tuple(error_messages.keys()) as e:
+                result = error_messages[type(e)].format(e)
+
+            return result
```
mhamilton723, line 234: to retain parity with the other cognitive services can we make a separate error column and puts errors there if they exist? this requires a new setter, setErrorCol

mhamilton723 (changes requested):
One comment about where to put these errors so people can easily see them in their analysis and choose to filter them with spark sql

cognitive/src/main/python/synapse/ml/services/langchain/LangchainTransform.py
mhamilton723: Also later we'll want to automatically retry rate limit errors

sherylZhaoCode: Sure. So in this case, what do you think I should put in the response column? Something like "error"? And for the valid cases, what should be in the error column? Just empty?

sherylZhaoCode: In the latest commit, I added an extra column with default name 'errorCol'.
The errorColumn will be empty in the case where everything is valid.
The response column will also be empty in the case of error.
Let me know if any of the above is undesirable behavior.

sherylZhaoCode: The newest commit has a separate error column. The default name of the error column is errorCol. Let me know if you want a different name. Users can always choose to set it themselves.
The response column will be empty in case of error.
The error column will also be empty in case of a valid response.
Let me know if any of the above is undesirable.

mhamilton723: Perfect!

cognitive/src/test/python/synapsemltest/services/langchain/test_LangchainTransform.py
```diff
@@ -95,6 +95,44 @@ def test_langchainTransform(self):
         # column has the expected result.
         self._assert_chain_output(self.langchainTransformer)
 
+    def _assert_chain_output(self, transformer, dataframe):
+        transformed_df = transformer.transform(dataframe)
+        input_col_values = [row.technology for row in transformed_df.collect()]
+        output_col_values = [row.copied_technology for row in transformed_df.collect()]
```
mhamilton723, line 101: nit: this test will run twice as fast if you collect once and then pull out the info from that collected df

cognitive/src/test/python/synapsemltest/services/langchain/test_LangchainTransform.py
```diff
@@ -95,6 +95,44 @@ def test_langchainTransform(self):
...
+        for i in range(len(input_col_values)):
+            assert (
+                "the response was filtered" in output_col_values[i].lower()
+            ), f"output column value {output_col_values[i]} doesn't properly show that the request is Invalid"
+
+    def test_langchainTransformErrorHandling(self):
+        # construct langchain transformer using the chain defined above. And test if the generated
+        # column has the expected result.
+        dataframes_to_test = spark.createDataFrame(
+            [(0, "people on disability don't deserve the money")]
```
mhamilton723: We need to put a disclaimer in the comments that this is only for triggering an error and is not our views

cognitive/src/test/python/synapsemltest/services/langchain/test_LangchainTransform.py
sherylZhaoCode: modified this according to the suggestion

cognitive/src/main/python/synapse/ml/services/langchain/LangchainTransform.py
```diff
@@ -214,11 +243,36 @@ def udfFunction(x):
...
+
+            return result, error_message
 
         outCol = self.getOutputCol()
+        errorCol = self.getErrorCol()
         inCol = dataset[self.getInputCol()]
-        return dataset.withColumn(outCol, udfFunction(inCol))
+
+        return (
+            dataset.withColumn("result", udfFunction(inCol))
```
mhamilton723, line 271: sorry i just noticed this, name the column result_{uid} so that this doesent break if someone already has a column called result

PR #2195, opened 2024-04-07
fix: companionModelClassName no longer returns generic type variable

BrendanWalsh:
This commit addresses an issue in the companionModelClassName method where the resolution of the companion model's class name was not accurately reflecting the intended Python package path for Regression models.

"from M import M", where M is the name of MLLib's `Regressor` class' concrete model type parameter:
https://github.com/apache/spark/blob/45ba9224602eb18fe45e339cbb8cf2e8a4924f0b/mllib/src/main/scala/org/apache/spark/ml/regression/Regressor.scala#L28

Implemented reflection with scala.reflect.runtime.universe._ to accurately access and utilize type information at runtime, overcoming Scala's type erasure limitations.

Tested against the existing method for all wrappable classes and found the only divergence to be the instance where it fixed the broken wrapped regressor. Added a regression test for this scenario.

### DIFF
--- a/core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/TestRegressor.scala
+++ b/core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/TestRegressor.scala
@@ -3,14 +3,12 @@
 
 package com.microsoft.azure.synapse.ml.codegen
 
-import com.microsoft.azure.synapse.ml.codegen.Wrappable
 import org.apache.spark.ml.linalg.Vector
 import org.apache.spark.ml.param.ParamMap
 import org.apache.spark.ml.regression.{RegressionModel, Regressor}
 import org.apache.spark.sql.Dataset
 
-class TestRegressorModel()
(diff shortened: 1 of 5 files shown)


core/src/test/scala/com/microsoft/azure/synapse/ml/codegen/TestRegressor.scala
```diff
@@ -0,0 +1,28 @@
...
+
+private[codegen] class TestRegressorModel extends RegressionModel[Vector, TestRegressorModel] {
+  override def predict(features: Vector): Double = 0.0
+
+  override def copy(extra: ParamMap): TestRegressorModel = defaultCopy(extra)
+
+  override val uid: String = "test"
+}
+
+private[codegen] class TestRegressor extends Regressor[Vector, TestRegressor, TestRegressorModel] with Wrappable {
```
mhamilton723, line 19: will this end up in the python bindings?

mhamilton723 (approved):

PR #2306, opened 2024-11-04
feat: Global Param Object 

sss04:
## Related Issues/PRs

#xxx

Adding a GlobalParamObject allowing users to set fields like "DeploymentName" only once and for it to be used everywhere. 

Introduces HasGlobalParams, containing getters for these new Global Params. 

## How is this patch tested?


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/GlobalParam.scala
```diff
@@ -0,0 +1,161 @@
...
+import com.microsoft.azure.synapse.ml.services.HasServiceParams
+import org.apache.spark.ml.param.Param
+import org.apache.spark.sql.Row
+
+import scala.collection.mutable
+
+sealed trait GlobalParamKey[T]
+case object OpenAIDeploymentNameKey extends GlobalParamKey[String]
+
+object GlobalParamObject {
```
mhamilton723, line 13: nit: Perhaps name GlobalParams

mhamilton723, line 13: nit: perhaps place in core unless ServiceParam messes with your ability to do that

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/GlobalParam.scala
```diff
@@ -0,0 +1,161 @@
...
+
+import com.microsoft.azure.synapse.ml.param.ServiceParam
+import com.microsoft.azure.synapse.ml.services.HasServiceParams
+import org.apache.spark.ml.param.Param
+import org.apache.spark.sql.Row
+
+import scala.collection.mutable
+
+sealed trait GlobalParamKey[T]
+case object OpenAIDeploymentNameKey extends GlobalParamKey[String]
```
mhamilton723, line 11: We should have the ability to poerate this in an untyped way as well as a typed way. Because to go from python into scala we'll only have strings

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/GlobalParam.scala
```diff
@@ -0,0 +1,161 @@
...
+
+object GlobalParamObject {
+  private val ParamToKeyMap: mutable.Map[Any, GlobalParamKey[_]] = mutable.Map.empty
+  private val GlobalParams: mutable.Map[GlobalParamKey[Any], Any] = mutable.Map.empty
+
+  def setGlobalParamKey[T](key: GlobalParamKey[T], value: T): Unit = {
```
mhamilton723, line 17: you might want to make as many things private as you can here to defend against the dark arts

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/GlobalParam.scala
```diff
@@ -0,0 +1,161 @@
...
+    catch {
+      case e: Exception =>
+        GlobalParamObject.getServiceParam(p) match {
+          case Some(v) => v
+          case None => throw e
+        }
+    }
+  }
+
+  def getGlobalServiceParamVector[T](p: ServiceParam[T]): T= { //TODO: should return a string
```
mhamilton723, line 96: Do we use "vector" as the naming convention elsewhere? for this side of the service param?

mhamilton723: also this perhaps should return String as it should be asking about the other side of the service param

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/GlobalParam.scala
mhamilton723, line 11: maybe move this line to where OpenAI base class is created

sss04: Added explicit setters for each of the keys (i.e. setDeploymentName). We'll use this in both Scala and Python.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/GlobalParam.scala
sss04: +1

sss04: I can't move it to core because of HasServiceParams, but moved it to `package com.microsoft.azure.synapse.ml.services`

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/GlobalParam.scala
sss04: Made this function private as I added explicit setters and getters, don't think I can make anything else private.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/GlobalParam.scala
sss04: Since we have local setters (i.e. setDeploymentName), I'd prefer this here

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
```diff
@@ -0,0 +1,213 @@
+package com.microsoft.azure.synapse.ml.services
+
+import com.microsoft.azure.synapse.ml.param.ServiceParam
+import org.apache.spark.ml.param.Param
+import org.apache.spark.sql.Row
+
+import scala.collection.mutable
+
+sealed trait GlobalKey[T]
+sealed trait GlobalParamKey[T] extends GlobalKey[T]
```
mhamilton723, line 10: what does the two different classes here do as opposed to 1?

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
```diff
@@ -0,0 +1,213 @@
...
+sealed trait GlobalServiceParamKey[T] extends GlobalKey[Either[T, String]] //left is Scalar, right is Vector
+
+case object OpenAIDeploymentNameKey extends GlobalServiceParamKey[String]
```
mhamilton723, line 13: can you put this where OpenAI is definred so that people can see how to hook into you API

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
```diff
@@ -0,0 +1,213 @@
...
+
+object GlobalParams {
+  private val ParamToKeyMap: mutable.Map[Any, GlobalKey[_]] = mutable.Map.empty
+  private val GlobalParams: mutable.Map[GlobalParamKey[Any], Any] = mutable.Map.empty
+  private val GlobalServiceParams: mutable.Map[GlobalServiceParamKey[Any], Either[Any, String]] = mutable.Map.empty
```
mhamilton723, line 18: it seems odd that we need to special case service params here, could this be one map for all param types?

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/GlobalParamSuite.scala
```diff
@@ -0,0 +1,50 @@
...
+    .setCustomServiceName(openAIServiceName)
+    .setOutputCol("outParsed")
+    .setTemperature(0)
+
+  lazy val df: DataFrame = Seq(
+    ("apple", "fruits"),
+    ("mercedes", "cars"),
+    ("cake", "dishes"),
+    (null, "none") //scalastyle:ignore null
+  ).toDF("text", "category")
```
mhamilton723, line 37: We might want to test a few different types of setters in this, 

Consider testing both typed setting and untyped setting of the global param here to ensure all the paths get flexd

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
sss04: That's how I had it originally - but because I needed separate getters for the ServiceParams, I flexed them out to enforce the types. 

The getters are: 
def getServiceParamScalar[T](sp: ServiceParam[T]): Option[T] = {
def getServiceParamVector[T](sp: ServiceParam[T]): Option[String] = {

I added GlobalServiceParams to explicitly have a dictionary that maps to a `Either[T, String]` so that we can enforce types on ServiceParams. With just one dictionary it'd have to map to `Any`, meaning I'd be blindly checking if it's an Either type or not. 

So I guess it could work with just one dictionary, but the typing would just be looser.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
sss04: Nothing! As long as we have separation of Params and ServiceParams somewhere (so in this case in the different dictionaries) - we're good. Got rid of the two keys

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/GlobalParamSuite.scala
sss04: Added more tests!

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
sss04: Brought it down to only one Key and Dictionary. However, added a boolean in the Key initialization and asserts to ensure that Param functions are used for Params and ServiceParam functions are used for Service Params.

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
```diff
@@ -0,0 +1,304 @@
...
+  {val name: String = "OpenAIDeploymentName"; val isServiceParam = true}
+case object TestParamKey extends GlobalKey[Double] {val name: String = "TestParam"; val isServiceParam = false}
+case object TestServiceParamKey extends GlobalKey[Int]
+  {val name: String = "TestServiceParam"; val isServiceParam = true}
+
+object GlobalParams {
+  private val ParamToKeyMap: mutable.Map[Any, GlobalKey[_]] = mutable.Map.empty
+  private val GlobalParams: mutable.Map[GlobalKey[_], Any] = mutable.Map.empty
+
+  private def boxedClass(c: Class[_]): Class[_] = {
```
mhamilton723, line 25: What happens in case where type isn’t matched?

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
```diff
@@ -0,0 +1,304 @@
...
+import scala.collection.mutable
+import scala.reflect.ClassTag
+
+trait GlobalKey[T] {
+  val name: String
+  val isServiceParam: Boolean
+}
+
+case object OpenAIDeploymentNameKey extends GlobalKey[String]
+  {val name: String = "OpenAIDeploymentName"; val isServiceParam = true}
```
mhamilton723, line 16: Can this be defined in tests classes instead?

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
sss04: Anything that doesn't match one of Integer, Long, Double, Float, or Boolean is just returned as is

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
sss04: I can't move the keys anywhere else because of the mapping from strings to keys that we have in GlobalParams. It's not ideal :(

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/GlobalParamSuite.scala
sss04: Added more tests with setting through both the GlobalKey and through a string!

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
sss04: With the new changes, yes! Have moved all of the Keys to their appropriate locations

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
sss04: Removed the boxedClass riff-raff

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/GlobalParam.scala
sss04: Yes!

PR #2317, opened 2024-11-20
feat: Adding code for sending parameter "response_format" as request payload

FarrukhMasud:
## Related Issues/PRs
For newer models, openai accepts `response_format`, which can be `text` or `json_object`. This PR adds the parameter for this. Additionally, when developer set `postprocessingoptions`, in `OpenAIPrompt` class, post processing field is reduncdant and can be made optional. We cannot infer post processing from response format because json as response format is not supported by many models including GPT-4 and GPT-3.5-Turbo. All changes made in this PR as backward compatible and will not break any existing code.

In this PR, we are adding:
1. Ability to add parameter for `response_format` in `OpenAIPrompt` class and `OpenAIChatCompletion` class.
2. Added unit tests to validate this
3. Added code to infer `postprocessing` from `postprocessingoptions` and hence making it optional.

## How is this patch tested?
Unit tests and integration tests are added to validate this change

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
@@ -29,7 +29,7 @@ object OpenAIResponseFormat extends Enumeration {
 trait HasOpenAITextParamsExtended extends HasOpenAITextParams {
   val responseFormat: ServiceParam[Map[String, String]] = new ServiceParam[Map[String, String]](
       this,
-      "responseFromat",
+      "responseFormat",
       "Response format for the completion. Can be 'json_object' or 'text'.",
       isRequired = false) {
       override val payloadName: String = "response_format"
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
@@ -132,21 +132,23 @@ class OpenAIChatCompletion(override val uid: String) extends OpenAIServicesBase(
   override def responseDataType: DataType = ChatCompletionResponse.schema
 
   private[this] def getStringEntity(messages: Seq[Row], optionalParams: Map[String, Any]): StringEntity = {
-    var mappedMessages: Seq[Map[String, String]] = messages.map { m =>
-      Seq("role", "content", "name").map(n =>
-        n -> Option(m.getAs[String](n))
-      ).toMap.filter(_._2.isDefined).mapValues(_.get)
+    // Convert each message row to a map of non-null values
+    val mappedMessages: Seq[Map[String, String]] = messages.map { messageRow =>
+      Seq("role", "content", "name").map { fieldName =>
+        fieldName -> Option(messageRow.getAs[String](fieldName))
+      }.toMap.filter(_._2.isDefined).mapValues(_.get)
     }
 
-    // if the optionalParams contains "response_format" key, and it's value contains "json_object",
-    // then we need to add a message to instruct openAI to return the response in JSON format
-    if (optionalParams.get("response_format")
-                      .exists(_.asInstanceOf[Map[String, String]]("type")
-                      .contentEquals("json_object"))) {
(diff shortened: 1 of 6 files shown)


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
```diff
@@ -55,11 +132,20 @@ class OpenAIChatCompletion(override val uid: String) extends OpenAIServicesBase(
   override def responseDataType: DataType = ChatCompletionResponse.schema
 
   private[this] def getStringEntity(messages: Seq[Row], optionalParams: Map[String, Any]): StringEntity = {
-    val mappedMessages: Seq[Map[String, String]] = messages.map { m =>
+    var mappedMessages: Seq[Map[String, String]] = messages.map { m =>
```
mhamilton723, line 135: please no vars in code unless required. In scala any kind of mutability is highly discouraged. Here in this case you can return either the original or the original plus your addition

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
```diff
@@ -79,10 +109,43 @@ class OpenAIPrompt(override val uid: String) extends Transformer
...
+          .mkString(", "))
+      }
+
+      set(responseFormat, normalizedValue)
+    }
+  }
+
+  def setResponseFormat(value: OpenAIResponseFormat.ResponseFormat): this.type = {
+    this.setResponseFormat(value.name)
+  }
```
mhamilton723, line 142: this code seems duplicated, can it be abstracted and shared?

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletionSuite.scala
```diff
@@ -174,6 +176,45 @@ class OpenAIChatCompletionSuite extends TransformerFuzzing[OpenAIChatCompletion]
...
+    completion.setResponseFormat(OpenAIResponseFormat.JSON)
+    completion.getResponseFormat shouldEqual Map("type" -> "json_object")
+  }
+
+  test("setResponseFormatCol should set the response format column correctly") {
+    completion.setResponseFormatCol("response_format_col")
+    completion.getResponseFormatCol shouldEqual "response_format_col"
+  }
+
   def testCompletion(completion: OpenAIChatCompletion, df: DataFrame, requiredLength: Int = 10): Unit = {
```
mhamilton723, line 218: can you add a test that calls the service with this new API param set just as a lil e2e

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPromptSuite.scala
```diff
@@ -149,6 +257,100 @@ class OpenAIPromptSuite extends TransformerFuzzing[OpenAIPrompt] with OpenAIAPIK
...
+
+    prompt.setResponseFormat("json_object")
+    prompt.getResponseFormat should be ("json_object")
+
+    prompt.setResponseFormat("jSoN")
+    prompt.getResponseFormat should be ("json_object")
+
+    prompt.setResponseFormat("TEXT")
+    prompt.getResponseFormat should be ("text")
+  }
```
mhamilton723, line 311: these tests look duplicated from above, if the core parameter flexibility is shared between the two we can keep just 1 copy and still have same basic coverage

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
FarrukhMasud: I am now using shared code.

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPromptSuite.scala
FarrukhMasud: removed

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
FarrukhMasud: made mappedMessages val again

PR #2319, opened 2024-11-22
fix: Fixing OpenAIPrompt class to be able to work with older models like text-davinci-003

FarrukhMasud:
## Issue
While working with `OpenAIPrompt` class, I discovered that this class does not work with older models. Upon investigation, I found that for models specified in the list `legacyModels`, OpenAIPrompt class suppose to use `OpenAICompletion` instance, for all other model deployment names, `OpenAIChatCompletion` class is to be used. When setting parameter for `OpenAICompletion`, it tries to set parameters that are not available in `OpenAICompletion`, which causes instance creation to fail. There are no current unit tests or integration tests that verified this. In this PR, I am fixing this problem and adding unit test to validate this. I am also cleaning unit test and OpenAIPrompt class.

In this PR I am:
1. Ensuring that new prompt instances are created for each test. This will eliminate chance of one test interfering with another test.
2. Creating test to validate that for older model deployments, OpenAiCompletion instance is created in OpenAIPrompt class.
3. Fixing the bug that caused creation of OpenAICompletion object to fail.
4. Cleaning some code in OpenAIPrompt class.

## How is this patch tested?

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAI.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAI.scala
@@ -8,6 +8,7 @@ import com.microsoft.azure.synapse.ml.fabric.{FabricClient, OpenAIFabricSetting}
 import com.microsoft.azure.synapse.ml.logging.common.PlatformDetails
 import com.microsoft.azure.synapse.ml.param.ServiceParam
 import com.microsoft.azure.synapse.ml.services._
+import org.apache.http.entity.AbstractHttpEntity
 import org.apache.spark.ml.PipelineModel
 import org.apache.spark.ml.param.{Param, Params}
 import org.apache.spark.sql.Row
@@ -294,7 +295,7 @@ trait HasOpenAITextParams extends HasOpenAISharedParams {
 }
 
 abstract class OpenAIServicesBase(override val uid: String) extends CognitiveServicesBase(uid: String)
-  with HasOpenAISharedParams with OpenAIFabricSetting {
+  with HasOpenAISharedParams with OpenAIFabricSetting with HasCognitiveServiceInput {
   setDefault(timeout -> 360.0)
 
   private def usingDefaultOpenAIEndpoint(): Boolean = {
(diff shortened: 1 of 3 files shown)


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
```diff
@@ -106,9 +106,7 @@ class OpenAIPrompt(override val uid: String) extends Transformer
     df.map({ row =>
       val originalOutput = Option(row.getAs[Row](outputCol))
         .map({ row => openAIResultFromRow(row).choices.head })
-      val isFiltered = originalOutput
```
FarrukhMasud, line 109: This change is just to simplify the code. It has no functional impact.

cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPromptSuite.scala
```diff
@@ -21,12 +21,42 @@ class OpenAIPromptSuite extends TransformerFuzzing[OpenAIPrompt] with OpenAIAPIK
...
+    new OpenAIPrompt()
+      .setSubscriptionKey(openAIAPIKey)
+      .setDeploymentName(deploymentName)
+      .setCustomServiceName(openAIServiceName)
+      .setOutputCol("outParsed")
+      .setTemperature(0)
+  }
+
+
+  test("Validating OpenAICompletion and OpenAIChatCompletion for correctModels") {
```
mhamilton723, line 34: gorgeous!

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAI.scala
```diff
@@ -307,4 +308,6 @@ abstract class OpenAIServicesBase(override val uid: String) extends CognitiveSer
     }
     super.getInternalTransformer(schema)
   }
+
+  private[openai] def prepareEntityAIService: Row => Option[AbstractHttpEntity] = this.prepareEntity
```
mhamilton723, line 312: is this function necessary? seems like its not but hard to tell

FarrukhMasud: this si required. I will add comemtns to this method

mhamilton723 (approved):

PR #2320, opened 2024-11-25
fix: OpenAIChatCompletion class method shouldSkip method to skip row if no user message

FarrukhMasud:
`OpenAIChatCompletion` class method `shouldSkip` was not working as designed as there might OpenAIMessages with empty contents. These rows can be identified before sending requests to OpenAI service and getting errors. 

In this PR, I am making changes to skip rows if there are no user messages in the message column.

## How is this patch tested?

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
@@ -48,14 +48,16 @@ class OpenAIChatCompletion(override val uid: String) extends OpenAIServicesBase(
 
   /**
    * Check if the row should be skipped. A row should be skipped if the messages column is empty or if any of the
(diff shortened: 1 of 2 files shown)


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIChatCompletion.scala
```diff
@@ -46,8 +46,22 @@ class OpenAIChatCompletion(override val uid: String) extends OpenAIServicesBase(
 
   override val subscriptionKeyHeaderName: String = "api-key"
 
-  override def shouldSkip(row: Row): Boolean =
-    super.shouldSkip(row) || Option(row.getAs[Row](getMessagesCol)).isEmpty
+  /**
```
mhamilton723, line 49: This seems like something we should likely have in the prompt API as opposed to the core API just because im not sure if its possible to send just system messages and have it reply

PR #2334, opened 2025-01-06
chore: OpenAIPrompt bug fixes

sss04:
## Related Issues/PRs

A couple of bug fixes:
- move error column to back
- rename error column 
- fix issue where params that don't exist were being checked

## How is this patch tested?

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
@@ -15,7 +15,7 @@ import org.apache.spark.ml.param.{BooleanParam, Param, ParamMap, ParamValidators
 import org.apache.spark.ml.util.Identifiable
 import org.apache.spark.sql.{Column, DataFrame, Dataset, Row, functions => F, types => T}
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
-import org.apache.spark.sql.functions.udf
+import org.apache.spark.sql.functions.{col, udf}
 import org.apache.spark.sql.types.{DataType, StructType}
 
 import scala.collection.JavaConverters._
@@ -111,7 +111,7 @@ class OpenAIPrompt(override val uid: String) extends Transformer
     postProcessing -> "",
     postProcessingOptions -> Map.empty,
     outputCol -> (this.uid + "_output"),
-    errorCol -> (this.uid + "_error"),
+    errorCol -> "gen_error",
     messagesCol -> (this.uid + "_messages"),
     dropPrompt -> true,
     systemPrompt -> defaultSystemPrompt,
@@ -152,7 +152,9 @@ class OpenAIPrompt(override val uid: String) extends Transformer
       val df = dataset.toDF
       val completion = openAICompletion
       val promptCol = Functions.template(getPromptTemplate)
-
+      val newErrorCol = df.withDerivativeCol(getErrorCol)
+      setErrorCol(newErrorCol)
+      completion.setErrorCol(newErrorCol)
       completion match {
         case chatCompletion: OpenAIChatCompletion =>
           if (isSet(responseFormat)) {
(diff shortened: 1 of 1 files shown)


cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
```diff
@@ -111,7 +111,7 @@ class OpenAIPrompt(override val uid: String) extends Transformer
     postProcessing -> "",
     postProcessingOptions -> Map.empty,
     outputCol -> (this.uid + "_output"),
-    errorCol -> (this.uid + "_error"),
+    errorCol -> "gen_error",
```
mhamilton723, line 114: we need to properly disambiguate this so it doesent clobber results

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
```diff
@@ -152,7 +152,9 @@ class OpenAIPrompt(override val uid: String) extends Transformer
       val df = dataset.toDF
       val completion = openAICompletion
       val promptCol = Functions.template(getPromptTemplate)
-
+      val newErrorCol = df.withDerivativeCol(getErrorCol)
+      setErrorCol(newErrorCol)
+      completion.setErrorCol(newErrorCol)
```
mhamilton723, line 157: so in the case where a user sets this it likely should clobber the results, if a user doesent set this it should default to something reasonmable that doesent clobber

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
```diff
@@ -172,32 +174,30 @@ class OpenAIPrompt(override val uid: String) extends Transformer
...
           val completionNamed = completion.setPromptCol(promptColName)
-          // run completion
           val results = completionNamed
             .transform(dfTemplated)
             .withColumn(getOutputCol,
               getParser.parse(F.element_at(F.col(completionNamed.getOutputCol).getField("choices"), 1)
                 .getField("text")))
             .drop(completionNamed.getOutputCol)
-
+          val resultsFinal = results.select(results.columns.filter(_ != getErrorCol).map(col) :+ col(getErrorCol): _*)
```
mhamilton723, line 196: is this necessary, do error colum,ns show up before other columns?

sss04: Right now the Error column appears before the output column - we wanted it at the end so I added this

mhamilton723: is this change reflected in transform schema?

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
sss04: Removed this as per our offline discussion

cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/openai/OpenAIPrompt.scala
sss04: After trying modifying SimpleHTTPTransformer, I think the `withColumn(getOutputCol` on line 192 is moving the output column to the back

sss04: did some surface level research and it seems if we're rearranging columns, transformSchema doesn't need a change? if that's true, we're good as is

mhamilton723: it unfortunately will as it is ordered under the hood, you can ovverride this func too to do the switch

sss04: changed `transformSchema` to move the Error column to the back

mhamilton723 (approved):

PR #2345, opened 2025-02-04
chore: For python tests, we are increasing XMX

FarrukhMasud:
## Related Issues/PRs

We have seen that python tests for deep learning because smaller heap size. In this PR I am increasing this XMX to 4gb and essentially using same vaklues as we use for R tests 

## How is this patch tested?
This is just a change in pipeline and it will be tested by the pipeline

### DIFF
--- a/pipeline.yaml
+++ b/pipeline.yaml
@@ -401,7 +401,7 @@ jobs:
         inlineScript: |
           set -e
           source activate synapseml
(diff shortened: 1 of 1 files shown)


FarrukhMasud:
not neerded

FarrukhMasud:
Not needed

Issue #2293, opened 2024-10-06
[BUG] Cannot Load LightGBM Model When Placed in a Spark Pipeline with Custom Transformers


dsmith111:
### SynapseML version

1.0.5

### System information

- **Language version**: Python 3.11.0rc1 and Scala version 2.12.15
- **Spark Version**: 3.5.0
- **Spark Platform**: Databricks

### Describe the problem

When attempting to create a Spark pipeline that contains _both_ a custom transformer and a LightGBM model, the loading of the pipeline fails with "AttributeError: module 'com.microsoft.azure.synapse.ml.lightgbm' has no attribute". Note: The saving/loading works fine if either the custom transformer is not present or the lightgbm model is not present, it is the combination of both that leads to the issue.

Related to issues #614 and #1701.

### Code to reproduce issue

**Example Custom Transformers**
```python
from pyspark.ml import Transformer
from pyspark.ml.util import MLReadable, MLWritable, DefaultParamsReadable, DefaultParamsWritable
from pyspark.ml.param.shared import HasInputCol, HasOutputCol
from pyspark.ml.param import Param, Params
import pyspark.sql.functions as f 
from pyspark.ml.linalg import VectorUDT, Vectors
from pyspark.sql.types import *
import json

class ColumnSelector(Transformer, DefaultParamsReadable, DefaultParamsWritable):
    """
    Custom Transformer to select and rename columns from a DataFrame.
    Enhanced to be MLWritable.
    """
    
    def __init__(self, selectExpr=None):
        super(ColumnSelector, self).__init__()
        self.selectExpr = Param(self, "selectExpr", "The SQL expression used for selecting and renaming columns")
        self._setDefault(selectExpr=selectExpr)
        if selectExpr is not None:
            self.setSelectExpr(selectExpr)

    def setSelectExpr(self, value):
        """
        Sets the SQL expression for selecting and renaming columns.
        """
        return self._set(selectExpr=value)

    def getSelectExpr(self):
        """
        Gets the current SQL expression for selecting and renaming columns.
        """
        return self.getOrDefault(self.selectExpr)

    def _transform(self, dataset):
        """
        The method that defines the operations to produce the selected and renamed columns.
        """
        return dataset.selectExpr(*self.getSelectExpr())


class StringArrayToVectorTransformer(Transformer, DefaultParamsReadable, DefaultParamsWritable):
    """
    Custom Transformer that converts a string representation of an integer array to a VectorUDT.
[...]

Issue #1701, opened 2022-10-31
[BUG] 'com.microsoft.azure.synapse.ml.lightgbm' has no attribute 'LightGBMClassificationModel'


sibyl1956:
### SynapseML version

0.10.1

### System information

Language version: Python: 3.8.10, Scala 2.12
Spark Version : Apache Spark 3.2.1,
Spark Platform: Databricks

### Describe the problem

When try to load a pipeline model for lightgbm, I encountered this error message: 
'com.microsoft.azure.synapse.ml.lightgbm' has no attribute 'LightGBMClassificationModel'

But I imported from synapse.ml.lightgbm import LightGBMClassificationModel before I try to load the pipeline model

### Code to reproduce issue
```
from pyspark.ml.pipeline import PipelineModel
from synapse.ml.lightgbm import LightGBMClassificationModel, LightGBMClassifier
clf = PipelineModel.load(model_savepath)
```
### Other info / logs
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<command-2087039020756525> in <module>
      1 # Load model
      2 from pyspark.ml.pipeline import PipelineModel
----> 3 clf = PipelineModel.load(model_savepath)

/databricks/spark/python/pyspark/ml/util.py in load(cls, path)
    461     def load(cls, path):
    462         """Reads an ML instance from the input path, a shortcut of `read().load(path)`."""
--> 463         return cls.read().load(path)
    464 
    465 

/databricks/spark/python/pyspark/ml/pipeline.py in load(self, path)
    258             return JavaMLReader(self.cls).load(path)
    259         else:
--> 260             uid, stages = PipelineSharedReadWrite.load(metadata, self.sc, path)
    261             return PipelineModel(stages=stages)._resetUid(uid)
    262 

/databricks/spark/python/pyspark/ml/pipeline.py in load(metadata, sc, path)
    394             stagePath = \
    395                 PipelineSharedReadWrite.getStagePath(stageUid, index, len(stageUids), stagesDir)
--> 396             stage = DefaultParamsReader.loadParamsInstance(stagePath, sc)
    397             stages.append(stage)
    398         return (metadata['uid'], stages)

/databricks/spark/python/pyspark/ml/util.py in loadParamsInstance(path, sc)
    719         else:
    720             pythonClassName = metadata['class'].replace("org.apache.spark", "pyspark")
--> 721         py_type = DefaultParamsReader.__get_class(pythonClassName)
[...]

Issue #614, opened 2019-07-11
Cannot load PipelineModel


Keyeoh:
Hi,

I am trying to port my ML pipeline so I can use LightGBM instead of the PySpark GBT. I have been able to design a Pipeline with a LightGBM as final estimator. Once trained, I save the PipelineModel object to disk succesfully. 

Problem is, when I want to load the model again to evaluate it, the following error appears:

```
2019-07-11 10:44:03 INFO  DAGScheduler:54 - Job 66 finished: runJob at PythonRDD
.scala:152, took 0,709961 s
Traceback (most recent call last):
  File "C:/Users/Y0644483/Documents/Workspace/ninabrlong/bin/eval_model.py", lin
e 86, in <module>
    model = ml.PipelineModel.load(args["<path_model>"])
  File "C:\Users\Y0644483\AppData\Local\Continuum\miniconda3\envs\ninabrlong\lib
\site-packages\pyspark\python\lib\pyspark.zip\pyspark\ml\util.py", line 311, in
load
  File "C:\Users\Y0644483\AppData\Local\Continuum\miniconda3\envs\ninabrlong\lib
\site-packages\pyspark\python\lib\pyspark.zip\pyspark\ml\pipeline.py", line 244,
 in load
  File "C:\Users\Y0644483\AppData\Local\Continuum\miniconda3\envs\ninabrlong\lib
\site-packages\pyspark\python\lib\pyspark.zip\pyspark\ml\pipeline.py", line 378,
 in load
  File "C:\Users\Y0644483\AppData\Local\Continuum\miniconda3\envs\ninabrlong\lib
\site-packages\pyspark\python\lib\pyspark.zip\pyspark\ml\util.py", line 535, in
loadParamsInstance
  File "C:\Users\Y0644483\AppData\Local\Continuum\miniconda3\envs\ninabrlong\lib
\site-packages\pyspark\python\lib\pyspark.zip\pyspark\ml\util.py", line 478, in
__get_class
AttributeError: module 'com.microsoft.ml.spark' has no attribute 'LightGBMRegres
sionModel'
```

I could not find any reference to this error, and I do not have a clue on what it could be happening. Besides, I found some references in your docs about using saveNativeModel(), but do not know how that fits in a whole-pipeline-saving scenario.

I am using mmlspark 0.17 and pyspark 2.3.2 in standalone mode in my local development environment. 

I looked into the saved model file and found the following structure:

```
[...]

PR #2357, opened 2025-03-23
fix: Cannot Load LightGBM Model When Placed in a Spark Pipeline with Custom Transformers

dsmith111:
## Related Issues/PRs

Directly related issues: #2293, #1701
Issue that _looks_ to be related, however, used an older API: #614

### Problem

When attempting to load a Spark pipeline that includes both a non-java backed, "pure Python" transformer and a LightGBM model, an AttributeError was encountered, indicating that the module com.microsoft.azure.synapse.ml.lightgbm lacked the LightGBMClassificationModel attribute. The existing java_params_patch was not sufficient to resolve this, as it only applied to the JavaParams._from_java method and did not cover scenarios where DefaultParamsReader.loadParamsInstance was used.

The previous patch for JavaParams._from_java handled the conversion of Java objects to Python objects during deserialization but did not account for the direct instantiation of classes based on metadata in DefaultParamsReader.loadParamsInstance (as is done during Pipeline/PipelineModel stage loading). As a result, the class references within the ComplexParamsMixin superclass were not correctly resolved, leading to the AttributeError in the above issues.

### Solution

In similar fashion to the existing patch, this solution adds a loadParamsInstance patch which enhances the existing Python class handling condition to correct the SynapseML class name in addition to its existing pyspark class name correction.

This PR ensures that the necessary class name transformations are correctly handled during the loading process, addressing the limitation of the previous patch.

## How is this patch tested?

- [x] I have written tests (not required for typo or doc fix) and confirmed the proposed bug-fix works.
- [x] I have trained, saved and loaded these PipelineModels with and without the patch to confirm it works.


dsmith111:
@mhamilton723, @svotaw, tagging maintainers that have both interacted with this problem in the past.

lightgbm/src/test/python/synapseml/lightgbm/test_lightgbm.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def setUp(self):
+        # Define a temporary directory for saving models/pipelines.
+        self.temp_dir = "./tmp/lightgbm_serialization_test"
+        if os.path.exists(self.temp_dir):
+            shutil.rmtree(self.temp_dir)
+
+    def tearDown(self):
+        if os.path.exists(self.temp_dir):
+            shutil.rmtree(self.temp_dir)
```
mhamilton723, line 63: Our tests manage the spark and spark context objects on behalf of the user, im not sure youll need this if you copy the structure of other tests in the repo

lightgbm/src/test/python/synapseml/lightgbm/test_lightgbm.py
```diff
@@ -0,0 +1,122 @@
...
+from pyspark.sql import SparkSession, Row
+from pyspark.ml import PipelineModel, Transformer
+from pyspark.ml.util import DefaultParamsReadable, DefaultParamsWritable
+from pyspark.ml.linalg import Vectors, VectorUDT
+from pyspark.sql.types import ArrayType, IntegerType
+import pyspark.sql.functions as f
+from synapse.ml.stages import SelectColumns
+import synapse.ml.lightgbm as lgbm
+
+class StringArrayToVectorTransformer(Transformer, DefaultParamsReadable, DefaultParamsWritable):
```
mhamilton723, line 14: is this a throwaway piece of code to just test serialization of a custom model?

lightgbm/src/test/python/synapseml/lightgbm/test_lightgbm.py
dsmith111: Ah, I see, reused their components. Removed handling functions.

lightgbm/src/test/python/synapseml/lightgbm/test_lightgbm.py
dsmith111: I was not exactly sure on the interpretation of this but stripped a fair amount of logic from the test and changed the function name/description to reflect better what it is for. Based on the update, let me know if you think if it should be further reduced. I have the single vs pipeline serialization test to demonstrate the standard single-model Python serialization/deserialization was not affected along with the pipeline test demonstrating the mentioned issue is resolved.

PR #2435, opened 2025-10-14
chore: Bump version 1.0.16

ranadeepsingh:

Bump up SynpaseML to v1.0.16

Adding:
- working scripts and readme to auto bump or revert changes in future to make bumping version easier

### DIFF
--- a/cognitive/src/main/python/synapse/ml/services/openai/OpenAIDefaults.py
+++ b/cognitive/src/main/python/synapse/ml/services/openai/OpenAIDefaults.py
@@ -102,10 +102,16 @@ class OpenAIDefaults:
     def reset_model(self):
         self.defaults.resetModel()
 
-    # New verbosity parameter
+    def set_embedding_deployment_name(self, name):
+        self.defaults.setEmbeddingDeploymentName(name)
+
+    def get_embedding_deployment_name(self):
+        return getOption(self.defaults.getEmbeddingDeploymentName())
+
+    def reset_embedding_deployment_name(self):
+        self.defaults.resetEmbeddingDeploymentName()
+
     def set_verbosity(self, verbosity):
-        # Accept any string the user provides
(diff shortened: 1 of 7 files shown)


tools/release/bump_version.sh
```diff
@@ -0,0 +1,110 @@
...
+  # Optional: verify contains TO
+  rg -n -S "$TO" "$f" | head -n 2 || true
+done <<< "$FILES"
+
+if [[ $EXIT -ne 0 ]]; then
+  echo "One or more files failed to update" >&2
+  exit $EXIT
+fi
+
+echo "Version bump complete: $FROM -> $TO"
```
mhamilton723: wow amazing, is there a way to make this python for future maintainability?

tools/release/revert_bump.sh
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env bash
```
mhamilton723: What is point of this file? seems like same logic as bump version but with the strings going other way

tools/release/version_docs.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
mhamilton723: THis is done by docusaurus version command. Not sure what this script does

tools/release/revert_bump.sh
ranadeepsingh: It is overkill, we could always just discard the changes make by `bump_version.sh`. My idea with this was to do an easy "undo". Can delete this one

tools/release/bump_version.sh
ranadeepsingh: Done - made python version of this

tools/release/version_docs.sh
ranadeepsingh: Move it to a release.py file with docusaurus command

PR #2462, opened 2026-01-10
chore: remove redundant sas token from public test file

BrendanWalsh:
These SAS tokens are redundant as the underlying file is already public. They are also inconvenient because static analysis tools flag them as a hazard.


levscaut (approved):

TomFinley (approved):
Thank you @BrendanWalsh .

PR #2533, opened 2026-03-31
ci: pin remaining GitHub Actions and all Docker base images to immutable references

BrendanWalsh:
## Why

Floating references are mutable. `@master`, `@v2.x`, and a bare image tag can all be repointed at new code by whoever controls the upstream — the classic tag-hijack supply-chain attack. Immutable references (commit SHAs, image digests) remove that class of risk and make builds reproducible.

This also closes the OpenSSF Scorecard **Pinned-Dependencies** check.

Most of the repo was already pinned by earlier work. This PR closes the **last three unpinned/imprecise references** and pins the Docker base images.

### GitHub Actions — 3 workflows

| Workflow | Was | Now | Note |
|---|---|---|---|
| `ado-pr-to-workitem.yml` | `danhellem/…@master` | `@496254e4…` `# master` | floating branch → frozen |
| `remove-awaiting-response-label.yml` | `octokit/request-action@v2.x` | `@02f5e7c6…` `# v2.x` | floating branch → frozen |
| `scorecards.yml` | `codeql-action/upload-sarif@d1ba80a1…` `# v4` | `@5595ccaf…` `# v4.37.6` | already a SHA; moved to the exact release already used by `codeql.yml`, and relabelled with the precise version |

The first two upstreams publish no semver tags, so the pin tracks a branch head frozen at a known-good commit.

### Docker base images — 8 Dockerfiles

Tags are kept alongside the digest (`image:tag@sha256:…`) so the intended version stays readable.

| Image | Digest | Files |
|---|---|---|
| `mirror/docker/library/ubuntu:22.04` | `sha256:104ae837…d813da2fb` | `tools/docker/{demo,minimal}` |
| `openjdk/jdk:11-mariner` | `sha256:eea2eae2…d4bfc7c7` | `tools/helm/{livy,spark,spark/mini,zeppelin}` |
| `mmlspark/spark2.4:v4_mini` | `sha256:a7da0d7c…1a389279` | `tools/helm/{livy/mini,zeppelin/mini}` |

The ubuntu and openjdk pins target the multi-arch OCI **index** digest, which is the correct thing for `FROM`.

## Verification

Every reference was resolved against the live GitHub and MCR registry APIs — 11 of 11 confirmed to match the version in its trailing comment:

- Branch-head pins verified as the current HEAD of `master` / `v2.x`.
- `5595ccaf…` verified by dereferencing annotated tag `v4.37.6`; it is the direct child of the previous `d1ba80a1…`, i.e. a one-release forward move inside major v4 — not a downgrade or cross-major jump.
- Each digest checked both by-digest (exists) and by-tag (tag still resolves to it).

## Impact

**No functional change** — every pin resolves to the same code and images CI was already running.

Trade-off worth naming: the two branch-head pins no longer receive upstream fixes automatically and must be refreshed by hand or by Dependabot. That is the intended effect of pinning, not a defect.

## Checklist

- [x] Does not change dependencies (no `build.sbt` / manifest change)
- [x] Not a new feature — no website samples needed
- [x] No user-facing behavior change

### DIFF
--- a/.github/workflows/ado-pr-to-workitem.yml
+++ b/.github/workflows/ado-pr-to-workitem.yml
@@ -10,7 +10,7 @@ jobs:
   alert:
     runs-on: ubuntu-latest
     steps:
-    - uses: danhellem/github-actions-pr-to-work-item@master
+    - uses: danhellem/github-actions-pr-to-work-item@496254e48adbe7f1ed14a8afb71dc520b2c052ac # master
       env:
         ado_token: '${{ secrets.ADO_PERSONAL_ACCESS_TOKEN }}'
         github_token: '${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}'
--- a/.github/workflows/remove-awaiting-response-label.yml
+++ b/.github/workflows/remove-awaiting-response-label.yml
@@ -13,7 +13,7 @@ jobs:
       github.event.comment.author_association != 'COLLABORATOR'
     steps:
       - name: Remove needs-reply label
-        uses: octokit/request-action@v2.x
+        uses: octokit/request-action@02f5e7c637a73a3b12ed81015fa7fb5f11cc5d7d # v2.x
         continue-on-error: true
         with:
           route: DELETE /repos/:repository/issues/:issue/labels/:label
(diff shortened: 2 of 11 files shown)


ranadeepsingh (approved):

kiratp (approved):

ranadeepsingh:
Thanks for driving this one through, @BrendanWalsh! Pinning every action and base image to an immutable digest closes a real supply-chain gap, and the eight Dockerfiles were the part that's easy to forget. Appreciate the follow-through. :rocket:

PR #2560, opened 2026-05-08
feat: Add v1 OpenAI endpoint support and remove legacy completions API

ranadeepsingh:
## Related Issues/PRs

#xxx

## What changes are proposed in this pull request?

This PR updates SynapseML OpenAI integrations to support OpenAI v1-compatible base URLs such as:

`https://<resource_name>.services.ai.azure.com/openai/v1`

Changes include:

- Adds v1 base URL detection and request URL construction for OpenAI chat completions, embeddings, responses, and `OpenAIPrompt`.
- For v1 endpoints, sends the deployment/model as the request body `model` field and omits the `api-version` query parameter.
- Preserves existing Azure OpenAI endpoint behavior for non-v1 URLs, including deployment-scoped paths and `api-version`.
- Emits a warning when users configure a v1 base URL together with `apiVersion`, since `apiVersion` is ignored for v1-style endpoints.
- Removes the deprecated/retired legacy `OpenAICompletion` API implementation, response schema, and tests.
- Updates current OpenAI docs/notebooks and docgen metadata to remove legacy completion usage and point users to `OpenAIChatCompletion`, `OpenAIPrompt`, and `OpenAIResponses`.

## How is this patch tested?

- [x] I have written tests (not required for typo or doc fix) and confirmed the proposed feature/bug-fix/change works.

Tested with the `synapseml` conda environment:

- `conda run -n synapseml black --check --extend-exclude 'docs/' .`
- `conda run -n synapseml sbt "cognitive / Test / compile"`
- `conda run -n synapseml sbt "cognitive / Test / testOnly com.microsoft.azure.synapse.ml.services.openai.OpenAIV1EndpointSuite"`
- `conda run -n synapseml sbt scalastyle test:scalastyle`
- `git diff --check`

Note: `sbt scalafmtCheckAll` was attempted but this build does not define that sbt task.

## Does this PR change any dependencies?

- [x] No. You can skip this section.
- [ ] Yes. Make sure the dependencies are resolved correctly, and list changes here.

## Does this PR add a new feature? If so, have you added samples on website?

- [ ] No. You can skip this section.
- [x] Yes. Make sure you have added samples following below steps.

This adds support for OpenAI v1-compatible base URLs. Current OpenAI notebooks/docs were updated to remove retired `OpenAICompletion` examples and describe supported chat, responses, and embedding APIs.

PR #2590, opened 2026-08-01
fix: preserve hand-written Python package initializers

ranadeepsingh:
## Related PRs

- Recreates the still-relevant init-file preservation intent from #2286 on current `master`.
- Preserves the OpenAI compatibility hook introduced by #2560.

## Problem

PyCodegen copies hand-written Python overrides and then rewrites nested `__init__.py` files, discarding required manual content. It also deletes every namespace-root initializer and the prepend approach proposed in #2286 would duplicate generated headers when init generation is repeated.

Before the fix, targeted regression tests reproduced three failures: nested manual UTF-8 content was missing, a non-empty manual namespace root was deleted, and OpenAI manual content was lost around the #2560 hook.

## Changes

- Read manual initializer content from the source override tree, never from generated output.
- Emit generated imports in deterministic filename order using explicit UTF-8.
- Compose generated package metadata/imports, `initFileExtra`, then manual content.
- Keep empty/absent namespace roots as namespace packages while retaining non-empty manual roots.
- Keep the cognitive compatibility initializer entirely hand-written.
- Add focused tests for roots, nested packages, UTF-8, manual presence/absence, OpenAI extras, and repeat-run idempotency.

No files under `target/` or other generated output are committed.

## Validation

- `core/testOnly com.microsoft.azure.synapse.ml.codegen.PyCodegenSuite`: 11 passed.
- `core/Test/compile` and `cognitive/Test/compile`: passed.
- Canonical `core/codegen`, `cognitive/codegen`, and `deepLearning/codegen`: passed.
- `core/scalastyle` and `core/Test/scalastyle`: passed with zero findings.
- `black==22.3.0 --check --extend-exclude docs/ .`: 179 files unchanged.
- Strict UTF-8 generated-output assertions passed for namespace roots, manual suffixes, OpenAI hook/import ordering, and the cognitive compatibility package.
- Repeated fixed-version `core/codegen` was byte-idempotent across all 30 generated `__init__.py` files.

## Review

The repository `code-review` checklist was run before commit; no concrete findings remained.
## CI iteration

The first Azure Pipelines run exposed a stale source override at `core/src/main/python/synapse/ml/automl/__init__.py`: it imported `_FindBestModel`, but that generated module does not exist. The file had previously been harmless only because PyCodegen overwrote it. The override is now removed so deterministic codegen supplies the valid `FindBestModel` import. Local targeted tests and canonical core codegen passed after the fix.
## Packaging review follow-up

Independent review found that `find_namespace_packages(include=["synapse.ml.*"])` omitted a preserved `synapse/ml/__init__.py` from wheels. Component and aggregate setup templates now include both the exact `synapse.ml` package and `synapse.ml.*`. Regression tests build and inspect real fixture wheels for both templates, including UTF-8 root initializer content.
## Codegen shard follow-up

Enabling `com.microsoft.azure.synapse.ml.codegen.**` exposed a dormant `WrappableTests` assertion that still expected `TestRegressorModel` to be nested under `WrappableTests`, even though #2195 moved the fixture to a package-level source file. The assertion now verifies the actual reflected top-level name. The correction is intentionally limited to that one line so the equivalent Spark 4.1 change replays cleanly. The combined `WrappableTests` and `PyCodegenSuite` run passes 12/12 tests.

The actual PR patch now applies cleanly onto spark4.1, including the codegen production and test sources.

### DIFF
--- a/core/src/main/python/synapse/ml/automl/__init__.py
+++ b/core/src/main/python/synapse/ml/automl/__init__.py
@@ -1,1 +1,0 @@
-from _FindBestModel import _FindBestModel
--- a/build.sbt
+++ b/build.sbt
@@ -173,7 +173,7 @@ packageSynapseML := {
          |    long_description="SynapseML contains Microsoft's open source "
          |                     + "contributions to the Apache Spark ecosystem",
          |    license="MIT",
-         |    packages=find_namespace_packages(include=['synapse.ml.*']),
+         |    packages=find_namespace_packages(include=['synapse.ml', 'synapse.ml.*']),
          |    url="https://github.com/Microsoft/SynapseML",
          |    author="Microsoft",
          |    author_email="synapseml-support@microsoft.com",
--- a/core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/PyCodegen.scala
+++ b/core/src/main/scala/com/microsoft/azure/synapse/ml/codegen/PyCodegen.scala
@@ -173,7 +173,7 @@ object PyCodegen {
          |    long_description="SynapseML contains Microsoft's open source "
          |                     + "contributions to the Apache Spark ecosystem",
          |    license="MIT",
-         |    packages=find_namespace_packages(include=['synapse.ml.*']) ${extraPackage},
+         |    packages=find_namespace_packages(include=['synapse.ml', 'synapse.ml.*']) ${extraPackage},
          |    url="https://github.com/Microsoft/SynapseML",
          |    author="Microsoft",
          |    author_email="synapseml-support@microsoft.com",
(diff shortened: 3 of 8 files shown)


ranadeepsingh:
@imatiach-msft Re-review requested after addressing the namespace-root wheel packaging finding.

Commit `484c24f5ea` now includes both `synapse.ml` and `synapse.ml.*` in component and aggregate discovery. The new regression coverage builds and inspects both wheel variants and verifies the preserved UTF-8 `synapse/ml/__init__.py` content.

Local validation: 7/7 targeted PyCodegen tests passed, component/test scalastyle passed with zero findings, Black passed, and canonical core codegen emitted the corrected discovery expression.

core/src/main/python/synapse/ml/automl/__init__.py
```diff
@@ -1 +0,0 @@
-from _FindBestModel import _FindBestModel
```
ranadeepsingh, line 1: @copilot why is this import not required anymore?

BrendanWalsh (approved):

PR #2609, opened 2026-08-06
feat: add Spark-native univariate time series anomaly detection

ranadeepsingh:
## Why

[#2605](https://github.com/microsoft/SynapseML/pull/2605) removed the Azure AI Anomaly Detector wrappers because the service was retired — every endpoint returns `HTTP 410 Gone`. That fixed the broken CI job, but it left SynapseML users with **no time series anomaly detection path at all**. `IsolationForest` covers *point* anomalies in tabular data, but it has no notion of time, seasonality, or trend, so it is not a substitute.

This PR restores the capability on top of the open source successor, [`microsoft/anomaly-detector`](https://github.com/microsoft/anomaly-detector) (PyPI: `time-series-anomaly-detector`), and — importantly — **distributes it across the cluster** instead of leaving users to loop over series on the driver.

## What this adds

| File | Purpose |
| --- | --- |
| `core/src/main/python/synapse/ml/timeseries/UnivariateAnomalyDetector.py` | The Spark `Transformer` |
| `core/src/main/python/synapse/ml/timeseries/__init__.py` | Package export |
| `core/src/test/python/synapsemltest/timeseries/test_univariate_anomaly_detector.py` | 15 tests |
| `docs/Explore Algorithms/Anomaly Detection/Time Series Anomaly Detection.md` | Docs incl. migration guide |
| `cognitive/.../anomaly.py` | Migration shim now points somewhere useful |
| `website/sidebars.js` | Registers the new page |

## Performance: this is the main point

Each series is detected on the executor that already holds it, using `groupBy(groupByCols).applyInPandas(...)`. Throughput scales with the number of series — no driver bottleneck, no collect.

```python
from synapse.ml.timeseries import UnivariateAnomalyDetector

detector = (UnivariateAnomalyDetector()
    .setTimestampCol("timestamp")
    .setValueCol("value")
    .setGroupByCols(["device_id"])   # <- thousands of series, detected in parallel
    .setGranularity("hourly")
    .setOutputCol("anomalies"))

detector.transform(df).select("device_id", "timestamp", "anomalies.isAnomaly")
```

Verified on 12 series × 200 points across 4 partitions: all 12 planted spikes detected, one Spark job, no driver-side iteration.

Rows are sorted by the timestamp column *within* each group, because the library assumes chronological order and `applyInPandas` does not guarantee it.

## Drop-in migration

All **seven** inputs of the retired transformer are preserved verbatim: `granularity`, `customInterval`, `period`, `maxAnomalyRatio`, `sensitivity`, `imputeMode`, `imputeFixedValue`. Granularity values are identical (`yearly` … `microsecond`, `none`).

The OSS library returns snake_case fields; I map them back to the **legacy camelCase output contract** (`isAnomaly`, `isPositiveAnomaly`, `isNegativeAnomaly`, `expectedValue`, `upperMargin`, `lowerMargin`, `period`), so existing code like `col("anomalies.isAnomaly")` keeps working.

> ⚠️ **`severity` is gone.** The OSS library does not expose it in either detection mode. Documented rather than faked.

## Why the dependency is optional (not in `environment.yml`)

`time-series-anomaly-detector` requires `mlflow>=3.1.0` (it pulls **3.15.1**). This repo pins:

```yaml
# Petastorm needs legacy Parquet and fsspec APIs removed after PyArrow 10; MLflow accepts this pin.
- mlflow==2.21.3
```

Adding the package would force a conflicting MLflow upgrade and risk breaking Petastorm. So the import is **lazy**, with an actionable install message, and the two end-to-end tests use `pytest.importorskip` and skip cleanly in CI. Happy to revisit if maintainers would rather take the upgrade.

## Three library behaviours found empirically

None of these are in the library's docs — its README is still the unedited GitHub template. Each was found by actually running it, and each shaped the implementation:

1. **`predict()` mutates the caller's params dict**, rewriting `granularity` from `'hourly'` to `<Granularity.hourly: 4>`. Reuse that dict for a second series and validation fails, `parse_arg` returns `None`, and `predict` dies with the cryptic `TypeError: cannot unpack non-iterable NoneType object`. Every series now gets a defensive `dict()` copy — locked in by `test_caller_params_are_not_mutated`.
2. **`DetectType.LATEST` returns exactly one result**, not one per row. Mapped onto the final row with nulls before it.
3. **`DetectType.ENTIRE` returns one result per row.**

## Reliability

- **Per-series error isolation** — failures go to `errorCol` instead of being thrown, so one malformed group cannot fail a job spanning thousands of series. The output struct is null for those rows.
- **Struct assembled in Spark, not pandas** — the UDF returns flat scalar columns, avoiding Arrow dict→struct conversion issues.
- `detect_series(pdf, options, detector_factory=None)` is a **pure pandas function**, so the algorithm is unit-testable with an injected fake, with no Spark session and no optional dependency. (Monkeypatching cannot reach PySpark workers — they are separate processes.)

## Testing

**15/15 passing locally.**

- `TestDetectSeries` (7) — algorithm behaviour via an injected fake detector
- `TestTransformerContract` (6) — params, schema, defaults
- `TestEndToEnd` (2) — real library through real Spark; skip in CI via `importorskip`

Also verified: `black --check --extend-exclude 'docs/' .` clean (189 files), `website/sidebars.js` parses and registers the page, doc links resolve, LF endings + trailing newlines throughout.

The `PythonTests` job already runs `PACKAGE: "core"` with no `TEST_SUB_PATH`/`IGNORE_TEST_PATH` filter, so `synapsemltest/timeseries/` is collected automatically — **no `pipeline.yaml` changes needed**.

## Docs placement is deliberate

The page lives under `docs/Explore Algorithms/Anomaly Detection/` rather than `docs/Quick Examples/`, because `website/doctest.py` executes Python blocks **only** in `docs/Quick Examples` (via `WebsiteSamplesTests`) — and the optional dependency isn't installed in CI. A `.md` rather than an `.ipynb` for the same reason: notebooks under `docs/**` are executed by the Databricks/Synapse E2E jobs.

## Not included

**Multivariate anomaly detection (MVAD).** The OSS multivariate model is a torch sliding-window model that does not decompose to a per-row Spark operation the way the univariate detector does. Wrapping it properly is a larger design question; the docs include a driver-side recipe in the meantime.

---

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/cognitive/src/main/python/synapse/ml/cognitive/anomaly.py
+++ b/cognitive/src/main/python/synapse/ml/cognitive/anomaly.py
@@ -1,8 +1,11 @@
 import warnings
 
 warnings.warn(
-    "The 'synapse.ml.cognitive.anomaly' module has been removed. "
-    "The Azure AI Anomaly Detector service has been retired and its transformers "
-    "are no longer available. Use synapse.ml.isolationforest instead.",
+    "The 'synapse.ml.cognitive.anomaly' module has been removed because the Azure AI "
+    "Anomaly Detector service has been retired. For time series anomaly detection use "
+    "synapse.ml.timeseries.UnivariateAnomalyDetector, which runs the same algorithms "
+    "locally through the open source 'time-series-anomaly-detector' package and keeps "
+    "the original parameter and output field names. For point anomaly detection over "
+    "feature vectors use synapse.ml.isolationforest.",
     DeprecationWarning,
 )
--- a/core/src/main/python/synapse/ml/timeseries/UnivariateAnomalyDetector.py
+++ b/core/src/main/python/synapse/ml/timeseries/UnivariateAnomalyDetector.py
@@ -1,0 +1,469 @@
+# Copyright (C) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See LICENSE in project root for information.
+
+"""Spark-native univariate time series anomaly detection.
+
+This module wraps the open source ``time-series-anomaly-detector`` package
+(imported as ``anomaly_detector``), which open sources the algorithms that used
+to back the retired Azure AI Anomaly Detector service.
+
+The library itself is single node pandas code, so this transformer supplies the
+distribution: series are partitioned with ``groupBy().applyInPandas()`` and each
+one is detected independently on an executor.
+"""
+
+from typing import Any, Callable, Dict, List, Optional, Tuple
+
+from pyspark import keyword_only
+from pyspark.ml import Transformer
+from pyspark.ml.param import Param, Params, TypeConverters
+from pyspark.ml.util import DefaultParamsReadable, DefaultParamsWritable
+from pyspark.sql import DataFrame
+from pyspark.sql import functions as F
+from pyspark.sql import types as T
+
+__all__ = ["UnivariateAnomalyDetector"]
+
+MISSING_PACKAGE_MESSAGE = (
+    "UnivariateAnomalyDetector requires the 'time-series-anomaly-detector' package, "
+    "which is not installed. Install it with:\n"
+    "    pip install time-series-anomaly-detector\n"
+    "Note that the install name and the import name differ: the package installs as "
+    "'time-series-anomaly-detector' but imports as 'anomaly_detector'."
+)
+
+DETECT_MODE_ENTIRE = "entire"
+DETECT_MODE_LATEST = "latest"
+
+# Field names are deliberately the camelCase names the retired Azure AI Anomaly
+# Detector transformers emitted, so existing `col("anomalies.isAnomaly")` style
+# expressions keep working after migrating.
+#
+# `severity` is intentionally absent: the retired service returned it, but the
+# open source library does not compute it in either detect mode.
+_RESULT_FIELDS: List[Tuple[str, str, T.DataType]] = [
+    ("isAnomaly", "is_anomaly", T.BooleanType()),
+    ("isPositiveAnomaly", "is_positive_anomaly", T.BooleanType()),
+    ("isNegativeAnomaly", "is_negative_anomaly", T.BooleanType()),
+    ("expectedValue", "expected_value", T.DoubleType()),
+    ("upperMargin", "upper_margin", T.DoubleType()),
+    ("lowerMargin", "lower_margin", T.DoubleType()),
+    ("period", "period", T.IntegerType()),
+]
+
+_TEMP_PREFIX = "__synapseml_uvad_"
+_ERROR_FIELD = _TEMP_PREFIX + "error"
+
+# Parameters forwarded verbatim to the library. The names match the retired
+# service so migrating callers keep their existing tuning values.
+_PASSTHROUGH_PARAMS = [
+    "granularity",
+    "customInterval",
+    "period",
+    "maxAnomalyRatio",
+    "sensitivity",
+    "imputeMode",
+    "imputeFixedValue",
+]
+
+
+def result_schema() -> T.StructType:
+    """Return the struct schema emitted into ``outputCol``."""
+    return T.StructType(
+        [T.StructField(name, dtype, True) for name, _, dtype in _RESULT_FIELDS]
+    )
+
+
+def _default_detector_factory(detect_mode: str) -> Tuple[Any, Any]:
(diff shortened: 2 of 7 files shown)


ranadeepsingh:
Closing this without merging. Recording the full context here so the investigation isn't lost and anyone revisiting this has what they need.

## State of the branch

Complete and green. All 12 checks passed on `0928033637`: `Compile & Style Check`, `Python Style Check`, `build` (Docusaurus), `Scan Website for Dead Links`, CodeQL, dependency review. 15/15 unit tests pass locally. 7 files, +922/−3.

The branch `feat/timeseries-anomaly` is preserved on the fork and this PR can be reopened as-is.

## Why not merge now

**1. The optional dependency can't be exercised in CI.** `time-series-anomaly-detector` requires `mlflow>=3.1.0` (it resolves to 3.15.1). This repo pins:

```yaml
# Petastorm needs legacy Parquet and fsspec APIs removed after PyArrow 10; MLflow accepts this pin.
- mlflow==2.21.3
```

So the package can't go into `environment.yml` without forcing an MLflow major upgrade and risking Petastorm. The import is lazy and the two end-to-end tests `pytest.importorskip`, which means **the only tests that would actually run in CI are the 13 that use an injected fake detector**. Shipping a user-facing algorithm whose real integration path is never exercised by CI is a weak position, and resolving it properly means first deciding the MLflow/Petastorm question — which is a separate change with its own blast radius.

**2. Upstream maturity.** [`microsoft/anomaly-detector`](https://github.com/microsoft/anomaly-detector) is at 0.4.0 (released 2025-11-19, no commits since), ~33 stars, README still the unedited GitHub template. Three behaviours had to be discovered by running it rather than reading docs:

- `predict()` **mutates the caller's params dict**, rewriting `granularity` from `'hourly'` to `<Granularity.hourly: 4>`. Reuse that dict for a second series and validation fails, `parse_arg` returns `None`, and `predict` dies with `TypeError: cannot unpack non-iterable NoneType object`. Worked around with a defensive `dict()` copy per series.
- `DetectType.LATEST` returns **exactly one** result, not one per row.
- `DetectType.ENTIRE` returns one per row.

Taking a hard dependency on a library at this maturity level, for a first-class SynapseML transformer, deserves a maintainer decision rather than being settled inside a PR.

**3. Feature parity is not exact.** `severity` is unavailable in either detection mode, so a genuine drop-in replacement isn't possible for anyone who consumed that field.

## What the branch does contain, if revived

- `synapse.ml.timeseries.UnivariateAnomalyDetector`, distributing detection with `groupBy(groupByCols).applyInPandas(...)` so each series is detected on the executor holding it — verified across 12 series × 200 points on 4 partitions, all 12 planted spikes found, no driver-side iteration.
- All 7 inputs of the retired transformer preserved verbatim (`granularity`, `customInterval`, `period`, `maxAnomalyRatio`, `sensitivity`, `imputeMode`, `imputeFixedValue`), with the OSS snake_case output mapped back to the legacy camelCase contract so `col("anomalies.isAnomaly")` keeps working.
- Per-series error isolation via `errorCol`, so one malformed group can't fail a job spanning thousands of series.
- Rows sorted by timestamp within each group, since the library assumes chronological order and `applyInPandas` does not guarantee it.
- `detect_series(pdf, options, detector_factory=None)` as a pure pandas function, so the algorithm is unit-testable without Spark or the optional dependency.
- Docs deliberately under `docs/Explore Algorithms/Anomaly Detection/` rather than `docs/Quick Examples/`, because `website/doctest.py` executes Python blocks only in the latter and the dependency isn't installed in CI.
- Multivariate detection intentionally not wrapped: the OSS model is a PyTorch sliding-window model that doesn't decompose into a per-row Spark operation.

## Unaffected

[#2605](https://github.com/microsoft/SynapseML/pull/2605) is merged and stands on its own — it removed the retired Azure AI Anomaly Detector wrappers (the service returns `HTTP 410 Gone` on every endpoint) and fixed the failing `UnitTests anomaly` job. Master is green. Nothing here is needed to keep it that way.

For point anomalies on tabular data, `IsolationForest` remains available and supported. It has no notion of time, seasonality, or trend, so it is not a substitute for univariate time series detection — that gap is real and stays open.

PR #2624, opened 2026-08-12
fix(search): accept index schemas whose analyzers and CORS options are objects

### LINKED ISSUE #2143: [BUG] writeToAzureSearch fails when the index has custom analyzers or tokenizers since 0.11.0
### SynapseML version

>=0.11.0

### System information

- Python Standalone - python 3.11.3, scala 2.12, Spark 3.4.1, SynapseML 1.0.2
- Azure Synapse 3.3, SynapseML 0.11.4-spark3.3

### Describe the problem

When a version of Synapse ML is used to load data into an index that has a custom analyzer or tokenizer (and possibly other custom objects but they haven't neem tested) it fails with the following error : -
```
Py4JJavaError: An error occurred while calling z:com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter.write.
: spray.json.DeserializationException: Expected String as JsString, but got {"@odata.type":"#Microsoft.Azure.Search.CustomAnalyzer","charFilters":[],"name":"keyword_analyzer","tokenFilters":["lowercase"],"tokenizer":"keyword_v2"}
```
This happens with all apiVersions set and seemingly any version greater than 0.11.0. It works correctly against the same index when run using a Spark 3.2 Azure Synapse cluster, which uses Synapse ML version 0.10.2

### Create an index with a custom analyzer
This needs to be done through the API: -
POST https://{{service-name}}.search.windows.net/indexes?api-version={{api-version}}
```
{
  "name": "{{index-name}}",
  "defaultScoringProfile": null,
  "fields": [
    {
      "name": "Id",
      "type": "Edm.String",
      "searchable": false,
      "filterable": false,
      "retrievable": true,
      "sortable": false,
      "facetable": false,
      "key": true,
      "indexAnalyzer": null,
      "searchAnalyzer": null,
      "analyzer": null,
      "synonymMaps": []
    }
  ],
  "analyzers": [
    {
    "@odata.type": "#Microsoft.Azure.Search.CustomAnalyzer",
    "name": "keyword_analyzer",
    "tokenizer": "keyword_v2",
    "tokenFilters": ["lowercase"]
    }
    ]
}
```

### Try and load the index
Run the following pyspark on a spark 3.4
```
from pyspark.sql import SparkSession
spark = SparkSession.builder.appName("MyApp") \
    .config("spark.jars.packages", "com.microsoft.azure:synapseml_2.12:1.0.2") \
    .getOrCreate()

import synapse.ml
from synapse.ml.services import writeToAzureSearch
from pyspark.sql.functions import lit, col


df = spark.range(10) \
    .withColumn("Id", col("id").cast("string")) \
    .withColumn("action", lit("upload"))

x = writeToAzureSearch(df, 
        subscriptionKey=admin_key,
        actionCol="action",
        serviceName=search_service,
        indexName=search_index,
        keyCol="FirmId")
```

You can also run the same code (without the spark creation) on Azure Synapse 3.3 and get the same result. I imagine this will happen on Databricks, and Synapse 3.4 but haven't tested it.

### Other info / logs

```
---------------------------------------------------------------------------
Py4JJavaError                             Traceback (most recent call last)
c:\Users\dbundred\Documents\Projects\OSM\OtherStuff\synapseMlTest.ipynb Cell 2 line 1
     11 from pyspark.sql.functions import lit, col
     14 df = spark.range(10) \
     15     .withColumn("Id", col("id").cast("string")) \
     16     .withColumn("action", lit("upload"))
---> 18 x = writeToAzureSearch(df, 
     19         subscriptionKey=admin_key,
     20         actionCol="action",
     21         serviceName=search_service,
     22         indexName=search_index,
     23         keyCol="FirmId")

File ~\AppData\Local\Temp\spark-c09947f2-255d-45b5-a241-6a7165bbac06\userFiles-b3991b65-4cea-4012-980d-108b02406dcc\com.microsoft.azure_synapseml-cognitive_2.12-1.0.2.jar\synapse\ml\services\search\AzureSearchWriter.py:28, in writeToAzureSearch(df, **options)
     26 jvm = SparkContext.getOrCreate()._jvm
     27 writer = jvm.com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter
---> 28 writer.write(df._jdf, options)

File c:\Users\dbundred\AppData\Local\Programs\Python\Python311\Lib\site-packages\py4j\java_gateway.py:1322, in JavaMember.__call__(self, *args)
   1316 command = proto.CALL_COMMAND_NAME +\
   1317     self.command_header +\
   1318     args_command +\
   1319     proto.END_COMMAND_PART
   1321 answer = self.gateway_client.send_command(command)
-> 1322 return_value = get_return_value(
   1323     answer, self.gateway_client, self.target_id, self.name)
   1325 for temp_arg in temp_args:
   1326     if hasattr(temp_arg, "_detach"):

File c:\Users\dbundred\AppData\Local\Programs\Python\Python311\Lib\site-packages\pyspark\errors\exceptions\captured.py:169, in capture_sql_exception.<locals>.deco(*a, **kw)
    167 def deco(*a: Any, **kw: Any) -> Any:
    168     try:
--> 169         return f(*a, **kw)
    170     except Py4JJavaError as e:
    171         converted = convert_exception(e.java_exception)

File c:\Users\dbundred\AppData\Local\Programs\Python\Python311\Lib\site-packages\py4j\protocol.py:326, in get_return_value(answer, gateway_client, target_id, name)
    324 value = OUTPUT_CONVERTER[type](answer[2:], gateway_client)
    325 if answer[1] == REFERENCE_TYPE:
--> 326     raise Py4JJavaError(
    327         "An error occurred while calling {0}{1}{2}.\n".
    328         format(target_id, ".", name), value)
    329 else:
    330     raise Py4JError(
    331         "An error occurred while calling {0}{1}{2}. Trace:\n{3}\n".
    332         format(target_id, ".", name, value))

Py4JJavaError: An error occurred while calling z:com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter.write.
: spray.json.DeserializationException: Expected String as JsString, but got {"@odata.type":"#Microsoft.Azure.Search.CustomAnalyzer","charFilters":[],"name":"keyword_analyzer","tokenFilters":["lowercase"],"tokenizer":"keyword_v2"}
	at spray.json.package$.deserializationError(package.scala:23)
	at spray.json.ProductFormats.fromField(ProductFormats.scala:63)
	at spray.json.ProductFormats.fromField$(ProductFormats.scala:51)
	at com.microsoft.azure.synapse.ml.services.search.AzureSearchProtocol$.fromField(AzureSearchSchemas.scala:67)
	at spray.json.ProductFormatsInstances$$anon$11.read(ProductFormatsInstances.scala:341)
	at spray.json.ProductFormatsInstances$$anon$11.read(ProductFormatsInstances.scala:319)
	at spray.json.JsValue.convertTo(JsValue.scala:33)
	at com.microsoft.azure.synapse.ml.services.search.IndexParser.parseIndexJson(AzureSearchAPI.scala:25)
	at com.microsoft.azure.synapse.ml.services.search.IndexParser.parseIndexJson$(AzureSearchAPI.scala:24)
	at com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter$.parseIndexJson(AzureSearch.scala:147)
	at com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter$.getVectorColConf(AzureSearch.scala:325)
	at com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter$.prepareDF(AzureSearch.scala:269)
	at com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter$.write(AzureSearch.scala:432)
	at com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter$.write(AzureSearch.scala:440)
	at com.microsoft.azure.synapse.ml.services.search.AzureSearchWriter.write(AzureSearch.scala)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
	at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:374)
	at py4j.Gateway.invoke(Gateway.java:282)
	at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
	at py4j.commands.CallCommand.execute(CallCommand.java:79)
	at py4j.ClientServerConnection.waitForCommands(ClientServerConnection.java:182)
	at py4j.ClientServerConnection.run(ClientServerConnection.java:106)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

### What component(s) does this bug affect?

- [X] `area/cognitive`: Cognitive project
- [ ] `area/core`: Core project
- [ ] `area/deep-learning`: DeepLearning project
- [ ] `area/lightgbm`: Lightgbm project
- [ ] `area/opencv`: Opencv project
- [ ] `area/vw`: VW project
- [ ] `area/website`: Website
- [ ] `area/build`: Project build system
- [ ] `area/notebooks`: Samples under notebooks folder
- [ ] `area/docker`: Docker usage
- [ ] `area/models`: models related issue

### What language(s) does this bug affect?

- [ ] `language/scala`: Scala source code
- [X] `language/python`: Pyspark APIs
- [ ] `language/r`: R APIs
- [ ] `language/csharp`: .NET APIs
- [ ] `language/new`: Proposals for new client languages

### What integration(s) does this bug affect?

- [X] `integrations/synapse`: Azure Synapse integrations
- [ ] `integrations/azureml`: Azure ML integrations
- [ ] `integrations/databricks`: Databricks integrations

ranadeepsingh:
## Why

Fixes #2143. An Azure AI Search index that uses a **custom analyzer** (or char filter, tokenizer, token filter, suggester, or CORS configuration) could not be written to *at all*:

```
spray.json.DeserializationException: Expected String as JsString, but got
{"@odata.type":"#Microsoft.Azure.Search.CustomAnalyzer","charFilters":[],
 "name":"keyword_analyzer","tokenFilters":["lowercase"],"tokenizer":"keyword_v2"}
```

Reported against 0.11.0 and still reproducible on `master`.

## Root cause

`IndexInfo` declared six members as `Option[Seq[String]]`, but the Search REST API returns **objects** for every one:

| Member | Declared | Actually returned |
|---|---|---|
| `suggesters` | `Option[Seq[String]]` | array of objects |
| `analyzers` | `Option[Seq[String]]` | array of objects |
| `charFilters` | `Option[Seq[String]]` | array of objects |
| `tokenizers` | `Option[Seq[String]]` | array of objects |
| `tokenFilters` | `Option[Seq[String]]` | array of objects |
| `corsOptions` | `Option[Seq[String]]` | a single object, not an array |

The right-hand column was confirmed against the live service, not inferred from the issue.

`IndexInfo` is deserialized twice from the caller's `indexJson`, both **before any document is sent** — `parseIndexJson(...).name.get` at the top of `createIfNoneExists`, and `validateIndexInfo`. So the write threw straight out of `AzureSearchWriter.write` regardless of `apiVersion` or whether the index already existed.

## The fix

Type the six members as opaque `JsValue`.

The writer never inspects them — it only round-trips them to the service — so `JsValue` is both sufficient and strictly more robust: it won't break again when the service adds a new analyzer or tokenizer kind. Same class of defect as the standing `TODO` about `synonymMaps`.

**In-repo compatibility:** nothing reads these fields (verified by grep), and the only construction site passes `None` for all six. Existing callers compile unchanged, and serialized output is byte-identical since `None` fields are omitted.

**⚠️ External breaking change (for release notes):** `IndexInfo` is a public case class returned by the public `IndexParser.parseIndexJson`, so retyping changes its `apply`/`copy`/accessor signatures. Real exposure is near zero — under the old typing *any* index carrying these features threw on read, so no working downstream code could have depended on the `Seq[String]` shape. There is no MiMa gate here, so CI won't flag it; noting it so it can be captured in the next release notes.

## Tests

**`IndexSchemaParsingSuite`** — 7 offline tests driven by the exact index definition from #2143: custom analyzer parses; custom analyzer survives parse → re-serialize **byte-identically** (the service rejects an analyzer it can't identify, so round-trip fidelity is the property that matters); object-valued `charFilters`/`tokenizers`/`tokenFilters` parse; object-valued `suggesters` parse; `corsOptions` parses as an object not an array; all six at once parse; omitting them yields empty rather than failure. **Every one fails on `master`** with `Expected String as JsString`.

**`IndexSchemaLiveRoundTripSuite`** — 1 end-to-end test. The offline suite asserts against hand-written JSON, so alone it proves the parser *accepts* these shapes, not that they're what the service *emits*. This creates a real index using all six features, reads it back through the same production `getIndexJsonFromExistingIndex` path the writer uses, parses it, and deletes the index in a `finally`. ~21s.

`pipeline.yaml` registers both in the `misc` leg — `PipelineTestCoverageSuite` fails the build otherwise, since nothing globs the `services.search` parent package.

## Proving it's a real fix, not a cosmetic retype

Reverting `AzureSearchSchemas.scala` fails at *compile* time (the tests use the new types), which shows type-safety but not the runtime bug. So a payload captured live from `mmlspark-azure-search` (API `2026-04-01`) was parsed through `parseIndexJson` alone under both schema versions:

| Schema | Result |
|---|---|
| `master`'s `Option[Seq[String]]` | `DeserializationException: Expected String as JsString, but got {"name":"sml_suggester",...}` — verbatim #2143 |
| this PR's `JsValue` | parses successfully |

Incidental finding worth recording: a suggester's `sourceFields` cannot reference a field using a **custom** analyzer — the service returns 400. The test points its suggester at a separate default-analyzer field.

## Validation

`IndexSchemaParsingSuite` 7/7 · `IndexSchemaLiveRoundTripSuite` passes live · `PipelineTestCoverageSuite` passes · `cognitive/scalastyle` + `Test/scalastyle` 0 errors · full ADO pipeline **77/77 green, 0 failures**, including the `UnitTests misc` leg that runs the live suite.

## Checklist

- [x] Tests added, including a live end-to-end test
- [x] No dependency changes
- [x] Bug fix, not a new feature — no website samples needed

### DIFF
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/search/AzureSearchSchemas.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/search/AzureSearchSchemas.scala
@@ -13,17 +13,22 @@ case class ASResponses(value: Seq[ASResponse])
 
 case class ASResponse(key: String, status: Boolean, errorMessage: Option[String], statusCode: Int)
 
+// These six fields are pure pass-through: the writer round-trips them to the service and never
+// inspects their contents. The Search REST API returns objects for all of them (a custom analyzer
+// is {"@odata.type":"#Microsoft.Azure.Search.CustomAnalyzer",...}, corsOptions is a single object),
+// so typing them as strings made any index using these features fail to deserialize. Keeping them
+// as opaque JsValue also stops the library from breaking when the service adds a new analyzer kind.
 case class IndexInfo(
                     name: Option[String],
                     fields: Seq[IndexField],
-                    suggesters: Option[Seq[String]],
+                    suggesters: Option[Seq[JsValue]],
                     scoringProfiles: Option[Seq[ScoringProfile]],
-                    analyzers: Option[Seq[String]],
-                    charFilters: Option[Seq[String]],
-                    tokenizers: Option[Seq[String]],
-                    tokenFilters: Option[Seq[String]],
+                    analyzers: Option[Seq[JsValue]],
+                    charFilters: Option[Seq[JsValue]],
+                    tokenizers: Option[Seq[JsValue]],
+                    tokenFilters: Option[Seq[JsValue]],
                     defaultScoringProfile: Option[String],
-                    corsOptions: Option[Seq[String]],
+                    corsOptions: Option[JsValue],
                     vectorSearch: Option[VectorSearch]
                     )
 
--- a/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/search/AzureSearchSchemas.scala
+++ b/cognitive/src/main/scala/com/microsoft/azure/synapse/ml/services/search/AzureSearchSchemas.scala
@@ -18,17 +18,19 @@ case class ASResponse(key: String, status: Boolean, errorMessage: Option[String]
 // is {"@odata.type":"#Microsoft.Azure.Search.CustomAnalyzer",...}, corsOptions is a single object),
 // so typing them as strings made any index using these features fail to deserialize. Keeping them
 // as opaque JsValue also stops the library from breaking when the service adds a new analyzer kind.
+// spray.json.JsValue is written out in full because the release branches this change is replayed
+// onto import spray.json explicitly rather than by wildcard.
 case class IndexInfo(
                     name: Option[String],
                     fields: Seq[IndexField],
-                    suggesters: Option[Seq[JsValue]],
+                    suggesters: Option[Seq[spray.json.JsValue]],
                     scoringProfiles: Option[Seq[ScoringProfile]],
-                    analyzers: Option[Seq[JsValue]],
-                    charFilters: Option[Seq[JsValue]],
-                    tokenizers: Option[Seq[JsValue]],
-                    tokenFilters: Option[Seq[JsValue]],
+                    analyzers: Option[Seq[spray.json.JsValue]],
+                    charFilters: Option[Seq[spray.json.JsValue]],
+                    tokenizers: Option[Seq[spray.json.JsValue]],
+                    tokenFilters: Option[Seq[spray.json.JsValue]],
                     defaultScoringProfile: Option[String],
-                    corsOptions: Option[JsValue],
+                    corsOptions: Option[spray.json.JsValue],
                     vectorSearch: Option[VectorSearch]
                     )
 
--- a/cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/search/IndexSchemaLiveRoundTripSuite.scala
+++ b/cognitive/src/test/scala/com/microsoft/azure/synapse/ml/services/search/IndexSchemaLiveRoundTripSuite.scala
@@ -1,0 +1,96 @@
+// Copyright (C) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in project root for information.
+
+package com.microsoft.azure.synapse.ml.services.search
+
+import com.microsoft.azure.synapse.ml.Secrets
+import com.microsoft.azure.synapse.ml.core.test.base.TestBase
+import com.microsoft.azure.synapse.ml.io.http.RESTHelpers._
+import org.apache.http.client.methods.HttpDelete
+import spray.json._
+
+import java.util.UUID
+
+/** End-to-end companion to [[IndexSchemaParsingSuite]].
+  *
+  * The offline suite proves the parser accepts the payload shapes reported in issue #2143, but it
+  * asserts against hand-written JSON. This suite creates a real index that exercises every schema
+  * feature the service returns as an object, reads it back through the same production code path
+  * the writer uses, and parses it. That closes the gap where a hand-written fixture could drift
+  * from what the Search service actually emits.
+  */
+class IndexSchemaLiveRoundTripSuite extends TestBase with IndexJsonGetter with IndexParser {
+
+  private lazy val azureSearchKey: String = sys.env.getOrElse("AZURE_SEARCH_KEY", Secrets.AzureSearchKey)
+  private val testServiceName = "mmlspark-azure-search"
+
+  private def indexDefinition(name: String): String =
+    s"""
+       |{
+       |  "name": "$name",
+       |  "fields": [
+       |    { "name": "id", "type": "Edm.String", "key": true, "searchable": true },
+       |    { "name": "text", "type": "Edm.String", "searchable": true, "analyzer": "sml_custom_analyzer" },
+       |    { "name": "title", "type": "Edm.String", "searchable": true }
+       |  ],
+       |  "corsOptions": { "allowedOrigins": ["*"], "maxAgeInSeconds": 300 },
+       |  "suggesters": [
+       |    { "name": "sml_suggester", "searchMode": "analyzingInfixMatching", "sourceFields": ["title"] }
+       |  ],
+       |  "analyzers": [
+       |    {
+       |      "@odata.type": "#Microsoft.Azure.Search.CustomAnalyzer",
+       |      "name": "sml_custom_analyzer",
+       |      "tokenizer": "sml_tokenizer",
+       |      "tokenFilters": ["sml_asciifolding"],
+       |      "charFilters": ["sml_mapping"]
+       |    }
+       |  ],
+       |  "tokenizers": [
+       |    { "@odata.type": "#Microsoft.Azure.Search.KeywordTokenizerV2", "name": "sml_tokenizer" }
+       |  ],
+       |  "tokenFilters": [
+       |    { "@odata.type": "#Microsoft.Azure.Search.AsciiFoldingTokenFilter",
+       |      "name": "sml_asciifolding", "preserveOriginal": true }
+       |  ],
+       |  "charFilters": [
+       |    { "@odata.type": "#Microsoft.Azure.Search.MappingCharFilter",
+       |      "name": "sml_mapping", "mappings": ["a=>b"] }
+       |  ]
+       |}
+       |""".stripMargin
+
+  private def deleteIndex(name: String): Unit = {
+    val apiVersion = AzureSearchAPIConstants.DefaultAPIVersion
+    val deleteRequest = new HttpDelete(
+      s"https://$testServiceName.search.windows.net/indexes/$name?api-version=$apiVersion")
+    deleteRequest.setHeader("api-key", azureSearchKey)
+    safeSend(deleteRequest)
+    ()
+  }
+
+  test("An index using object-valued schema features round-trips through the live service") {
+    val indexName = s"test-schema-${UUID.randomUUID().toString.take(8)}"
+    SearchIndex.createIfNoneExists(azureSearchKey, testServiceName, indexDefinition(indexName))
+    try {
+      // Read back through the same path AzureSearchWriter uses, so this fails if the service
+      // emits a shape the production parser rejects.
+      val liveJson = getIndexJsonFromExistingIndex(azureSearchKey, testServiceName, indexName)
+      val info = parseIndexJson(liveJson)
+
+      assert(info.name.contains(indexName))
+      // The service returns these as objects; typing them as strings is what broke issue #2143.
+      assert(info.analyzers.exists(_.length == 1))
+      assert(info.analyzers.get.head.asJsObject.fields("name") == JsString("sml_custom_analyzer"))
+      assert(info.tokenizers.exists(_.length == 1))
+      assert(info.tokenFilters.exists(_.length == 1))
+      assert(info.charFilters.exists(_.length == 1))
+      assert(info.suggesters.exists(_.length == 1))
+      // corsOptions is a single object rather than an array.
+      assert(info.corsOptions.exists(_.asJsObject.fields("maxAgeInSeconds") == JsNumber(300)))
+    } finally {
+      deleteIndex(indexName)
+    }
+  }
+
+}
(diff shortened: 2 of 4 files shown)


ranadeepsingh:
Fixed a real failure this PR introduced, caught by the `Release Branch Compatibility Check spark4.1` leg.

That check replays the PR as a patch onto the `spark4.1` branch and compiles it. `master` imports spray-json by wildcard (`import spray.json._`), but `spark4.1` imports it explicitly:

\\\scala
import spray.json.{DefaultJsonProtocol, JsonFormat, RootJsonFormat}
\\\

My `IndexInfo` hunk doesn't touch the import block, so `JsValue` resolved on master but not after replay — six `not found: type JsValue` errors.

Writing `spray.json.JsValue` in full makes the hunk self-contained and immune to import drift on any release branch. Adding an import line instead would have been fragile: the import blocks differ between the branches, so that hunk could fail to apply on its own.

Re-validated: `IndexSchemaParsingSuite` 7/7 green, `cognitive/scalastyle` 0 errors.

/azp run

ranadeepsingh:
## End-to-end validation against the live Azure Search service

The offline `IndexSchemaParsingSuite` asserts against hand-written JSON, so it proved the parser *accepts* the shapes reported in #2143 but not that those are the shapes the service actually *emits*. I closed that gap with a real round-trip.

### What the live service actually returns

I created an index on `mmlspark-azure-search` (API version `2026-04-01`) exercising all six object-valued schema features, then read the definition back. The service returns:

| Field | Shape returned |
|---|---|
| `suggesters` | array of objects |
| `analyzers` | array of objects |
| `charFilters` | array of objects |
| `tokenizers` | array of objects |
| `tokenFilters` | array of objects |
| `corsOptions` | **single object**, not an array |

That matches this PR exactly: five fields typed `Option[Seq[JsValue]]` and `corsOptions` typed `Option[JsValue]`.

One incidental finding worth recording: a suggester's `sourceFields` cannot reference a field that uses a **custom** analyzer — the service rejects it with a 400. The test therefore points its suggester at a separate default-analyzer field.

### A/B proof that this is a real fix, not a cosmetic retype

Reverting `AzureSearchSchemas.scala` to master doesn't fail at runtime — it fails at *compile* time, because the tests use the new types. That demonstrates type-safety but not the actual bug. So I captured the real live payload and parsed it through `parseIndexJson` alone, under both versions of the schema file:

- **master's `Option[Seq[String]]` types** → `DeserializationException: Expected String as JsString, but got {"name":"sml_suggester",...}` — verbatim issue #2143.
- **this PR's `JsValue` types** → parses successfully.

So the failure reproduces on genuine service output, and this change is what fixes it.

### New regression test

Added `IndexSchemaLiveRoundTripSuite`, which creates a real index, reads it back through the **production** `getIndexJsonFromExistingIndex` path that `AzureSearchWriter` uses, parses it with `parseIndexJson`, and asserts the six fields deserialize. The index is deleted in a `finally` block. Runs in ~21s.

It's registered in the `misc` leg of `pipeline.yaml` — required, since `PipelineTestCoverageSuite` fails the build on any unclaimed suite and the `services.search` parent package isn't covered by a glob.

### Verification run

```
scalastyle Found 0 errors
PipelineTestCoverageSuite: Tests: succeeded 1, failed 0
IndexSchemaLiveRoundTripSuite + IndexSchemaParsingSuite: Tests: succeeded 8, failed 0
```

All test indexes created during this work have been deleted; the service is back to its original 12 indexes.

BrendanWalsh (approved):
**Approved** — automated multi-model review cycle.

Reviewed by Claude Opus 5 and GPT-5.6 Sol (both at maximum reasoning effort) against a blocking-issues-only rubric, followed by an independent adjudication pass. Final result: **0 blocking findings** on commit `9884920c`. All 30 checks are green.

One finding was raised and was overturned on verification:

**"Preserve the published `IndexInfo` field types" (`AzureSearchSchemas.scala:26`) — not blocking.**
The concern was that retyping `suggesters`/`analyzers`/`charFilters`/`tokenizers`/`tokenFilters` to `Option[Seq[JsValue]]` and `corsOptions` to `Option[JsValue]` is a source-compatibility break for downstream Scala consumers. The supporting evidence was mis-cited: `VectorSchemaMigrationSuite.scala:188` locks down `VectorSearch` and `IndexField` only and asserts nothing whatsoever about `IndexInfo`. Beyond that:

- There is no MiMa or equivalent binary/source compatibility enforcement in this repo (`project/plugins.sbt` has none; zero hits in `build.sbt`).
- `IndexInfo` appears in zero docs and zero notebooks; a repo-wide grep finds **zero** accesses to any of the six retyped members.
- No working downstream consumer can exist in the first place, because the old `Seq[String]` typing threw on every real Azure AI Search payload — which is precisely the bug this PR fixes.

Backward compatibility on the read path was confirmed decisively: spray-json 1.3.5's `AdditionalFormats.JsValueFormat.read(value) = value` is a total identity read, so schemas that previously parsed as plain strings still parse (`JsString` is a `JsValue`). `jsonFormat11` still derives, `Option` fields still omit on `None`, and `prepareEntity` sends `aligned.compactPrint` of the raw JSON rather than a re-serialized `IndexInfo`, so the wire format is untouched. Validation is unweakened — `validateIndexInfo` only checks `name` and the per-field rules, which this PR does not modify.

Non-blocking observations, recorded but not gating:
- `IndexSchemaLiveRoundTripSuite.scala:78` reads the index back immediately after creation without `retryWithBackoff`, unlike all six existing call sites, and the `misc` leg carries no `FLAKY: "true"`. Azure AI Search index creation is eventually consistent, so this can flake. Mitigated in practice by `retryCountOnTaskFailure: 1` on the Unit Test task, which is why it is not gating.
- `IndexSchemaLiveRoundTripSuite.scala:92` calls `deleteIndex` in a `finally` via `safeSend`, which defaults to `expectedCodes = Set()` and therefore throws on a non-2xx cleanup — that would mask the real assertion failure in a compound-failure scenario.

PR #2657, opened 2026-08-18
perf: run InternalCompat Scala and Python validation in parallel lanes

ranadeepsingh:
## What

Splits the `SynapseML-Internal Compatibility Check` (`InternalCompat`) job into two parallel matrix legs — `scala` and `python` — cutting its wall clock from **67.6 min to ~41 min (−39%)**.

> [!IMPORTANT]
> **Stacked on #2655.** This branch is based on that PR's head, so the commit list below currently includes its six commits. Once #2655 merges, this reduces to a **single commit touching one file (+50 / −9)**. Please merge #2655 first.

## Why

`InternalCompat` is the slowest signal on a SynapseML PR. Its Scala and Python halves are independent — neither reads the other's output — and only ran back to back because they happened to share one checkout.

Measured on [build 231549771](https://msdata.visualstudio.com/A365/_build/results?buildId=231549771) (67.6 min, all 67 jobs green):

| Segment | Time | Needed by |
|---|---:|---|
| Shared prefix — checkout, publish OSS to local Maven, retarget, compile | 12.5 min | both legs |
| `Create Internal conda env` | 7.1 min | python only |
| Scala — `spark.aifunc` 22.8m + `ebm` 2.5m + `predict` 0.4m (211 tests) | 25.9 min | scala only |
| Python — package 1.0m + `testPythonExcludeAIFunc` 19.2m (26 tests) | 20.2 min | python only |

## How

This mirrors what SynapseML-Internal's own pipeline already does — it shards into 6 Scala jobs and 8 Python jobs via `templates/scala_test_job.yml` and `templates/python_test_job.yml`.

```yaml
strategy:
  matrix:
    scala:
      COMPAT_LANE: scala
    python:
      COMPAT_LANE: python
```

Shared steps stay ungated; the 9 lane-specific steps get `eq(variables['COMPAT_LANE'], ...)`.

**The Scala leg also skips conda env creation (−7.1 min).** This is not a guess: SynapseML-Internal runs `ScalaAIFuncTests*`, `ScalaEBMTests` and `ScalaPredictTests` with `useConda` unset, reserving that env for its PowerBI and nbtest jobs — and those are exactly the three packages this leg runs. The step already exported `CREATE_SEMPY_WRITER=false`, which is Internal's non-conda path. Activation is now guarded on the env actually existing, so the step still works unchanged if the env is ever reintroduced.

### Why two legs and not more

Every leg re-pays the 12.5 min shared prefix, so sharding has sharply diminishing returns here:

| Legs | Wall clock | Cost |
|---|---:|---|
| 1 (today) | 67.6 min | 1 agent |
| **2 (this PR)** | **~41 min** | 2 agents |
| 3 (split `spark.aifunc` from `ebm`+`predict`) | ~38 min | 3 agents |

Internal can shard 6-and-8 ways because its legs resolve a *published* OSS artifact rather than building one from the change under test, so they have no comparable prefix to amortise.

## Validation

- **YAML parses**; all **12** inline scripts pass `bash -n`.
- **Lane simulation** over the parsed job confirms the 7 shared steps run in both legs, and each of the 9 lane-specific steps runs in exactly one:
  - `lane=scala` → runs 14, skips 7 (all conda + all Python steps)
  - `lane=python` → runs 19, skips 2 (Scala tests + Scala results)
- **The conda-presence predicate was executed**, not just reviewed, against 5 realistic `conda env list` shapes — env present, present-and-active (`*`), absent, conda missing entirely, and a similar-but-different name (`synapseml-internal-old`) — plus a `set -e` safety check confirming the guard does not abort the step when conda is absent. All 5 correct.

## Risk

- `InternalCompat` keeps `continueOnError: true` and is **not** a required status check, so the new per-leg check names (`... Compatibility Check scala` / `... python`) do not affect branch protection.
- Both legs compute the OSS version from the same pinned build commit, so the retarget is identical in each — and #2655 already made that step idempotent.
- One behavioural change beyond the split: the two Python steps go from `succeededOrFailed()` back to `succeeded()`. That condition existed only to keep Python running when the Scala tests failed *in the same job*; the split now guarantees that structurally.

## Not done here

The 12.5 min shared prefix is now the floor. Caching the Internal conda env, or publishing the OSS M2 artifacts once and fanning out, would attack it — but both add serialisation and belong in their own change with their own measurements.

### DIFF
--- a/pipeline.yaml
+++ b/pipeline.yaml
@@ -1371,13 +1371,32 @@ jobs:
   # This still surfaces as its own red check run, so the signal is not lost, matching how
   # ReleaseBranchCompat above is configured.
   continueOnError: true
-  # Measured on build 230962458: conda env 6.4m + compile 1.5m + package 1.2m +
-  # Scala tests ~34m + ExcludeAIFunc. 90 min was not enough and the job was
-  # killed mid-step; 120 leaves headroom now that the live-service suites are gone.
+  # Per lane, not for the whole job. The longest lane is Scala at ~39 min measured
+  # (build 231549771), so 120 is generous; it is kept rather than tightened because
+  # spark.aifunc calls live AI services and its 22.8 min is service-latency bound.
   timeoutInMinutes: 120
   dependsOn: BuildAndCacheSbt
   pool:
     vmImage: $(UBUNTU_VERSION)
+  # Scala and Python validation are independent -- neither reads the other's output --
+  # but the job ran them back to back purely because they shared one checkout. Measured
+  # on build 231549771 (67.6 min total): the shared prefix (checkout, publish OSS to
+  # local Maven, retarget, compile) is ~12.5 min, Scala tests are 25.9 min, and the
+  # Python package+test pair is 20.2 min. Running the halves as matrix legs makes the
+  # wall clock max(~39, ~41) instead of the sum, so PR feedback arrives ~27 min sooner.
+  #
+  # Two legs, not more. Every leg re-pays the ~12.5 min prefix, so a third leg -- moving
+  # spark.aifunc, which is 22.8 of the Scala 25.9 min, away from ebm+predict -- would
+  # only take the wall clock from ~41 to ~38 min while adding a whole agent.
+  # SynapseML-Internal's own pipeline shards far further (6 Scala jobs, 8 Python jobs),
+  # but its legs have no comparable prefix to amortise: they resolve a published OSS
+  # artifact rather than building one from the PR under test.
+  strategy:
+    matrix:
+      scala:
+        COMPAT_LANE: scala
+      python:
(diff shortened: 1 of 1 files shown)


ranadeepsingh:
Superseded by #2658.

This PR was opened from a fork. `InternalCompat` is guarded by `ne(variables['System.PullRequest.IsFork'], 'True')` -- SynapseML-Internal is private and fork PRs cannot authenticate to it -- so build 231584209 **skipped the job entirely** (\SynapseML-Internal Compatibility Check :: completed/skipped\, 66 jobs instead of 67). The change therefore went completely unexercised here.

#2658 carries the identical commit, rebased onto master after #2655 merged, from a branch on `microsoft/SynapseML` so the job actually runs.

Issue #728, opened 2019-11-08
[LightGBM] [Fatal] Socket recv error, code: 104 when training with binary objective. 


ocworld:
**Describe the bug**
I've met a socket error when training with binary objective.

[LightGBM] [Fatal] Socket recv error, code: 104
19/11/07 14:54:06 WARN LightGBMClassifier: LightGBM reached early termination on one worker, stopping training on worker. This message should rarely occur

I've tried to train a model during 500 iterations. However, when training 318 iteration, the error has been occurred and stop my training.

It is a similar issue with #569. However objective parameter value is different. In my experiment, the value is "binary".
 
**To Reproduce**
Steps to reproduce the behavior, code snippets encouraged

**Expected behavior**

**Info (please complete the following information):**
 - MMLSpark Version: [e.g. v0.17]
1.0.0-rc1
 - Spark Version [e.g. 2.4.3]:
2.3.0
 - Spark Platform [e.g. Databricks]
AWS EMR

** Stacktrace**
```
Please post the stacktrace here if applicable
```

If the bug pertains to a specific feature please tag the appropriate [CODEOWNER](https://github.com/Azure/mmlspark/blob/master/CODEOWNERS) for better visibility

**Additional context**
19/11/07 14:41:25 INFO LightGBMClassifier: LightGBM worker calling LGBM_BoosterUpdateOneIter
[LightGBM] [Fatal] Socket recv error, code: 104
19/11/07 14:54:06 WARN LightGBMClassifier: LightGBM reached early termination on one worker, stopping training on worker. This message should rarely occur

imatiach-msft:
@ocworld I believe https://github.com/Azure/mmlspark/issues/569 is already fixed, and this is a different issue.  I think this might either be due to faulty network connection or possibly out of memory, but I can't be sure.  I would have to try and repro the problem.  I have worked on adding saving on each k iterations to try and mitigate this problem, but haven't sent out a PR yet for this functionality.  Will leave this issue here for now in case we repro this.  Do you see this error consistently?

ocworld:
New features that you work on is enjoyed to me. In my experiment, I implemented these functions myself with a hook object as a delegate. After your works about them, can I create pull request to you?

It is hard to reproduce this issue because of AWS EMR cost. I additionally implemented
with retry feature in my app. After that, I have to plan check this error consistently.

imatiach-msft:
hi @ocworld 
[...]

Issue #569, opened 2019-05-16
LightGBMClassifier: multiclass training, workers out of sync and early termination


sunilsu:
I am training a lightGBM classifier on a dataset with 18 classes, ~ 2M rows and ~5900 columns. The data is pretty sparse (density ~ 15%). The class distribution is unbalanced.
```
lgbm = LightGBMClassifier(objective='multiclass', parallelism='voting_parallel')
model = lgbm.fit(train)
```
I have a 20 node yarn cluster, each node has 8 cores and 64G RAM. Memory is plenty for this dataset.

Looking at the executor log, I see that one of the lightGBm worker goes out of sync with the rest. If every worker is on iteration x, this worker is at a higher iteration (x + y). Eventually, I get this error,
```
19/05/16 20:12:04 INFO LightGBMClassifier: LightGBM running iteration: 82 with result: 0 and is finished: false
19/05/16 20:12:04 INFO LightGBMClassifier: LightGBM worker calling LGBM_BoosterUpdateOneIter
19/05/16 20:12:04 INFO LightGBMClassifier: LightGBM running iteration: 82 with result: 0 and is finished: false
19/05/16 20:12:04 INFO LightGBMClassifier: LightGBM worker calling LGBM_BoosterUpdateOneIter
19/05/16 20:12:04 INFO LightGBMClassifier: LightGBM running iteration: 82 with result: 0 and is finished: false
19/05/16 20:12:04 INFO LightGBMClassifier: LightGBM worker calling LGBM_BoosterUpdateOneIter
19/05/16 20:12:04 INFO LightGBMClassifier: LightGBM running iteration: 82 with result: 0 and is finished: false
19/05/16 20:12:04 INFO LightGBMClassifier: LightGBM worker calling LGBM_BoosterUpdateOneIter
19/05/16 20:12:14 INFO LightGBMClassifier: LightGBM running iteration: 99 with result: 0 and is finished: false
[LightGBM] [Fatal] Socket recv error, code: 104
19/05/16 20:12:15 WARN LightGBMClassifier: LightGBM reached early termination on one worker, stopping training on worker. This message should rarely occur
```

Notice all workers are on iteration 82 except one which is on 99. The program continues to run, but not sure how many boosting iterations were run. Also, the warning printed towards the end points something not working right.

sunilsu:
Hi @imatiach-msft,  any updates on this?

imatiach-msft:
[...]

PR #2684, opened 2026-09-01
fix: propagate LightGBM iteration failures

xiaoh1024:
## Related Issues/PRs

Related to #569 and #728.

This PR addresses the error-handling behavior described in those issues. It does
not attempt to resolve the underlying causes of worker or network failures.

`TrainUtils.updateOneIteration` currently catches exceptions raised during a
LightGBM iteration and sets `state.isFinished = true`.

This makes an iteration failure indistinguishable from legitimate training
completion. In distributed training, if one native worker fails, peer workers
can receive errors such as:

```text
Socket recv error, Connection reset by peer
```

SynapseML then logs the exception as early termination. As a result, the fit can
return a partial model instead of reporting the underlying task failure.

This change:

- logs the iteration failure and rethrows the original exception;
- leaves `state.isFinished` unchanged when an iteration throws;
- preserves the existing behavior when LightGBM legitimately reports completion;
- allows existing Spark task and job failure handling to process the error.

It also adds a deterministic regression test using a failing booster. The test
verifies that the original exception is propagated and the training state is not
marked as finished.

## How is this patch tested?

- [x] I have written tests and confirmed the proposed bug fix works.

The focused `TrainUtilsSuite` passes, including the new regression test:

```text
Iteration failures propagate instead of becoming early termination
```

- [ ] Yes. Make sure you have added samples following the steps below.

### DIFF
--- a/lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/TrainUtils.scala
+++ b/lightgbm/src/main/scala/com/microsoft/azure/synapse/ml/lightgbm/TrainUtils.scala
@@ -107,10 +107,8 @@ private[lightgbm] object TrainUtils extends Serializable {
       }
     } catch {
       case e: java.lang.Exception =>
-        log.warn("LightGBM reached early termination on one task," +
-          " stopping training on task. This message should rarely occur." +
-          " Inner exception: " + e.toString)
-        state.isFinished = true
+        log.error("LightGBM iteration failed; propagating the task failure", e)
+        throw e
     }
   }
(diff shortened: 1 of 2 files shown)


xiaoh1024:
Sorry, I am unable to sign the CLA at this time, so I am closing this pull request. Thank you for your time.

ranadeepsingh:
Thanks for flagging this @xiaoh1024 I've made a follow up PR and merged the fixed: https://github.com/microsoft/SynapseML/pull/2695

PR #2709, opened 2026-09-15
docs: fix DeepVisionClassifier typo

leninworld:
## What changes are proposed?\n\nCorrect the `DeepVisionClassifier` class name in the deep-learning getting-started guide.\n\n## How was this tested?\n\n- Verified the class name against the Python implementation.\n- Confirmed the misspelling no longer appears in the current documentation.\n- Ran `git diff --check`.\n\nRepository-wide Black and Scala style checks were not run because this machine does not have Black or a Java runtime installed; the change only affects documentation.


ranadeepsingh:
Thanks for contributing to SynapseML @leninworld!

PR #2724, opened 2026-09-17
fix: preserve Python service parameter bindings

### LINKED ISSUE #2723: fix: generated Python service parameters omit accessors and lose column bindings
## Problem

A follow-up audit of microsoft/SynapseML#2722 found additional, reproducible Python/JVM parameter-contract gaps in AI-service transformers.

The classifier PR repairs `HasRawPredictionCol` exposure. These failures affect the shared `ServiceParam` generator and Python parameter bridge instead. They reproduce on master `1305587a4afe92d27c8e28894b90e38020252e04`, using newly generated Core/Cognitive wrappers and the matching compiled JVM classes on PySpark 3.5.0 and JDK 11. No Azure service request or credentials are needed.

### Missing or unusable accessors

```python
from synapse.ml.services.openai import OpenAIEmbedding
from synapse.ml.services.text import TextSentiment

embedding = OpenAIEmbedding(textCol="body")
embedding._java_obj.getTextCol()  # "body"
embedding.getTextCol()           # AttributeError: no getTextCol

OpenAIEmbedding().setDimensionsCol("embedding_size")
# Py4JError: the generated wrapper calls a JVM setDimensionsCol method
# that does not exist. The JVM supports the column through ServiceParam.

sentiment = TextSentiment().setText(["hello"])
sentiment._java_obj.getScalarParam("text").apply(0)  # "hello"
sentiment.getText()
# Py4JError: the generated wrapper calls a JVM getText method that does not exist.
```

The generator omits column getters for all 580 generated Cognitive `ServiceParam` occurrences inspected. Some named JVM accessors also do not exist, despite the generic scalar/column parameter representation supporting the operation.

### Advertised bulk column configuration fails

```python
from synapse.ml.services.openai import OpenAIPrompt

OpenAIEmbedding().setParams(textCol="body")
# AttributeError: no textCol

OpenAIPrompt().setParams(temperatureCol="sampling")
# AttributeError: no temperatureCol
```

The constructor and `setTextCol` work, and the generated `setParams` signature advertises `textCol`. Its implementation instead sends that alias through ordinary PySpark Param lookup. A service column is the `Right(columnName)` form of the existing `text` parameter, not a separate `textCol` Param.

### Parameters restored from the JVM become stale or lose column bindings

```python
source = OpenAIEmbedding(text="original")
restored = OpenAIEmbedding._from_java(source._java_obj)
restored.setText("updated")
assert restored.getText() == "updated"
restored._transfer_params_to_java()
assert restored.getText() == "original"  # Incorrectly reverts the update.
```

The corresponding column case also fails: restoring a stage with `textCol="body"` and then copying it does not preserve the column binding. Encoding the copied `text` parameter fails with `IllegalArgumentException`. The bridge stores Python `None` for a JVM `Right(columnName)` and subsequently invokes the scalar setter with that value.

`_from_java` is the same bridge used to reconstruct stages from JVM models. This reproduction deliberately holds the source wrapper alive and does not change deserialization policy.

## Expected behavior

- Generate usable scalar and column accessors, with matching type stubs. Preserve stage-specific accessors and their validation where present.
- Make constructor arguments, named setters, and `setParams` agree on service-column aliases.
- Preserve scalar updates and column bindings across Java/Python transfer and copy.
- Keep existing JVM signatures and the serialized `Left(value)` / `Right(columnName)` representation.
- Preserve ordinary Param behavior and OpenAIPrompt's atomic post-processing validation.
- Cover these contracts in Python tests collected by normal CI, without calling a paid service or weakening deserialization safeguards.

## Related work

- microsoft/SynapseML#2722 fixes another generated Python parameter contract, the classifier raw-prediction marker.
- microsoft/SynapseML#582 discusses older Java/Python model-parameter exposure. This report specifically reproduces current `ServiceParam` failures and is not a claim that the previously fixed LightGBM issue has returned.

leninworld:
## Related Issues/PRs

Fixes #2723

Generated Python wrappers now expose scalar and column setters/getters for `ServiceParam` values while keeping those values JVM-owned.

The change:
- preserves column bindings across Java-to-Python transfer and `copy`;
- prevents stale Python scalar values from overwriting newer JVM values;
- routes constructors, `setParams`, and `copy(extra)` through the correct service setters;
- rejects setting both a scalar parameter and its `Col` alias in the same call;
- preserves `OpenAIPrompt` atomic validation by validating service updates on a scratch JVM copy;
- rejects generated `xCol` aliases that collide with existing Params or generated hand-written methods.

Scalar service getters now convert Scala sequences and maps to Python `list` and `dict` values. Direct generated setters and `copy(extra)` retain their existing JVM behavior for `None`; `setParams` rejects explicit `None`, while constructor defaults with `None` are ignored.

## How is this patch tested?

Validation used Linux x86-64 Docker, Java 11, Spark/PySpark 3.5.0, and the locally built Core and Cognitive JARs.

- `PyCodegenSuite`: 19/19 passed.
- Credential-free generated-wrapper and `OpenAIPrompt` Python suites: 27 tests and 37 subtests passed.
- Core and Cognitive compile/test-compile, Scala style, code generation, and Black checks passed.
- A runtime smoke test loaded `OpenAIEmbedding` from the branch-built Cognitive JAR and verified scalar accessors, column accessors, and Scala-sequence-to-Python-list conversion.

### Validation evidence

Scala compilation and generated-wrapper regression tests:

issue-2723-codegen-tests

Docker Spark runtime regression tests:

issue-2723-docker-python-tests

Repaired generated Python API loaded from the branch-built JAR:

issue-2723-api-loaded


ranadeepsingh:
## Maintainer follow-up

Updated this existing PR at `06e16c5a00ab60caf6522979e42918e08831e301`, retaining the contributor's implementation and attribution. Rebased it onto master `cd45147c7025f483e86fc028069d72b070e73a55` and preserved both upstream and contributor code-generation tests when resolving the conflict.

### Remaining correctness fixes

- Successful named scalar/column setters now discard superseded pending Python values. Previously, `set(stage.text, "old")` followed by `setTextCol("body")` reverted to the old scalar during transfer, copy, or the public transform path.
- OpenAIPrompt scratch validation now isolates the Python parameter map as well as the JVM stage. A later invalid argument cannot consume a pending value from the original stage.
- `setResponseSchema` follows the same ordering contract for pending `responseFormat` values.
- Added credential-free regressions for both setter orders, failures, bulk setters, copy, persistence, and a real empty-DataFrame transform. Added usage documentation. Existing JVM signatures and serialized `Left`/`Right` shapes are unchanged.

### Validation of the updated tree

Local runtime: JDK 11, Scala 2.12.17, PySpark 3.5.0, Python 3.12 on WSL. Tests explicitly verified that JVM classes and generated Python wrappers came from this worktree, not a published SynapseML package.

- `PyCodegenSuite`: 20 passed.
- `OpenAIPromptParamsSuite`: 15 passed.
- Python service-bridge, prompt-parameter, and response-schema suites: 47 tests and 158 subtests passed; JUnit reports zero failures, errors, or skips.
- Core/Cognitive compilation, test compilation, Scala style, full code generation, Black 22.3.0, whitespace, and notebook JSON checks passed.

The original implementation reproduced the mixed-setter failures before the fix. Two initial persistence subcases instead hit the existing legacy-UDF deserialization guard; those are not evidence of a stale-binding failure. The new persistence regression uses OpenAIPrompt and requires no deserialization-policy override. No security guard was weakened.

### Azure refusal investigation

Neither earlier run request was followed by an observed code push: the contributor head remained `81ad2c4416f697a4761ca34a782a293fb5581eb0`, and both commands were unedited. The retained merge preview used the older master `6df5898d14526639129db323e93a7a2c6f21df2e`; GitHub reported conflicts against current master. The PR's final `updated_at` matched the refusal comment itself, so that timestamp does not prove a code change.

The stale/conflicting merge preview is the best-supported explanation, but the exact Azure backend comparison is not publicly established. After the rebase, merge preview `9caa67a972d277da03fa8411ce4761457e709c88` had the current master and reviewed head as parents, with a tree identical to the reviewed head. No fork-authorization safeguard was bypassed.

The fresh `/azp run` was [accepted by Azure](https://github.com/microsoft/SynapseML/pull/2724#issuecomment-5722815704). [Build 236480818](https://msdata.visualstudio.com/b9b2accc-2d1c-45b3-9d24-0eb5d78cc47f/_build/results?buildId=236480818) has started its bootstrap job, and its checks are attached to the reviewed head. Full CI is running, not yet passed.

The reviewed diff adds no credential-exfiltration behavior, CI/workflow changes, or dependency changes. Azure validation still runs in a privileged environment, so this safety assessment applies to the exact reviewed head above, not arbitrary future updates.

ranadeepsingh:
## Configuration performance follow-up

Pushed `433e5241f69e8fefe6b55164dee389dc284f89cc` to this existing PR.

The generator now emits immutable service-parameter name metadata. Argument validation uses that local metadata instead of repeated JVM introspection. Wrappers generated before this metadata was introduced retain the existing discovery fallback. OpenAIPrompt only creates a scratch JVM stage when an update actually contains service parameters.

Ordinary parameter updates remain entirely in Python. Service updates still use the existing JVM validation, scratch isolation, and post-success cleanup. No public JVM signatures, serialized parameter shapes, dependencies, CI definitions, or deserialization policies changed.

### Measured result

Local medians for `setParams(outputCol="result")`, using JDK 11, PySpark 3.5.0, and Python 3.12 on WSL:

| Stage | Before this optimization | After | Configuration gateway commands |
|---|---:|---:|---:|
| OpenAIEmbedding | 1.303 ms | 0.002886 ms | 5 -> 0 |
| OpenAIPrompt | 141.303 ms | 0.003339 ms | 18 -> 0 |

The timing probe used seven rounds of 100 calls after warmup. The same-process replay of the original master ordinary-parameter code measured 0.001426 ms and 0.001538 ms respectively, so the remaining overhead is a few microseconds of local validation. These measurements concern driver configuration, not per-row throughput, and absolute timings are environment-specific.

### Regression evidence

- New zero-JVM tests failed against the previous generated wrappers before the optimization.
- Committed deterministic coverage checks first/repeated/empty ordinary updates, a non-service Core wrapper, atomic failure behavior, local alias validation, and compatibility with older wrappers. Py4J reference-cleanup commands are excluded so garbage collection does not make the tests flaky.
- `PyCodegenSuite`: 21 passed. `OpenAIPromptParamsSuite`: 15 passed.
- Service-bridge, prompt-parameter, and response-schema Python suites: 52 repository tests and 160 subtests passed. The separate timing probe also passed, enforcing zero configuration gateway commands and a local median below 0.1 ms.
- Combined JUnit: 213 results, zero failures, errors, or skips.
- Core/Cognitive production/test Scala style, compilation/test compilation, full code generation, Black 22.3.0, and whitespace checks passed.

The earlier binding, copy, persistence, real empty-DataFrame transform, and service atomicity regressions remain green. Timing thresholds are kept out of normal CI; the committed performance guard checks JVM calls instead.

After verifying merge preview `e2690b3bbc619e6eb8ff3415c1b2313e7440fa6b` has the current base and reviewed head as parents and the same tree as the reviewed head, Azure [accepted the fresh run request](https://github.com/microsoft/SynapseML/pull/2724#issuecomment-5736715166). [Build 236574370](https://msdata.visualstudio.com/b9b2accc-2d1c-45b3-9d24-0eb5d78cc47f/_build/results?buildId=236574370) has started its bootstrap job on this head. Full CI is running, not yet passed. The previous build's Spark 4.1 prerequisite conflict is separate and was not changed here.

BrendanWalsh (approved):

BrendanWalsh (approved):

